PR #2, opened 2020-09-07
Statistics funcs cov

Alexander-Makaryev:
Naive implementation np.cov with using of already implemented another functions is added.

Tests for statistics functions are added, all that are available in statistics_tests except few specific (cupy.core._accelerator).
Paths and file names are changed to be the same as in original test suite, test file for numpy.mean is replaced by original.

3 test for 'mean' are passed as it was before changes.
+2 passed tests from 'histogram'.
+2 passed tests from 'correlation' (np.cov).


Alexander-Makaryev:
@Rubtsowa please save original paths, file names and file contents (as it possible) when you add the tests. Please look at these changes for better explanation what I mean.

dpnp/backend_statistics.pyx
```diff
@@ -38,10 +38,44 @@ from dpnp.dpnp_utils cimport checker_throw_type_error, normalize_axis
...
 __all__ += [
+    "dpnp_cov",
     "dpnp_mean"
 ]
 
 
+cpdef dparray dpnp_cov(dparray array1):
+    # behaviour of original numpy
+    if array1.ndim > 2:
+        raise ValueError("array has more than 2 dimensions")
```
shssf, line 49: use function from utils

dpnp/dpnp_iface_statistics.py
```diff
@@ -47,10 +47,37 @@
...
 
 
+def cov(in_array1, y=None, rowvar=True, bias=False, ddof=None, fweights=None, aweights=None):
+    """
+    Estimate a covariance matrix, given data and weights.
+    """
+
+    is_dparray1 = isinstance(in_array1, dparray)
+
+    if (not use_origin_backend(in_array1) and is_dparray1):
```
shssf, line 62: 👍

shssf (approved):

PR #6, opened 2020-09-09
change impl mean

Rubtsowa:


dpnp/backend_statistics.pyx
```diff
@@ -76,65 +77,37 @@ cpdef dparray dpnp_cov(dparray array1):
     return res
 
 
-cpdef dparray dpnp_mean(dparray a, axis):
-    cdef dparray_shape_type shape_a = a.shape
-    cdef long size_a = a.size
-    cdef size_t dim_a = a.ndim
+cpdef dparray dpnp_mean(dparray input, axis):
```
shssf, line 80: 👍

dpnp/dpnp_iface_statistics.py
```diff
@@ -104,31 +105,40 @@ def mean(a, axis=None):
...
+        raise NotImplementedError
 
-    if isinstance(a, dparray):
-        result = dpnp_mean(a, axis=axis)
+    is_input_dparray = isinstance(input, dparray)
+
+    if axis is not None and (axis >= dim_input or -1 * axis >=dim_input):
+        return numpy.mean(input, axis=axis)
+
+    if not use_origin_backend(input) and is_input_dparray:
```
shssf, line 126: it would be better to combine these 3 checks into one.
on lines 110, 123 and 126

shssf (approved):

dpnp/dpnp_iface_statistics.py
shssf: Thank you!
Is it possible to combine last two (112 and 116 lines) into one?
as a result we will have only one `numpy.mean` call in this fucntion.

Rubtsowa: @shssf  112 call numpy.mean, 116 - dpnp_mean

Alexander-Makaryev (approved):
LGTM)

dpnp/dpnp_iface_statistics.py
shssf: yes, and 131 calls numpy.mean also :-)
anyway, let's merge this because it is already looks much better :-)

shssf (approved):

PR #8, opened 2020-09-09
add absolute

Rubtsowa:


dpnp/backend_mathematical.pyx
```diff
@@ -112,6 +113,27 @@ func_map[add_name][int32_name][float32_name] = [float64_name, & custom_elemwise_
...
+    cdef dparray_shape_type shape_x = x.shape
+    cdef size_t dim_x = x.ndim
+
+    if dim_x > 2:
+        raise NotImplementedError
+
+    if dim_x == 2:
+        for i in range(shape_x[0]):
+            for j in range(shape_x[1]):
+                result[i, j] = x[i, j] if x[i, j] >= 0 else -1*x[i, j]
```
shssf, line 129:
```suggestion
                elem = x[i, j]
                result[i, j] = elem if elem >= 0 else -elem
```
rhetorical question - how many time in original code the operator `[]` (__get_item__) called?

shssf: Also, it looks like all elements are involved into this.
So, it is simpler to use something like:
```
for i in range(x.size):
    result.item(i) = x.item(i)
```

tests/test_absolute.py
```diff
@@ -0,0 +1,32 @@
...
+import numpy
+
+
+@pytest.mark.parametrize("type",
+                         [numpy.float64],
+                         ids=['float64'])
+def test_absolute1(type):
+    n = 10**5
+    a = numpy.random.random(n)*2-1
+    ia = inp.array(a)
```
shssf, line 14: Please do not use random in tests and use the same procedure for numpy and dpnp to generate input data.

dpnp/dpnp_iface_mathematical.py
```diff
@@ -102,7 +102,7 @@ def absolute(input):
     input1 = dpnp.asnumpy(input) if is_input_dparray else input
 
     # TODO need to put dparray memory into NumPy call
-    result_numpy = numpy.abs(input)
+    result_numpy = numpy.abs(input1)
```
shssf, line 105: Good catch!

dpnp/backend_mathematical.pyx
```diff
@@ -115,21 +115,23 @@ func_map[add_name][int32_name][int32_name] = [int32_name, & custom_elemwise_add_
 
 
 cpdef dparray dpnp_absolute(dparray x):
-    result = dparray(x.shape, dtype=x.dtype)
-
     cdef dparray_shape_type shape_x = x.shape
     cdef size_t dim_x = x.ndim
 
+    result = dparray(shape_x, dtype=x.dtype)
```
shssf, line 121: 👍 👍 👍 🥇

shssf (approved):

PR #15, opened 2020-09-15
add min

Rubtsowa:


Alexander-Makaryev:
@Rubtsowa 
I don't see 'min' in statistics functions
https://numpy.org/doc/stable/reference/routines.statistics.html
Possible, it should be placed in some other place/files.
But list of statistics functions contains 'amin', that looks very similar to 'min'. Maybe it is alias.
Please look at documentation and clarify about it.
If there are some aliases then add them to interface too.

dpnp/dparray.pyx
```diff
@@ -39,6 +39,7 @@ from libcpp cimport bool
 from dpnp.dpnp_iface_types import *
 from dpnp.dpnp_iface import *
 from dpnp.backend cimport *
+from dpnp.dpnp_iface_statistics import min
```
shssf, line 42: this file should be included as all other in backend.pyx

Alexander-Makaryev:
@Rubtsowa could you please merge changes from master branch?
I expect that at least IntelPython.dpnp (testing on Ubuntu 20.04) test in CI will be green.

PR #49, opened 2020-09-23
fix: random.random param

samir-nasibli:
* processing `size` param when it is scalar int


dpnp/random/_random.pyx
```diff
@@ -285,10 +285,14 @@ def random(size):
...
     if (use_origin_backend(size)):
         return numpy.random.random(size)
-
-    for dim in size:
-        if not isinstance(dim, int):
-            raise TypeError(f"Intel NumPy random.random(): Unsupported dim={type(dim)}")
+    elif isinstance(size, tuple):
+        for dim in size:
+            if not isinstance(dim, int):
+                raise TypeError(f"Intel NumPy random.sample(): Unsupported dim={type(dim)}")
```
shssf, line 291: please use functions from utils to throw an exceptions

shssf (changes requested):
rebase on master
overall idea of the control flow in function is wrong. Most of the cases should go to the default path (numpy).
Example: https://github.com/IntelPython/dpnp/blob/c13df89e02f0d787d0d0b1952e024a14210a002a/dpnp/dpnp_iface_sorting.py#L111-L123

dpnp/random/_random.pyx
samir-nasibli: done

samir-nasibli:
will refactor error raise for other funcs on other PR

PR #103, opened 2020-10-01
fix: update fmap for random module

samir-nasibli:
* removed `DPNP_FN_RANDOM` from `DPNPFuncName` enum

* using `DPNP_FN_UNIFORM` instead of `DPNP_FN_RANDOM`

* Removed duplications


samir-nasibli:
Removed duplications

PR #112, opened 2020-10-05
EIG combine existing MKL and non-existing SYCL to one kernel

densmirn:


dpnp/backend/custom_kernels.cpp
```diff
@@ -161,6 +163,60 @@ void custom_blas_dot_c(void* array1_in, void* array2_in, void* result1, size_t s
...
+{
+    if (!size)
+    {
+        return;
+    }
+
+    _DataType* array = reinterpret_cast<_DataType*>(array_in);
+    _DataType* result = reinterpret_cast<_DataType*>(result1);
+
+    if constexpr (std::is_same<_DataType, double>::value || std::is_same<_DataType, float>::value)
```
shssf, line 177: this code works for `long` and `int`. No reason to limit it by floating point types.

dpnp/backend/custom_kernels.cpp
```diff
@@ -161,6 +163,60 @@ void custom_blas_dot_c(void* array1_in, void* array2_in, void* result1, size_t s
...
+        dpnp_memory_free_c(scratchpad);
+
+        custom_elemwise_transpose_c<_DataType>(
+            syevd_array, {(long)size, (long)size}, {(long)size, (long)size}, {1, 0}, array, size * size);
+
+        dpnp_memory_free_c(syevd_array);
+    }
+    else
+    {
+        // TODO: implement SYCL kernel for int/long input
```
shssf, line 211: must throw an exception

dpnp/backend/backend_iface_fptr.cpp
```diff
@@ -315,10 +315,10 @@ static func_map_t func_map_init()
     fmap[DPNPFuncName::DPNP_FN_DOT][eft_FLT][eft_FLT] = {eft_FLT, (void*)mkl_blas_dot_c<float>};
     fmap[DPNPFuncName::DPNP_FN_DOT][eft_DBL][eft_DBL] = {eft_DBL, (void*)mkl_blas_dot_c<double>};
 
-    fmap[DPNPFuncName::DPNP_FN_EIG][eft_INT][eft_INT] = {eft_DBL, (void*)mkl_lapack_syevd_c<double>};
-    fmap[DPNPFuncName::DPNP_FN_EIG][eft_LNG][eft_LNG] = {eft_DBL, (void*)mkl_lapack_syevd_c<double>};
-    fmap[DPNPFuncName::DPNP_FN_EIG][eft_FLT][eft_FLT] = {eft_FLT, (void*)mkl_lapack_syevd_c<float>};
-    fmap[DPNPFuncName::DPNP_FN_EIG][eft_DBL][eft_DBL] = {eft_DBL, (void*)mkl_lapack_syevd_c<double>};
+    fmap[DPNPFuncName::DPNP_FN_EIG][eft_INT][eft_INT] = {eft_DBL, (void*)custom_lapack_syevd_c<double>};
```
shssf, line 318: Don't understand yet how it worked before. In `double` variant we provide a kernel to the user with `int` input? it should not work.

examples/example7.cpp
```diff
@@ -28,7 +28,7 @@ int main(int, char**)
         array[size * i + i] = i + 1;
     }
 
-    mkl_lapack_syevd_c<double>(array, result, size);
+    custom_lapack_syevd_c<double>(array, result, size);
```
shssf, line 31: could you please change this example input datatypes to `int` (at line 14)?

shssf (approved):
Don't spend more time on this code. I'll take care of it.

densmirn:
> Don't spend more time on this code. I'll take care of it.

Got it, won't prepare patch for this.

dpnp/backend/backend_iface_fptr.cpp
Alexander-Makaryev: It doesn't worked before, `eig` function supports only `float` and `double`.
@densmirn added  `int` and `long` types in map here
https://github.com/IntelPython/dpnp/pull/86/files#diff-cc4d5932a758a21f53df6d834b67eab0R313-R314
Copy-paste mistake, I think.

Alexander-Makaryev:
@densmirn  @shssf 
Computation of eigenvectors and eigenvalues in integer types is like computation of square root when yoy expect that result will be integer. 
Another example, when you use matrix multiplication you can expect that result will be integer if inputs are integer but in case of `eig` floating point result is expected.
If you want to add integer types as input for this function then just make a conversion to floating point types before computation.

PR #123, opened 2020-10-07
build with beta10 packages

Alexander-Makaryev:
Changes in MKL header files make it incompatible with beta09 version.
Don't merge until beta10 will be available for all.


shssf:
The PR has split.
first part is PR #126

PR #124, opened 2020-10-07
MIN and MAX add MKL kernels

densmirn:


dpnp/backend/custom_kernels_statistics.cpp
```diff
@@ -288,17 +300,28 @@ void custom_min_c(void* array1_in, void* result1, const size_t* shape, size_t nd
...
-    policy.queue().wait();
+        event.wait();
+    }
+    else
+    {
+        auto policy = oneapi::dpl::execution::make_device_policy<class custom_min_c_kernel<_DataType>>(DPNP_QUEUE);
 
-    result[0] = *res;
+        _DataType* res = std::min_element(policy, array_1, array_1 + size);
+        policy.queue().wait();
```
shssf, line 321: It looks like you `wait()` for whole `queue` instead wafting for particular `event`.

Alexander-Makaryev: Yes, i think so too. But at this moment we don't know the way how to get single event from PSTL call.

densmirn: I didn't find anything useful about PSTL calls synchronization to fix that somehow.

PR #127, opened 2020-10-08
ABS add MKL kernel

densmirn:


dpnp/backend/custom_kernels_mathematical.cpp
```diff
@@ -50,25 +50,33 @@ void custom_elemwise_absolute_c(void* array1_in, const std::vector<long>& input_
...
-        }
-    };
+            if (array1[idx] >= 0)
+            {
+                result[idx] = array1[idx];
+            }
+            else
+            {
+                result[idx] = -1 * array1[idx];
+            }
```
samir-nasibli, line 71: Please, fix the indentation. Seems like you accidently changed it. Indentation doesn't have any sense for c/c++, but in any way.

PR #129, opened 2020-10-08
add kernel min. 

Rubtsowa:


Alexander-Makaryev:
@Rubtsowa 
I made a commit with changes, please look into it.
I tried to be close to your previous code, I hope that it will help to see difference.

Alexander-Makaryev:
@Rubtsowa 
You can see that parameter `size_t ind` is unnecessary. Possible we should remove `size_t* res_shape` and `size_t res_ndim` too, because they can be easily computed inside of function, in this case our interface will be like in original Numpy.

dpnp/backend.pxd
```diff
@@ -75,6 +75,7 @@ cdef extern from "backend/backend_iface_fptr.hpp" namespace "DPNPFuncName":  # n
         DPNP_FN_MEAN
         DPNP_FN_MEDIAN
         DPNP_FN_MIN
+        DPNP_FN_MIN_AXIS
```
shssf, line 78: Please merge `DPNP_FN_MIN` and `DPNP_FN_MIN_AXIS`

shssf (approved):

PR #135, opened 2020-10-12
tests: more tests for random funcs

samir-nasibli:
* added `test_randn_normal_distribution`. See: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.normaltest.html
* added `test_radnom_seed`
* skipped tests

TODO:
* add `scipy` dependency into test build

### DIFF
--- a/tests/test_random.py
+++ b/tests/test_random.py
@@ -3,7 +3,7 @@ import pytest
 import dpnp.random
 import numpy
 # from scipy import stats
-from numpy.testing import assert_array_equal
+from numpy.testing import assert_allclose
 
 @pytest.mark.parametrize("func",
                          [dpnp.random.rand,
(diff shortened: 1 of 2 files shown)


tests/test_random.py
```diff
@@ -49,3 +50,36 @@ def test_random_check_otput(func):
...
         assert res[i] >= 0.0
         assert res[i] < 1.0
+
+
+def test_randn_normal_distribution():
+    pts = 1000
+    alpha = 0.05
+    dpnp.random.seed(28041990)
+    x = dpnp.random.randn(pts)
+    _, p = stats.normaltest(x)
```
shssf, line 60: How it is works is `stats` commented out above?

tests/test_random.py
```diff
@@ -49,3 +50,36 @@ def test_random_check_otput(func):
...
+    shape = (100, 1)
+    if func in [dpnp.random.rand, dpnp.random.randn]:
+        args = size
+    else:
+        args = shape
+    dpnp.random.seed(seed)
+    a1 = func(args)
+    dpnp.random.seed(seed)
+    a2 = func(args)
+    assert_array_equal(a1, a2)
```
shssf, line 85: I'm not sure this function is suitable to compare floating point arrays. I think `assert_allclose` function should be used
Example:
https://github.com/IntelPython/dpnp/blob/b26d624a8841292ed5c976cd59af1fad69ac9c67/tests/test_linalg.py#L59

shssf: `rtol=1e-05` is suitable for `float` datatype. For `double` it should be much stronger.

shssf (approved):
LGTM in general. Minor changes might be done

tests/test_random.py
samir-nasibli: Added comments

PR #140, opened 2020-10-13
Move search for MKL from setup.py

densmirn:
Tested on branch `feature/find_mkl2` checked out from old `master` (https://github.com/IntelPython/dpnp/compare/master...feature/find_mkl2).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,7 @@ from Cython.Compiler import Options as cython_options
 from utils.command_style import source_style
 from utils.command_clean import source_clean
 from utils.command_build_clib import custom_build_clib
+from utils.dpnp_build_utils import find_mkl
 
 
 """
@@ -94,8 +95,6 @@ Operating System :: POSIX
 Operating System :: Unix
 Operating System :: MacOS
 """
-
-IS_CONDA_BUILD = True if os.environ.get("CONDA_BUILD", None) == "1" else False
 
 IS_WIN = False
 IS_MAC = False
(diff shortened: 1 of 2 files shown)


utils/dpnp_build_utils.py
```diff
@@ -0,0 +1,131 @@
...
+# THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+
+import os
+
+
+IS_CONDA_BUILD = os.environ.get("CONDA_BUILD") == "1"
+
+
+def _find_mkl_by_var(name, rel_include_path="include", rel_libdir_path="lib",
```
shssf, line 34: interface in general not what we expect.
`_find_mkl_in_conda_root` and `_find_mkl_by_var` must be the same function.
I mean, we need a function to search `somelibrary` and `someheader` in `somepaths`. It will be applied to search MKL in conda env, search MKL in oneapi env, and search compiler in some env.
It should be a general function, not explicitly focused to MKL.

You already do it in right way - you implement parameters `rel_include_path`  and `rel_libdir_path`. This is good.
one more minor step - make it more general to allow searching diffrent libraries and includes by given path(directories).

utils/dpnp_build_utils.py
```diff
@@ -0,0 +1,131 @@
...
+    rel_libdir_path : str
+        relative path to the library directory
+    rel_header_path : str
+        relative path to the header
+    rel_lib_path : str
+        relative path to the library
+    verbose : bool
+        to print paths to include and library directories
+    Returns
+    -------
```
shssf, line 53: documentation is great! Thank you!

setup.py
```diff
@@ -197,60 +196,24 @@
...
-
-    if (os.path.exists(_required_header) and os.path.exists(_required_library)):
-        print(
-            f"Intel DPNP: using $MKLROOT based math library. include={_mkl_include_find}, libpath={_mkl_libpath_find}")
-        _mkl_include = [_mkl_include_find]
-        _mkl_libpath = [_mkl_libpath_find]
-
-if _mkl_include is None or _mkl_libpath is None:
-    raise EnvironmentError("Intel DPNP: Unable to find math library")
+_mathlib_include, _mathlib_path = find_mathlib(verbose=True)
```
shssf, line 207: Could you please use the same approach for `find_cmlr` (for the code below)?

shssf: If you like I can merge this PR and next step will be in other PR

densmirn: Let me do the next step in other PR.

utils/dpnp_build_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    ----------
+    verbose : bool
+        to print paths to include and library directories
+
+    Returns
+    -------
+    tuple(list(str), list(str))
+        path to include directory, path to library directory
+    """
+    conda_root_var = "PREFIX" if IS_CONDA_BUILD else "CONDA_PREFIX"
```
shssf, line 93: 👍

PR #142, opened 2020-10-14
BITWISE add SYCL kernels

densmirn:


dpnp/dpnp_iface_bitwise.py
```diff
@@ -59,11 +59,11 @@
 ]
 
 
-def use_dpnp_backend(x1, x2, valid_types):
+def use_dpnp_backend(x1, x2):
     for x in (x1, x2):
         if not isinstance(x, dparray):
             return False
-        if x.dtype not in valid_types:
+        if x.dtype not in (numpy.int64, numpy.int32):
```
shssf, line 66: It is not true for all other functions in the project. This is common function.

dpnp/backend_bitwise.pyx
```diff
@@ -47,54 +47,23 @@ __all__ += [
...
 
 cpdef dparray dpnp_left_shift(dparray array1, dparray array2):
-    cdef dparray result = dparray(array1.shape, dtype=array1.dtype)
-
-    for i in range(result.size):
-        result[i] = array1[i] << array2[i]
-
-    return result
-
+    return call_fptr_2in_1out(DPNP_FN_LEFT_SHIFT, array1, array2, array1.shape)
```
shssf, line 66: Where result array memory is created?

dpnp/dpnp_iface_bitwise.py
```diff
@@ -59,11 +59,11 @@
 ]
 
 
-def use_dpnp_backend(x1, x2, valid_types):
+def use_dpnp_backend(x1, x2):
```
shssf, line 62: It looks like this function is similar to `use_origin_backend`
perhaps it will be better to use `use_origin_backend` instead duplication

dpnp/backend_bitwise.pyx
densmirn: The result array is created inside of `call_fptr_2in_1out`:
https://github.com/IntelPython/dpnp/blob/07bee6d87cd613224059ff30ce4d2511a89f69c3/dpnp/backend.pyx#L284.

dpnp/dpnp_iface_bitwise.py
```diff
@@ -132,20 +105,8 @@ def bitwise_or(x1, x2):
...
-    valid_types = (numpy.int64, numpy.int32, numpy.bool)
-    if use_dpnp_backend(x1, x2, valid_types):
-        return dpnp_bitwise_or(x1, x2)
-
-    return numpy_bitwise_or(x1, x2)
+    valid_dtypes = [numpy.int64, numpy.int32]
+    return check2call(numpy.bitwise_or, dpnp_bitwise_or, valid_dtypes, x1, x2, check_shapes=True, check_dtypes=True)
 
 
 def bitwise_xor(x1, x2):
```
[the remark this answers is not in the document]
shssf: checks for `[numpy.int64, numpy.int32]` can be used in the `x1.shape != x2.shape` way but with no throw exception.
if x1.dtype is in supported_types:
    return dpnp_bitwise_xor()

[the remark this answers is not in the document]
densmirn: @shssf what is better approach in case of different shapes, sizes or dtypes of input arrays? Do we need to raise exception with `checker_throw_value_error` or give a chance to `numpy` to handle this case by calling original (`numpy`) function? I would like to polish function `check2call` (or better name it `check_nd_call`) to handle these cases.

I mean something like that:
```python
if (x1.size != x2.size): 
    checker_throw_value_error("add", "size", x1.size, x2.size)
```
or
```python
if (x1.size != x2.size):
    call_origin(numpy.add, x1, x2, **kwargs)
```

P.S. what do you think the function `check2call` makes sense or not?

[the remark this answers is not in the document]
shssf: I think the best practice to let numpy decide.
We need to do our calculation for only cases where it is profitable. All other must go to fallback.
So, if we have no support for some `sizes` - it must go to fallback (numpy)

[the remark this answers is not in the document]
shssf: I'm not sure but it looks like `check2call` is duplication of existed functions (similar) in other modules.

dpnp/dpnp_utils.pyx
```diff
@@ -95,6 +96,75 @@ def call_origin(function, *args, **kwargs):
...
+    input_arrays : tuple(arrays)
+        input arrays
+    """
+    unique_vals = set(getattr(x, attr_name) for x in input_arrays)
+    if len(unique_vals) > 1:
+        x1, x2, *_ = list(unique_vals)
+        checker_throw_value_error(func_name, attr_name, x1, x2)
+
+
+def check_nd_call(origin_func, dpnp_func, supported_dtypes, *input_arrays,
```
shssf, line 119: We don't need this function. `call_origin` is completely enough. make it simpler.

dpnp/dpnp_iface_bitwise.py
```diff
@@ -295,17 +216,6 @@ def right_shift(x1, x2):
...
-
-    if use_origin_backend(x1):
-        return numpy_right_shift(x1, x2)
-
-    valid_types = (numpy.int64, numpy.int32, numpy.bool)
-    if use_dpnp_backend(x1, x2, valid_types):
-        return dpnp_right_shift(x1, x2)
-
-    return numpy_right_shift(x1, x2)
+    supported_dtypes = [numpy.int64, numpy.int32]
```
shssf, line 219: we don't need this check

PR #143, opened 2020-10-14
ENH: added using `random.seed`

samir-nasibli:
* added `random.seed` func

* added container for random number generation engine into `backend_sycl` class

* update `mkl_rng_gaussian` and `mkl_rng_uniform` funcs

* using engine by reference

* refactor `dpnp/backend/backend_iface.hpp`, `dpnp/backend/backend_iface_fptr.hpp`, `dpnp/random/_random.pyx`

* added `dpnp_srand` cython declared function

* disabled skipping tests for random module.

* update examples/example5.cpp

## Checklist

- [x] Unit tests: disabled skipped


dpnp/backend/queue_sycl.cpp
```diff
@@ -118,7 +119,39 @@ void backend_sycl::backend_sycl_queue_init(QueueOptions selector)
...
+        backend_sycl::destroy_rng_engine();
+    }
+    // TODO:
+    // choose engine as is in numpy
+    // seed number
+    size_t seed = std::time(nullptr);
+    rng_engine = new mkl_rng::philox4x32x10(DPNP_QUEUE, seed);
+}
+
+void backend_sycl::backend_sycl_rng_engine_init(size_t seed)
```
shssf, line 135: Use default argument for the function to avoid function duplication
https://en.cppreference.com/w/cpp/language/default_arguments

dpnp/backend/queue_sycl.cpp
```diff
@@ -118,7 +119,39 @@ void backend_sycl::backend_sycl_queue_init(QueueOptions selector)
...
+    }
+    rng_engine = new mkl_rng::philox4x32x10(DPNP_QUEUE, seed);
+}
+
 void dpnp_queue_initialize_c(QueueOptions selector)
 {
     backend_sycl::backend_sycl_queue_init(selector);
 }
+
+void dpnp_srand_c()
```
shssf, line 149: Are you sure we need two functions with and without argument?
I would propose to use one function and use some `std::time(0)` as initializer at python level

dpnp/random/_random.pyx
```diff
@@ -48,6 +48,17 @@ ctypedef void(*fptr_mkl_rng_gaussian_1out_t)(void *, size_t)
 ctypedef void(*fptr_mkl_rng_uniform_1out_t)(void *, long, long, size_t)
 
 
+cpdef dpnp_srand(seed=None):
+    """
+    Initialize basic random number generator.
+
+    """
+    if seed is None:
+        dpnp_srand_c()
```
shssf, line 57: I would propose to use one function and use some `std::time(0)` as initializer at python level

dpnp/backend/queue_sycl.cpp
samir-nasibli: @shssf Of course, we can add initializing on to python level and use functions with default arguments, but then we will not have default (std::time(0) initializing ) for the backend interface. If it is needed, I can change it.

dpnp/__init__.py
```diff
@@ -39,3 +39,7 @@
 '''
 SYCL queue initialization
 '''
+dpnp_engine_initialize()
```
shssf, line 42: Move this function inside `dpnp_queue_initialize`

dpnp/backend/queue_sycl.cpp
```diff
@@ -83,7 +84,7 @@ void backend_sycl::backend_sycl_queue_init(QueueOptions selector)
 
     if (queue)
     {
-        backend_sycl::destroy();
+        backend_sycl::destroy_queue();
```
shssf, line 87: revert original name

examples/example5.cpp
```diff
@@ -2,6 +2,16 @@
 
 #include <backend_iface.hpp>
 
+void print_dpnp_array(double* arr, size_t size)
```
shssf, line 5: This function is better to use from
https://github.com/IntelPython/dpnp/blob/6d2aaafe24ac927749dcd3c729d798988ffadb98/dpnp/backend/backend_utils.hpp#L52

shssf (approved):

PR #145, opened 2020-10-15
beta10/snyk

pcheremu:


DmitryKizaev:
I think maintaining requirements file in repo is not needed any more, as automatic configuration for Snyk scan had been created by **pcheremu**.

PR #146, opened 2020-10-15
Move search for compiler and omp from setup.py

densmirn:


setup.py
```diff
@@ -49,7 +49,7 @@
 from utils.command_style import source_style
 from utils.command_clean import source_clean
 from utils.command_build_clib import custom_build_clib
-from utils.dpnp_build_utils import find_mathlib
+from utils.dpnp_build_utils import find_cmplr, find_mathlib, find_omp, IS_LIN, IS_WIN
```
shssf, line 52: moving IS_LIN or IS_WIN logic to `utils` is completely wrong. In utils we have to have simple procedure to find library (headers) by paths. Nothing else.

PR #150, opened 2020-10-19
Kernel matrix rank

Rubtsowa:


dpnp/backend/custom_kernels_linalg.cpp
```diff
@@ -0,0 +1,80 @@
...
+        size_t ind = 0;
+        for (size_t j = 0; j < ndim; j++)
+        {
+            ind += (shape[j] - 1) * i;
+        }
+        rank_val += array_1[ind];
+    }
+    result[0] = rank_val;
+
+#if 0
```
shssf, line 68: Please, delete this debug code

dpnp/backend/custom_kernels_linalg.cpp
```diff
@@ -0,0 +1,80 @@
...
+    std::cout << "matrix_rank result " << result[0] << "\n";
+#endif
+}
+
+
+template void
```
shssf, line 74: We have some new method of the kernel specializations. Could you please use it in this code?
Example is in https://github.com/IntelPython/dpnp/blob/85ce442322ad256366122aae283e8292594dd4f8/dpnp/backend/custom_kernels_mathematical.cpp#L139-L159
or 
you could look into PR #151 in case of overall idea

shssf:
if you think it is better to merge this PR and address kernel specialization procedure later - let me know. The PR looks good to me and we can merge it before other changes.

shssf (approved):

PR #153, opened 2020-10-19
REFACTOR: added python layer for random module

samir-nasibli:
* added python layer for `random` module
* moved python declared funcs to python srcs

### DIFF
--- a/dpnp/random/random_iface.py
+++ b/dpnp/random/random_iface.py
@@ -1,0 +1,423 @@
+# cython: language_level=3
+# -*- coding: utf-8 -*-
+# *****************************************************************************
+# Copyright (c) 2016-2020, Intel Corporation
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+# - Redistributions of source code must retain the above copyright notice,
+#   this list of conditions and the following disclaimer.
+# - Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+#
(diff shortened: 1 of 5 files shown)


shssf:
perhaps it is better to be aligned with other modules and rename `random_iface.py` to `dpnp_iface_random.py`

dpnp/random/random_iface.py
```diff
@@ -0,0 +1,423 @@
...
+import numpy
+
+from dpnp.dparray import dparray
+from dpnp.dpnp_utils import *
+from dpnp.random._random import *
+
+
+__all__ = [
+    'rand',
+    'randf',
```
shssf, line 48: I can't find this function in NumPy API. is it in the API?

dpnp/random/random_iface.py
```diff
@@ -0,0 +1,423 @@
...
+    out : (d0, d1, ..., dn)-shaped array of floating-point samples from the standard normal distribution.
+
+    See Also
+    --------
+    standard_normal
+    normal
+
+    """
+
+    if (use_origin_backend(d0)):
```
shssf, line 213: Could you please align overall approach in algorithm with other functions?
example:
https://github.com/IntelPython/dpnp/blob/fb5e82c850bade4e23457e19b1c7826e4c761019/dpnp/dpnp_iface_mathematical.py#L139-L151

shssf:
Many week ago we used wrong approach - we called numpy with some corner cases of the algorithms. Modern direction - call `dpnp_` functions in corner cases selected by `if/else` and all other cases must go to fallback (numpy for now) by default

dpnp/random/random_iface.py
samir-nasibli: You absolutely right, this not in Numpy API. This is actually `ranf`, some mismatch

shssf (approved):

PR #154, opened 2020-10-19
ENH: enhaced mkl_rng_gaussian function signature

samir-nasibli:
* enhaced `mkl_rng_gaussian` function signature

### DIFF
--- a/dpnp/backend/mkl_wrap_rng.cpp
+++ b/dpnp/backend/mkl_wrap_rng.cpp
@@ -24,6 +24,7 @@
 //*****************************************************************************
 
 #include <backend_iface.hpp>
+#include "backend_fptr.hpp"
 #include "backend_utils.hpp"
 #include "queue_sycl.hpp"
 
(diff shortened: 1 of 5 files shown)


dpnp/backend/mkl_wrap_rng.cpp
```diff
@@ -70,8 +65,8 @@ void mkl_rng_uniform(void* result, long low, long high, size_t size)
 
 }
 
-template void mkl_rng_gaussian<double>(void* result, size_t size);
-template void mkl_rng_gaussian<float>(void* result, size_t size);
+template void mkl_rng_gaussian<double>(void* result, double mean, double stddev, size_t size);
```
shssf, line 68: Could you please align perametrization with new approach?
Example:
https://github.com/IntelPython/dpnp/blob/fb5e82c850bade4e23457e19b1c7826e4c761019/dpnp/backend/custom_kernels_linalg.cpp#L72-L80

Or you could look into PR #151 
Also, one more example might be last commit in PR #150

dpnp/random/_random.pyx
```diff
@@ -44,7 +44,7 @@ from dpnp.dparray cimport dparray
 from dpnp.dpnp_utils cimport *
 
 
-ctypedef void(*fptr_mkl_rng_gaussian_1out_t)(void *, size_t)
+ctypedef void(*fptr_mkl_rng_gaussian_1out_t)(void *, double, double, size_t)
```
shssf, line 47: It might be better to use some common type for 2 and 3 parameter (double, double here) to avoid all these types duplication.
Later, these parameters can be casted to _DataType inside kernel

shssf (approved):

PR #159, opened 2020-10-21
ENH: added `random.chisquare`

samir-nasibli:
## Description
chisquare(df[, size]) Draw samples from a chi-square distribution.

## TODO:
* use `ctypes.c_int32` instead of `int` for `df` param ?
* float to int conversions as is in `numpy` ?
* array-like `df` ?

## Checklist

- [x] Unit tests:


shssf (approved):
Great! Thank you!
Need to resolve code conflicts only

PR #168, opened 2020-10-22
ENH: added `random.beta`

samir-nasibli:
## Description
beta(a, b[, size]) Draw samples from a Beta distribution.
Note: tests skipped for GPU, since MKL doesn't support GPU device.

## Reproduce:
```python
>>> np.mean(dpnp.random.beta(a=2.56, b=0.9, size=10**6))
0.7403067647588499
>>> np.var(dpnp.random.beta(a=2.56, b=0.9, size=10**6))
0.04317382590461775
>>> np.mean(np.random.beta(a=2.56, b=0.9, size=10**6))
0.7403013977948621
>>> np.var(np.random.beta(a=2.56, b=0.9, size=10**6))
0.04318079638614026
```

## Checklist

- [x] Unit tests:


samir-nasibli:
@oleksandr-pavlyk Could you please help with setting coeff p and q as is in `numpy.beta`
see : https://github.com/IntelPython/dpnp/blob/6f3010af46ce68b420b4e70c6feac6f9046433a7/dpnp/backend/custom_kernels_random.cpp#L44

shssf (approved):

oleksandr-pavlyk (approved):

dpnp/random/dpnp_iface_random.py
```diff
@@ -60,6 +61,69 @@
...
+                    checker_throw_value_error("beta", "type(dim)", type(dim), int)
+        elif not isinstance(size, int):
+            checker_throw_value_error("beta", "type(size)", type(size), int)
+
+        if a <= 0:
+            checker_throw_value_error("beta", "a", a, "positive")
+        if b <= 0:
+            checker_throw_value_error("beta", "b", b, "positive")
+
+        return dpnp_beta(a, b, size)
```
shssf, line 122: Please change algorithm here to avoid throwing exceptions. All unhandled/exceptional cases should go to fallback
Example
https://github.com/IntelPython/dpnp/blob/a1324565c29084484e89d51400c1717a9e581c22/dpnp/dpnp_iface_statistics.py#L340-L358

shssf (changes requested):
change interface handling algorithm

dpnp/random/dpnp_iface_random.py
samir-nasibli: @shssf all these exceptions as is in numpy API. But I will change it, and move exceptions handling go to fallback

samir-nasibli: But I think this is really bad design. We need our own exception handlers.

PR #170, opened 2020-10-22
ENH: added `random.gamma`

samir-nasibli:
## Description
gamma(shape[, scale, size]) Draw samples from a Gamma distribution.

## TODO:
* array_like of floats for `scale` and `shape` params

## Checklist

- [X] Unit tests:
- [X] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)

### DIFF
--- a/dpnp/random/_random.pyx
+++ b/dpnp/random/_random.pyx
@@ -115,10 +115,10 @@ cpdef dparray dpnp_exponential(double beta, size):
 
 cpdef dparray dpnp_gamma(double shape, double scale, size):
     """
-    Return a random matrix with data from the "gamma" distribution.
+    Returns an array populated with samples from gamma distribution.
 
     `dpnp_gamma` generates a matrix filled with random floats sampled from a
-    univariate "gamma" distribution of `shape` and `scale`.
+    univariate gamma distribution of `shape` and `scale`.
 
     """
 
--- a/dpnp/random/_random.pyx
+++ b/dpnp/random/_random.pyx
@@ -45,6 +45,7 @@ cimport numpy
 __all__ = [
     "dpnp_chisquare",
     "dpnp_exponential",
+    "dpnp_gamma",
     "dpnp_randn",
     "dpnp_random",
     "dpnp_srand",
@@ -54,6 +55,7 @@ __all__ = [
 
 ctypedef void(*fptr_custom_rng_chi_square_c_1out_t)(void *, int, size_t)
 ctypedef void(*fptr_custom_rng_exponential_c_1out_t)(void *, double, double, size_t)
+ctypedef void(*fptr_custom_rng_gamma_c_1out_t)(void *, double, double, size_t)
 ctypedef void(*fptr_custom_rng_gaussian_c_1out_t)(void *, double, double, size_t)
 ctypedef void(*fptr_custom_rng_uniform_c_1out_t)(void *, long, long, size_t)
 
@@ -107,6 +109,33 @@ cpdef dparray dpnp_exponential(double a, double beta, size):
     cdef fptr_custom_rng_exponential_c_1out_t func = <fptr_custom_rng_exponential_c_1out_t > kernel_data.ptr
     # call FPTR function
     func(result.get_data(), a, beta, result.size)
+
+    return result
+
+
+cpdef dparray dpnp_gamma(double shape, double scale, size):
+    """
+    Return a random matrix with data from the "gamma" distribution.
+
+    `dpnp_gamma` generates a matrix filled with random floats sampled from a
+    univariate "gamma" distribution of `shape` and `scale`.
+
+    """
+
+    dtype = numpy.float64
+    # convert string type names (dparray.dtype) to C enum DPNPFuncType
+    cdef DPNPFuncType param1_type = dpnp_dtype_to_DPNPFuncType(dtype)
+
+    # get the FPTR data structure
+    cdef DPNPFuncData kernel_data = get_dpnp_function_ptr(DPNP_FN_EXPONENTIAL, param1_type, param1_type)
+
+    result_type = dpnp_DPNPFuncType_to_dtype( < size_t > kernel_data.return_type)
+    # ceate result array with type given by FPTR data
+    cdef dparray result = dparray(size, dtype=dtype)
+
+    cdef fptr_custom_rng_gamma_c_1out_t func = <fptr_custom_rng_gamma_c_1out_t > kernel_data.ptr
+    # call FPTR function
+    func(result.get_data(), shape, scale, result.size)
 
     return result
 
--- a/dpnp/random/_random.pyx
+++ b/dpnp/random/_random.pyx
@@ -127,7 +127,7 @@ cpdef dparray dpnp_gamma(double shape, double scale, size):
     cdef DPNPFuncType param1_type = dpnp_dtype_to_DPNPFuncType(dtype)
 
     # get the FPTR data structure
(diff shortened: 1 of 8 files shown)


samir-nasibli:
Need to be redesigned, because supported devices are Host and CPU.

https://software.intel.com/content/www/us/en/develop/documentation/oneapi-mkl-dpcpp-developer-reference/top/random-number-generators/distributions/oneapi-mkl-rng-gamma.html

dpnp/random/_random.pyx
```diff
@@ -111,6 +113,33 @@ cpdef dparray dpnp_exponential(double beta, size):
     return result
 
 
+cpdef dparray dpnp_gamma(double shape, double scale, size):
+    """
+    Return a random matrix with data from the "gamma" distribution.
```
oleksandr-pavlyk, line 118: Why is it call "random matrix" ? The code populates a flat buffer with random numbers. 

"Random matrix" has special meaning in computation math, and this has nothing to do with it.

Returns an array populated with samples from gamma distribution. 

Quotes around gamma are not needed.

tests/skipped_tests.tbl
```diff
@@ -20,6 +20,8 @@ tests/test_linalg.py::test_eig_arange[300-float32]
 tests/test_linalg.py::test_eig_arange[300-int64]
 tests/test_linalg.py::test_eig_arange[300-int32]
 tests/test_random.py::test_randn_normal_distribution
+tests/test_random.py::test_radnom_gamma_seed
```
oleksandr-pavlyk, line 23: typo: should be ``test_random_gamma_seed``

dpnp/random/_random.pyx
samir-nasibli: Thank you. This comment is common for all other functions. I will refactor on other PR

oleksandr-pavlyk:
Something is amiss. Basic moments do not match:

```
In [24]: np.var(np.random.gamma(shape=2.56, scale=0.9, size=10**6))
Out[24]: 2.0689620288819444

In [25]: np.mean(np.random.gamma(shape=2.56, scale=0.8, size=10**6))
Out[25]: 2.0468521330146694

In [26]: np.var(np.random.gamma(shape=2.56, scale=0.8, size=10**6))
Out[26]: 1.6367525135977068

In [27]: dpnp.mean(dpnp.random.gamma(shape=2.56, scale=0.8, size=10**6))
Out[27]: 2.559302218900502

In [28]: dpnp.var(dpnp.random.gamma(shape=2.56, scale=0.8, size=10**6))
Out[28]: 6.545711234562073
```

dpnp/random/_random.pyx
samir-nasibli: Fix for all rest functions on #186

samir-nasibli:
> Something is amiss. Basic moments do not match:
> 
> ```
> In [24]: np.var(np.random.gamma(shape=2.56, scale=0.9, size=10**6))
> Out[24]: 2.0689620288819444
> 
> In [25]: np.mean(np.random.gamma(shape=2.56, scale=0.8, size=10**6))
> Out[25]: 2.0468521330146694
> 
> In [26]: np.var(np.random.gamma(shape=2.56, scale=0.8, size=10**6))
> Out[26]: 1.6367525135977068
> 
> In [27]: dpnp.mean(dpnp.random.gamma(shape=2.56, scale=0.8, size=10**6))
> Out[27]: 2.559302218900502
> 
> In [28]: dpnp.var(dpnp.random.gamma(shape=2.56, scale=0.8, size=10**6))
> Out[28]: 6.545711234562073
> ```

@oleksandr-pavlyk Thank you for the report. I have fixed it.
Reproduce:
````python
>>> np.var(np.random.gamma(shape=2.56, scale=0.8, size=10**6))
1.6414887360289188
>>> np.mean(np.random.gamma(shape=2.56, scale=0.8, size=10**6))
2.0480613902279794
>>> np.var(dpnp.random.gamma(shape=2.56, scale=0.8, size=10**6))
1.6337041157687837
>>> np.mean(dpnp.random.gamma(shape=2.56, scale=0.8, size=10**6))
2.049404203568917
```

shssf (approved):

oleksandr-pavlyk:
Expectations of bounded functions, or moments for light-tailed distributions is appropriate.

An appropriate check is that sample expectation is within several numbers (say 5) of standard deviations of the estimator from its mean. The mean needs to be known analytically, while the standard deviation may be estimated from the sample

oleksandr-pavlyk:
Having built `dpnp` with oneAPI 2020.8.0.1005:

```
In [1]: import dpnp
Running on: Intel(R) Gen9
DPCtrl SYCL queue used
SYCL kernels link time: 0.182214 (sec.)


In [2]: import dpctl

In [3]: dpctl.set_default_queue('opencl', 'cpu', 0)

In [5]: X = dpnp.random.gamma(shape=125.4, scale=0.02, size=10**5)
terminate called after throwing an instance of 'oneapi::mkl::unsupported_device'
  what():  oneMKL: rng/generate: Intel(R) Gen9 is not supported
Aborted (core dumped) 
``` 

I was expecting `dpnp` to use queue from dpctl which was just set to be CPU.

It would greatly enhance user's  experience if these C++ exceptions where intercepted and rethrown as Python exceptions.

oleksandr-pavlyk:
> Having built `dpnp` with oneAPI 2020.8.0.1005 (beta 10):
> 
> ```
> In [1]: import dpnp
> Running on: Intel(R) Gen9
> DPCtrl SYCL queue used
> SYCL kernels link time: 0.182214 (sec.)
> 
> 
> In [2]: import dpctl
> 
> In [3]: dpctl.set_default_queue('opencl', 'cpu', 0)
> 
> In [5]: X = dpnp.random.gamma(shape=125.4, scale=0.02, size=10**5)
> terminate called after throwing an instance of 'oneapi::mkl::unsupported_device'
>   what():  oneMKL: rng/generate: Intel(R) Gen9 is not supported
> Aborted (core dumped) 
> ```
> 
> I was expecting `dpnp` to use queue from dpctl which was just set to be CPU.
> 

Random number generation is not honoring `dpctl` supplied queue because it is using the cached random engine, which does have a queue retrieved at the module initialization time stored in the instance, so while `DPNP_CURRENT_QUEUE` gives the queue associated with OpenCL CPU device, the engine still uses the Level-0 GPU device. 

@shssf I think this is a problem.

samir-nasibli:
> It would greatly enhance user's experience if these C++ exceptions where intercepted and rethrown as Python exceptions.

 Added issue #201 for this. Will be fixed in other PR

oleksandr-pavlyk:
`shape=0` parameter causes oneMKL exception. These can be either disallowed, or handled as a special case using `dpnp.zeros`.

samir-nasibli:
> `shape=0` parameter causes oneMKL exception. These can be either disallowed, or handled as a special case using `dpnp.zeros`.

AttributeError: type object 'dpnp.dparray.dparray' has no attribute 'zeros'.
But I have fixed case, when `shape=0`

oleksandr-pavlyk:
> > `shape=0` parameter causes oneMKL exception. These can be either disallowed, or handled as a special case using `dpnp.zeros`.
> 
> AttributeError: type object 'dpnp.dparray' has no attribute 'zeros'.
> But I have fixed case, when `shape=0` via `dparray.fill`

Fill works. `zeros` is not a method of `dparray`, it is function in module, `dpnp.zeros`. It will produce a new array, so instead of 

```
result = dpnp.array(size, dtype=dtype)
result.fill(0.0)
```

you could simply write

```
result = dpnp.zeros(size, dtype=dtype)
```

oleksandr-pavlyk (approved):
Further changes must be done outside of this PR.

shssf (approved):

PR #174, opened 2020-10-22
FIX: `random.chisquare` fallback to `numpy`

samir-nasibli:
* `random.chisquare` fallback to `numpy`
* removed tests ???

## TODO:
NOTE: problems with seed, when using `original_call`. Need to be fixed or documented `dpnp.random.seed` and `np.random.seed`

### DIFF
--- a/dpnp/random/dpnp_iface_random.py
+++ b/dpnp/random/dpnp_iface_random.py
@@ -98,26 +98,12 @@ def chisquare(df, size=None):
 
     """
 
-    if not use_origin_backend(df):
-        if size is None:
-            size = 1
-        elif isinstance(size, tuple):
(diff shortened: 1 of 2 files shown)


samir-nasibli:
Need to be redesigned, because supported devices are Host and CPU.

https://software.intel.com/content/www/us/en/develop/documentation/oneapi-mkl-dpcpp-developer-reference/top/random-number-generators/distributions/oneapi-mkl-rng-chi-square.html

@shssf Is it needed to remove backend funcs and call at all? And how manage it when `if not use_origin_backend(df):` is `True`

samir-nasibli:
was partially resolved in #183 will be fixed in another way.

PR #184, opened 2020-10-23
TEST: update tests configuration

samir-nasibli:
* added `tests/skipped_tests_gpu.tbl` for functionality, where only `gpu` device is not supported, but `cpu` and `host` is OK

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -32,13 +32,25 @@ skip_mark = pytest.mark.skip(reason='Skipping test.')
 
 
 def pytest_collection_modifyitems(config, items):
-    excluded_tests = []
+    def get_excluded_tests(test_exclude_file):
+        if os.path.exists(test_exclude_file):
+            excluded_tests = []
+            with open(test_exclude_file) as skip_names_file:
+                excluded_tests = skip_names_file.readlines()
+            return excluded_tests
+
     # global skip file
     test_path = os.path.split(__file__)[0]
     test_exclude_file = os.path.join(test_path, 'skipped_tests.tbl')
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -32,13 +32,25 @@
...
+            return excluded_tests
+
     # global skip file
     test_path = os.path.split(__file__)[0]
     test_exclude_file = os.path.join(test_path, 'skipped_tests.tbl')
-    if os.path.exists(test_exclude_file):
-        with open(test_exclude_file) as skip_names_file:
-            excluded_tests = skip_names_file.readlines()
+
+    # additional skip file, where gpu queue not supported
```
shssf, line 46: it should not be additional file. It should be standalone files with all tests disabled for GPU.

tests/conftest.py
```diff
@@ -32,13 +32,25 @@
 
 
 def pytest_collection_modifyitems(config, items):
-    excluded_tests = []
+    def get_excluded_tests(test_exclude_file):
```
shssf, line 35: move this function out. It is better to use standalone functions instead nested.

tests/skipped_tests_gpu.tbl
```diff
@@ -0,0 +1,4 @@
+tests/test_random.py::test_random_input_size[chisquare]
+tests/test_random.py::test_random_input_shape[chisquare]
+tests/test_random.py::test_radnom_chisquare_seed
+tests/test_random.py::test_chisquare_invalid_df
```
shssf, line 4: do not forget to put `end-of-line` at the end of the file.

PR #191, opened 2020-10-25
remove string mkl

shssf:


Alexander-Makaryev:
@shssf This command removes MKL from Python environment if `MKLROOT` environment variable was provided, and build/test should be made with MKL from `MKLROOT`. It was necessary when we had not actual MKL Python package, and old version of MKL was installed in Python environment as dependency. If with new MKL Python packages it works fine, then this command can be removed. But I think that if `MKLROOT` is provided, then it is nice to be sure that you have no another MKL and exactly MKL from `MKLROOT` is used.

Issue #94, opened 2020-09-29
Performance testing.


shssf:
Need to implement performance testing.
- It should be easy to use for developer - like running regular test system.
- Output must be readable and ready to be pasted to a spreadsheet for further analysis.

Example
```
unning on: Intel(R) Core(TM) i5-4570 CPU @ 3.20GHz
queue initialization time:0.000660675 (sec.)
...Test data type is <class 'numpy.float64'>, each test repetitions 5
type:float64:N:  16:Python:9.775e-06:MKL:1.054e-04:ratio:  0.09:verification:True
type:float64:N:  32:Python:3.409e-05:MKL:3.567e-03:ratio:  0.01:verification:True
type:float64:N:  64:Python:7.796e-05:MKL:5.098e-03:ratio:  0.02:verification:True
type:float64:N: 128:Python:3.386e-04:MKL:8.929e-04:ratio:  0.38:verification:True
type:float64:N: 256:Python:1.574e-03:MKL:6.076e-03:ratio:  0.26:verification:True
type:float64:N: 512:Python:7.821e-03:MKL:6.583e-03:ratio:  1.19:verification:True
type:float64:N:1024:Python:3.879e-02:MKL:5.012e-02:ratio:  0.77:verification:True
...Test data type is <class 'numpy.float32'>, each test repetitions 5
type:float32:N:  16:Python:5.007e-06:MKL:5.565e-04:ratio:  0.01:verification:True
type:float32:N:  32:Python:5.722e-06:MKL:5.448e-04:ratio:  0.01:verification:True
type:float32:N:  64:Python:7.653e-05:MKL:8.216e-04:ratio:  0.09:verification:True
type:float32:N: 128:Python:2.210e-04:MKL:4.613e-03:ratio:  0.05:verification:True
```
fields delimiter is ":" by default

PR #192, opened 2020-10-25
Prototype performance testing

densmirn:
It's an attempt to prototype performance testing system based on SDC performance testing system.

Example of results printed to console and dumped to `perf_results.csv`:
```
Performance testing results:
                                   min       max    median
name   lib   dtype   size
matmul dpnp  float32 262144   0.000019  0.000019  0.000019
                     1048576  0.000713  0.002256  0.001130
                     4194304  0.004295  0.006102  0.005592
             float64 262144   0.000027  0.000027  0.000027
                     1048576  0.002318  0.004182  0.003549
                     4194304  0.008195  0.011136  0.009003
             int32   262144   0.000271  0.000325  0.000276
                     1048576  0.000938  0.000978  0.000946
                     4194304  0.003910  0.003976  0.003933
             int64   262144   0.000208  0.000231  0.000210
                     1048576  0.000978  0.000993  0.000985
                     4194304  0.004292  0.004414  0.004302
       numpy float32 262144   0.000023  0.000025  0.000024
                     1048576  0.000280  0.000592  0.000412
                     4194304  0.005013  0.006214  0.005770
             float64 262144   0.000018  0.000018  0.000018
                     1048576  0.000273  0.000794  0.000388
                     4194304  0.004611  0.006216  0.005701
             int32   262144   0.000145  0.000146  0.000145
                     1048576  0.000580  0.001509  0.000621
                     4194304  0.003369  0.003444  0.003374
             int64   262144   0.000153  0.000154  0.000154
                     1048576  0.000941  0.000970  0.000955
                     4194304  0.004201  0.004259  0.004222
```


shssf:
@densmirn Is it possible to change numbers format(ex. `0.000023`) to some scientific one (ex. `2.31E-05`)? two digit after dot is enough.

shssf:
Just for output format example https://github.com/IntelPython/dpnp/blob/2630983ebec98cfdb6137d8625e863826350e9ad/examples/example1.py#L89

tests/tests_perf/test_perf_utils.py
```diff
@@ -0,0 +1,99 @@
...
+    exec_times = []
+
+    # Warming up
+    f(*args)
+
+    for _ in range(repeat):
+        start = time.time()
+        for _ in range(number):
+            f(*args)
+        finish = time.time()
```
shssf, line 62: Not sure it is good idea. It gives us average among many function calls. I'm not sure about it. need to think.

densmirn: You are right. So I applied module `timeit` to measure execution times. This module also uses parameters `repeat` and `number`. The parameter `number` doesn't average time among many function calls and just returns full time of all the function calls.

densmirn: Moreover we are able to change the parameters values in the tests. So if it needed we can call the function only once.

densmirn: Current example of the results:
```
Performance testing results:
                                   min       max    median
name   lib   dtype   size
matmul dpnp  float32 262144  1.880e-03 1.915e-03 1.882e-03
                     1048576 8.903e-02 1.838e-01 1.008e-01
                     4194304 4.716e-01 5.019e-01 4.901e-01
             float64 262144  2.743e-03 2.784e-03 2.750e-03
                     1048576 1.338e-01 3.595e-01 3.000e-01
                     4194304 8.860e-01 1.142e+00 9.886e-01
             int32   262144  2.822e-02 4.114e-02 2.903e-02
                     1048576 9.454e-02 9.820e-02 9.774e-02
                     4194304 3.919e-01 3.968e-01 3.937e-01
             int64   262144  2.111e-02 2.403e-02 2.141e-02
                     1048576 9.981e-02 1.012e-01 1.000e-01
                     4194304 4.261e-01 4.317e-01 4.280e-01
       numpy float32 262144  1.760e-03 1.786e-03 1.763e-03
                     1048576 2.955e-02 9.488e-02 6.367e-02
                     4194304 5.124e-01 6.166e-01 5.485e-01
             float64 262144  1.772e-03 1.828e-03 1.779e-03
                     1048576 3.245e-02 7.371e-02 4.669e-02
                     4194304 4.638e-01 6.217e-01 5.696e-01
             int32   262144  1.452e-02 1.457e-02 1.454e-02
                     1048576 6.012e-02 1.048e-01 6.061e-02
                     4194304 3.205e-01 3.313e-01 3.290e-01
             int64   262144  1.490e-02 1.537e-02 1.503e-02
                     1048576 9.386e-02 9.508e-02 9.464e-02
                     4194304 4.207e-01 4.335e-01 4.267e-01
```

Now we don't divide execution time by number=100.

shssf:
overall feeling is `too complicated` for this proposal. For example, I don't think we need Pandas dependency.

I think we need some graphical representation of the numbers. Do you have some idea which library is better to use for this?
I looked into http://www.pygal.org/en/latest/documentation/types/line.html and https://matplotlib.org/ but the last one looks a bit overcomplicated.

densmirn:
> overall feeling is `too complicated` for this proposal. For example, I don't think we need Pandas dependency.
> 
> I think we need some graphical representation of the numbers. Do you have some idea which library is better to use for this?
> I looked into http://www.pygal.org/en/latest/documentation/types/line.html and https://matplotlib.org/ but the last one looks a bit overcomplicated.

Maybe this proposal is really too complicated, but we can keep it as one of possible options at the least in the separate branch.
I think it's complicated to represent the data in a readable form if it's needed without `Pandas`. As other option we can store result in dictionary just for plotting.
I didn't heard about `line` at all, but a bit used `matplotlib` and would say that it doesn't look complicated to draw 2 lines on the graph with `matplotlib`. I vote for `matplotlib` because used only this library for plotting and it looks like the most popular library for the plotting .

tests/tests_perf/test_perf_base.py
```diff
@@ -27,28 +27,46 @@
...
+    repeat=15
+    sep = ":"
+    results_data = dict()
+    print_width = [10, 8, 6, 10]
+    print_num_width = 10
+    
     @classmethod
     def setup_class(cls):
-        cls.test_results = TestResults()
+        results_data = dict()
```
densmirn: Python has `defaultdict` that is able to simplify method `add`:
```python
from collections import defaultdict
nested_dict = lambda: defaultdict(nested_dict)
nest = nested_dict()
```
```python
def add(self, name, lib, dtype, size, result):
   self.results_data[name][dtype][lib][size] = result

File: tests/tests_perf/test_perf_base.py [hunk 3]
```diff
@@ -0,0 +1,166 @@
...
+    print_num_width = 10
+    
+    @classmethod
+    def setup_class(cls):
+        cls.results_data.clear()
+
+
+    @classmethod
+    def teardown_class(cls):
+        cls.print_csv(cls)
```
densmirn, line 50: Python has `defaultdict` that is able to simplify method `add`:
```python
from collections import defaultdict
nested_dict = lambda: defaultdict(nested_dict)
nest = nested_dict()
```
```python
def add(self, name, lib, dtype, size, result):
   self.results_data[name][dtype][lib][size] = result
```

tests/tests_perf/test_perf_base.py
```diff
@@ -0,0 +1,166 @@
...
+    seed = 777
+    repeat=15
+    sep = ":"
+    results_data = dict()
+    print_width = [10, 8, 6, 10]
+    print_num_width = 10
+    
+    @classmethod
+    def setup_class(cls):
+        cls.results_data.clear()
```
densmirn, line 45: Python has `defaultdict` that is able to simplify method `add`:
```python
from collections import defaultdict
nested_dict = lambda: defaultdict(nested_dict)
cls.results_data = nested_dict()
```
```python
def add(self, name, lib, dtype, size, result):
   self.results_data[name][dtype][lib][size] = result
```

shssf: I tried this And found it was not work as expected

shssf:
This PR is related to issue #94

PR #193, opened 2020-10-26
add SYCL queue type check

shssf:


tests/conftest.py
```diff
@@ -48,7 +49,9 @@ def pytest_collection_modifyitems(config, items):
     # global skip file, where gpu device is not supported
     test_exclude_file_gpu = os.path.join(test_path, 'skipped_tests_gpu.tbl')
 
-    if 'DPNP_QUEUE_GPU' in os.environ and os.getenv('DPNP_QUEUE_GPU') == '1':
+    current_queue_is_cpu = dpnp.dpnp_queue_is_cpu()
+    print(f"\nDPNP: current queue is CPU: {current_queue_is_cpu}")
+    if not current_queue_is_cpu or ('DPNP_QUEUE_GPU' in os.environ and os.getenv('DPNP_QUEUE_GPU') == '1'):
```
densmirn, line 54: It will be enough:
```suggestion
    if not current_queue_is_cpu or os.getenv('DPNP_QUEUE_GPU') == '1':
```

dpnp/backend/backend_iface.hpp
```diff
@@ -78,6 +78,14 @@ enum class QueueOptions : uint32_t
  */
 INP_DLLEXPORT void dpnp_queue_initialize_c(QueueOptions selector);
 
+/**
+ * @ingroup BACKEND_API
+ * @brief SYCL queue device status.
+ *
+ * Return 1 is current @ref queue is related to cpu or host device. return 0 otherwise.
```
Alexander-Makaryev, line 85:
```suggestion
 * Return 1 if current @ref queue is related to cpu or host device. return 0 otherwise.
```

dpnp/backend/queue_sycl.hpp
```diff
@@ -101,6 +101,11 @@ class backend_sycl
      */
     static void backend_sycl_queue_init(QueueOptions selector = QueueOptions::CPU_SELECTOR);
 
+    /**
+     * Return True is current @ref queue is related to cpu or host device
```
Alexander-Makaryev, line 105:
```suggestion
     * Return True if current @ref queue is related to cpu or host device
```

densmirn (approved):

PR #194, opened 2020-10-26
Feature/corr

densmirn:


dpnp/backend/custom_kernels.cpp
```diff
@@ -111,23 +111,27 @@ void dpnp_matmul_c(void* array1_in, void* array2_in, void* result1, size_t size_
...
 
     if (!size)
     {
         return;
     }
 
-    if constexpr (std::is_same<_DataType, double>::value || std::is_same<_DataType, float>::value)
+    if constexpr ((std::is_same<_DataType_input1, double>::value ||
+                   std::is_same<_DataType_input1, float>::value) &&
+                  std::is_same<_DataType_input2, _DataType_input1>::value)
```
shssf, line 134: `_DataType_output` should also be involved into this

shssf: because potentially it might be `double`, `double`, `int`

shssf (changes requested):
perfect. one small change is required

PR #195, opened 2020-10-27
add axis in sum

Rubtsowa:


tests/skipped_tests.tbl
```diff
@@ -557,12 +557,21 @@ tests/third_party/cupy/math_tests/test_sumprod.py::TestNansumNanprodLong_param_1
...
 tests/third_party/cupy/math_tests/test_sumprod.py::TestNansumNanprodLong_param_9_{axis=0, func='nanprod', keepdims=True, shape=(2, 3, 4), transpose_axes=False}::test_nansum_all
 tests/third_party/cupy/math_tests/test_sumprod.py::TestNansumNanprodLong_param_9_{axis=0, func='nanprod', keepdims=True, shape=(2, 3, 4), transpose_axes=False}::test_nansum_axis_transposed
+tests/third_party/cupy/math_tests/test_sumprod.py::TestSumprod::test_external_sum_axis
 tests/third_party/cupy/math_tests/test_sumprod.py::TestSumprod::test_prod_all
 tests/third_party/cupy/math_tests/test_sumprod.py::TestSumprod::test_prod_axis
 tests/third_party/cupy/math_tests/test_sumprod.py::TestSumprod::test_prod_dtype
 tests/third_party/cupy/math_tests/test_sumprod.py::TestSumprod::test_sum_all2
 tests/third_party/cupy/math_tests/test_sumprod.py::TestSumprod::test_sum_all_transposed2
+tests/third_party/cupy/math_tests/test_sumprod.py::TestSumprod::test_sum_axes
+tests/third_party/cupy/math_tests/test_sumprod.py::TestSumprod::test_sum_axes2
```
shssf, line 567: These tests stops working in this PR? It is not expected.

dpnp/backend_mathematical.pyx
```diff
@@ -205,7 +205,7 @@ cpdef dparray dpnp_subtract(dparray x1, dparray x2):
     return call_fptr_2in_1out(DPNP_FN_SUBTRACT, x1, x2, x1.shape)
 
 
-cpdef dpnp_sum(dparray x1):
+cpdef dparray _dpnp_sum(dparray x1):
```
shssf, line 208: please rename it to `dpnp_sum_no_axis` or better move it into original function `dpnp_sum`

shssf: it will allow to reuse `return_type ` instead `res_type`

dpnp/backend_mathematical.pyx
```diff
@@ -221,7 +221,91 @@ cpdef dpnp_sum(dparray x1):
...
+
+    if input.dtype == numpy.int32:
+        res_type = numpy.int64
+    else:
+        res_type = input.dtype
+
+    if isinstance(axis, int):
+        axis_ = tuple([axis])
+    else:
+        axis_ = axis
```
shssf, line 245: Perhaps it is better to use https://github.com/IntelPython/dpnp/blob/e3d18674e64ba2c65c47fea0e3c4d0346898d532/dpnp/dpnp_utils.pyx#L261

dpnp/backend_mathematical.pyx
```diff
@@ -221,7 +221,91 @@ cpdef dpnp_sum(dparray x1):
...
+    output_shape = dparray(len(shape_input) - len(axis_), dtype=numpy.int64)
+    ind = 0
+    for id, shape_axis in enumerate(shape_input):
+        if id not in axis_:
+            output_shape[ind] = shape_axis
+            ind += 1
+
+    cdef long prod = 1
+    for i in range(len(output_shape)):
+        if output_shape[i] != 0:
```
shssf, line 256: Why this check is needed? If we have `0` in dimension - no memory created for this input array.
Also, Please create (or use if it exists) a function to do such shape calculation somewhere in https://github.com/IntelPython/dpnp/blob/e3d18674e64ba2c65c47fea0e3c4d0346898d532/dpnp/dpnp_utils.pyx#L157

dpnp/backend_mathematical.pyx
```diff
@@ -221,7 +221,91 @@ cpdef dpnp_sum(dparray x1):
...
+        acc *= shape_input[ind]
+
+    output_shape_offsets = [None] * len(shape_input)
+    acc = 1
+
+    for i in range(len(output_shape)):
+        ind = len(output_shape) - 1 - i
+        output_shape_offsets[ind] = acc
+        acc *= output_shape[ind]
+        result_offsets = input_shape_offsets[:]  # need copy. not a reference
```
shssf, line 275: Please create a function in `utils` to calculate shape offsets

shssf (changes requested):
These are not full but enough comments at this time

tests/skipped_tests.tbl
Alexander-Makaryev: I already explained to @Rubtsowa why it happens. Previously axis parameter was in kwargs. There is a condition
https://github.com/IntelPython/dpnp/blob/e3d18674e64ba2c65c47fea0e3c4d0346898d532/dpnp/dpnp_iface_mathematical.py#L640-L643
so when axis parameter was defined we had a fallback to original Numpy. That is why previously tests where passed.

dpnp/dpnp_iface_mathematical.py
```diff
@@ -637,7 +637,9 @@ def sum(x1, axis=None, **kwargs):
 
     is_x1_dparray = isinstance(x1, dparray)
 
-    if (not use_origin_backend(x1) and is_x1_dparray and not kwargs):
+    if (not use_origin_backend(x1) and is_x1_dparray and len(kwargs) <= 1):
+        axis = kwargs.pop('axis') if len(kwargs) == 1 else None
```
shssf, line 641: Just to beautify this line of code
https://docs.python.org/3/library/stdtypes.html#dict.pop

shssf: Perhaps https://docs.python.org/3/library/stdtypes.html#dict.get is applicable here

shssf (approved):

PR #196, opened 2020-10-27
SORT enable tests

densmirn:


dpnp/dpnp_iface_sorting.py
```diff
@@ -124,7 +124,7 @@ def argsort(in_array1, axis=-1, kind=None, order=None):
     return numpy.argsort(in_array1, axis, kind, order)
 
 
-def sort(in_array1, axis=-1, kind=None, order=None):
+def sort(a, axis=-1, kind=None, order=None):
```
shssf, line 127: please rename input array variable to something more than one letter. `x1` or `input`

shssf: also, it might be better to move all named parameters to kwargs and use it

densmirn: Ok, I will pack the input parameters to `kwargs` and check `kwargs` instead of checking each input parameter. However if a user pass default value for some parameter, e.g. `np.sort(arr, axis=-1)`, we fallback to `numpy`, but able to execute the case by `dpnp`.

PR #198, opened 2020-10-28
build on windows part 1

Alexander-Makaryev:
Previous part of changes was done during adding of SDL flags in https://github.com/IntelPython/dpnp/pull/178.
For now we work with activation of oneAPI beta10 environment.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -164,9 +164,9 @@ if IS_WIN:
     _project_cmplr_flag_lib = ['/DLL']
     _project_cmplr_macro = [("_WIN", "1")]
     _project_rpath = []
-    # TODO obtain setuptools.compiler.buildline options line and replace /MD with /MT instead adding it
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -290,7 +293,7 @@
             ],
             "include_dirs": _cmplr_include + _mathlib_include + _project_backend_dir + _dpctrl_include,
             "library_dirs": _mathlib_path + _omp_libpath + _dpctrl_libpath,
-            "runtime_library_dirs": _project_rpath + _mathlib_rpath + _cmplr_rpath + _omp_rpath + _dpctrl_libpath,
+            "runtime_library_dirs": _runtime_library_dirs,
```
shssf, line 296: We can remove all these runtime library paths for all configurations

Alexander-Makaryev: done

PR #206, opened 2020-11-03
TC test old compiler

Alexander-Makaryev:


shssf:
@Alexander-Makaryev Should we close it?

Alexander-Makaryev:
> @Alexander-Makaryev Should we close it?

I thought to leave it while we didn't finish with release...
But, yes, I will close it.

PR #208, opened 2020-11-06
Windows: linking with dpcpp, build scripts

Alexander-Makaryev:


utils/command_build_clib.py
```diff
@@ -116,21 +117,37 @@ def build_libraries(self, libraries):
...
-                                              output_dir=self.build_clib,
-                                              libraries=libraries,
-                                              library_dirs=library_dirs,
-                                              runtime_library_dirs=runtime_library_dirs,
-                                              extra_preargs=extra_preargs + extra_link_preargs,
-                                              extra_postargs=extra_link_postargs,
-                                              debug=self.debug,
-                                              build_temp=self.build_temp,
-                                              target_lang=language)
+                # TODO very brute way, need to refactor
```
shssf, line 120: it looks like `setuptools` is useless on Windows.
Perhaps it is better to move to cmake (at least for C++ backend). I mean to create CMakeFile in the `backend` directory and call it from `setup.py`

PR #215, opened 2020-11-10
Improve API Reference

densmirn:


shssf:
@densmirn some conflicts should be resolved

PR #219, opened 2020-11-11
fix min

Rubtsowa:


dpnp/backend/custom_kernels_statistics.cpp
```diff
@@ -324,10 +324,13 @@ void custom_min_c(void* array1_in, void* result1, const size_t* shape, size_t nd
 
         size_t output_shape_offsets[res_ndim];
         acc = 1;
-        for (size_t i = res_ndim - 1; i > 0; --i)
+        if (res_ndim > 0)
```
densmirn, line 327: Does new code really work in other way than old one? It seems to me we go through the loop if `res_ndim > 1` in both new and old codes. Am I incorrect?

PR #235, opened 2020-11-13
DOCS fix building backend API reference

densmirn:


shssf:
it looks like path hardcode moved from "public" place to some "hidden" place inside "github" directory.
So, it brings more complexity and less scalability.
Perhaps, it needs to be addressed later.

shssf (approved):

PR #247, opened 2020-11-16
add __sycl_usm_array_interface__ support

oleksandr-pavlyk:
```
In [2]: X = dpnp.arange(10)

In [3]: X.__sycl_usm_array_interface__
Out[3]:
{'data': (94447986515968, False),
 'shape': (10,),
 'strides': None,
 'typestr': '=f8',
 'version': 1,
 'syclobj': <dpctl._sycl_core.SyclQueue at 0x7f148de5fb90>}
```

Hand off memory to `dpctl.memory.MemoryUSMShared` object:

```
In [4]: ms = dpctl.memory.MemoryUSMShared(X)

In [5]: ms.copy_to_host()
Out[5]:
array([  0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,
         0, 240,  63,   0,   0,   0,   0,   0,   0,   0,  64,   0,   0,
         0,   0,   0,   0,   8,  64,   0,   0,   0,   0,   0,   0,  16,
        64,   0,   0,   0,   0,   0,   0,  20,  64,   0,   0,   0,   0,
         0,   0,  24,  64,   0,   0,   0,   0,   0,   0,  28,  64,   0,
         0,   0,   0,   0,   0,  32,  64,   0,   0,   0,   0,   0,   0,
        34,  64], dtype=uint8)

In [6]: _.view(X.dtype)
Out[6]: array([0., 1., 2., 3., 4., 5., 6., 7., 8., 9.])
```

Create a new `dparray`, hand it off to another `MemoryUSMShared` object, and use `copy_from_device` to copy from `ms2` to `ms`.

```
In [7]: X2 = dpnp.random.randn(10)

In [8]: ms2 = dpctl.memory.MemoryUSMShared(X2)

In [9]: ms.copy_from_device(ms2)
```

Verify that both X and X2 hold the same values:

```
In [10]: print(X)
-0.3086549527322641 0.47275601111442156 0.33836205748412956 0.7102158308134265 -0.6514968520151948 -1.5033187336729206 -0.22150961174533537 1.2182308804007054 0.5806215114059609 -0.7123286786650291 <__str__ TODO>

In [11]: print(X2)
-0.3086549527322641 0.47275601111442156 0.33836205748412956 0.7102158308134265 -0.6514968520151948 -1.5033187336729206 -0.22150961174533537 1.2182308804007054 0.5806215114059609 -0.7123286786650291 <__str__ TODO>
```


dpnp/dparray.pyx
```diff
@@ -281,6 +372,26 @@ cdef class dparray:
 
         return interface_dict
 
+
+    @property
+    def __sycl_usm_array_interface__(self):
+        cdef dict iface = {
```
shssf, line 378: why you use Cython specific things in pure Python procedure? It will requires extra data conversion.

dpnp/dparray.pyx
```diff
@@ -281,6 +372,26 @@ cdef class dparray:
 
         return interface_dict
 
+
+    @property
+    def __sycl_usm_array_interface__(self):
```
shssf, line 377: This is a code duplication to `__array_interface__`. I see one field diffrence only. It looks like it is better to add this field into `__array_interface__`

dpnp/dparray.pxd
```diff
@@ -32,7 +32,7 @@ Represents multi-dimensional array using USM interface for an Intel GPU device.
 
 
 from libcpp.vector cimport vector
-
+from dpctl cimport SyclQueue
```
shssf, line 35: What happened if no `dpctl` exists?

dpnp/dparray.pyx
```diff
@@ -113,17 +193,27 @@ cdef class dparray:
             self._dparray_strides.push_back(strides_it)
 
         # data
-        if memptr is None:
+        if memory is None:
             self._dparray_data = dpnp_memory_alloc_c(self.nbytes)
+            self.queue = c_dpctl.get_current_queue()
```
shssf, line 198: DPNP localized all queue related things inside the backend. Python and Cython layers has no relations to SYCL queue

shssf: move this to backend or use existed backend interface

dpnp/dparray.pxd
oleksandr-pavlyk: This change makes `dpctl` a hard build-time dependency of `dpnp`.

setup.py
```diff
@@ -312,7 +312,7 @@
     name="dpnp.backend",
     sources=["dpnp/backend.pyx"],
     libraries=[],
-    include_dirs=[numpy.get_include()] + _project_backend_dir,
+    include_dirs=[numpy.get_include()] + _project_backend_dir + [dpctl.get_include()],
```
shssf, line 315: use existed variable instead direct call

dpnp/dparray.pyx
```diff
@@ -281,6 +372,26 @@ cdef class dparray:
 
         return interface_dict
 
+
+    @property
+    def __sycl_usm_array_interface__(self):
+        cdef dict iface = {
+            "data": (<Py_ssize_t>(<void *>self._dparray_data),
+                     False),
```
shssf, line 380: put it at the same line above

dpnp/dparray.pyx
oleksandr-pavlyk: Yes, it does. When `dpctl` changes the default queue, the `sycl::get_pointer_type( _dparray_data, DPNP_CURRENT_QUEUE().get_context())`  may return `"uknown"`, having the `queue` onto which the memory was allocated at least would allow `dpnp` to access the data.

dpnp/dparray.pyx
```diff
@@ -45,19 +45,98 @@ import numpy
...
+
+    @staticmethod
+    cdef _ArrayInterfaceData parse_sycl_usm_array_interface(object memory):
+        cdef dict iface
+        cdef _ArrayInterfaceData res_iface
+        cdef sycl_ctx
+        cdef char * iface_pointer
+        
+        if not hasattr(memory, "__sycl_usm_array_interface__"):
+            raise TypeError("Memory object does not expose __sycl_usm_array_interface__")
```
shssf, line 70: use predefined functions from `utility` to raise exceptions

dpnp/dparray.pyx
```diff
@@ -45,19 +45,98 @@ import numpy
...
+        else:
+            # Obtain device from pointer and context
+            sycl_ctx = <c_dpctl.SyclContext> syclobj
+            sycl_dev = c_dpctl_mem._Memory.get_pointer_device(<c_dpctl.DPPLSyclUSMRef>iface_pointer, ctx)
+            usm_type = c_dpctl_mem._Memory.get_pointer_type(<c_dpctl.DPPLSyclUSMRef>iface_pointer, ctx)
+            if (usm_type != b'shared'):
+                raise TypeError("Expecting USM shared memory.")
+            # Use context and device to create a queue to
+            # be able to copy memory
+            iface_queue = c_dpcl.SyclQueue._create_from_context_and_device(sycl_ctx, sycl_dev)
```
shssf, line 115: it is not quite clear why you need to `create` a queue. Use the `queue` to copy data.

dpnp/dparray.pyx
```diff
@@ -77,9 +156,10 @@ cdef class dparray:
 
     """
 
-    def __init__(self, shape, dtype=float64, memptr=None, strides=None, order=b'C'):
+    def __init__(self, shape, dtype=float64, memory=None, strides=None, order=b'C'):
```
shssf, line 159: it looks like it is not the best name for this parameter. This is not a memory or memory pointer.
It looks like it is `external object`
Do you have any examples in other API where such functionality used? Perhaps it is better to be aligned with names,

dpnp/dparray.pyx
oleksandr-pavlyk: This is the case when `__sycl_usm_array_interface__` is used to shared data allocated without a queue, using `sycl::malloc(size_t n, sycl::context, sycl::device, sycl::usm::kind)`, and only records the `context`.

We then use `sycl::get_pointer_device(ptr, ctx)` to look up the device on which memory was allocated, and need to create a queue against which operations on this memory can be performed.

dpnp/dparray.pyx
oleksandr-pavlyk: In NumPy, the keyword is call `buffer=`. The buffer needs to be a byte-like object. NumPy would not allow an arbitrary object to be used in its place even if that object implements `__array_interface__`. 

NumPy can consume duck arrays that advertise themselves as arrays only via `__array_interface__` property only via `numpy.asarray`, or `numpy.asanyarray`.

dpnp/dparray.pyx
oleksandr-pavlyk: No, this is implementation of a different mechanism, which should not be confused with `__array_interface__`. 

It also serves as a signal that an object `pyobj` can export USM memory, detectable via `hasattr(pyobj, "__sycl_usm_array_interface__")`.

dpnp/dparray.pyx
oleksandr-pavlyk: The only free-form exceptions there is for run-time error. These are all `TypeError`s. 
The `checker_throw_type_error` is not sufficiently flexible in wording. 

Is the intent of suggestion to conform to a uniform style of exception reporting?

dpnp/dparray.pyx
oleksandr-pavlyk: I removed it.

oleksandr-pavlyk:
I am going to merge this to unblock numba/dpctl work, and address outstanding issues (supporting objects exporting the interface in `dpnp.array`) in separate PRs

PR #248, opened 2020-11-17
ENH: added `random.standard_normal`

samir-nasibli:
## Description
standard_normal([size]) Draw samples from a standard Normal distribution (mean=0, stdev=1).

## TODO
* after merge of this PR, `randn` will be redesigned and `normal` will be added
* is it needed `standard_normal` on backend? We just can call `normal` with params (mean=0, stdev=1) from `standard_normal` on python level.
## Reproduce

```python
>>> np.mean(dpnp.random.standard_normal(10**6));np.var(dpnp.random.standard_normal(10**6))
0.00030427453642504083
1.0013896813115708
>>> np.mean(np.random.standard_normal(10**6));np.var(np.random.standard_normal(10**6))
0.00011309454634553527
1.0003356224869078
```

## Checklist

- [x] Unit tests:

### DIFF
--- a/dpnp/backend.pxd
+++ b/dpnp/backend.pxd
@@ -108,6 +108,7 @@ cdef extern from "backend/backend_iface_fptr.hpp" namespace "DPNPFuncName":  # n
         DPNP_FN_SQRT
         DPNP_FN_SQUARE
         DPNP_FN_STANDARD_CAUCHY
+        DPNP_FN_STANDARD_NORMAL
         DPNP_FN_STD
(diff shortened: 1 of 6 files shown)


samir-nasibli:
@shssf  * is it needed `standard_normal` on backend? We just can call `normal` with params (mean=0, stdev=1) from `standard_normal` on python level. The same question for `standard_gamma`, `standard_exponential` and etc.

shssf (approved):
No need to reimplement this at this time. Let it be as you already implemented.

PR #250, opened 2020-11-17
Add docstring template with limitations

densmirn:


doc/docstring_template.py
```diff
@@ -0,0 +1,76 @@
...
+    out : dparray
+        Description of the output.
+
+    Limitations
+    -----------
+    Some limitations in comparison to baseline, ex:
+    Prameters `c` is supported only with default value `True`.
+    Otherwise the functions will be executed sequentially on CPU.
+
+    See Also
```
shssf, line 57: Do we have an option to keep the same format but with less lines?
I mean, this section format takes 2 lines. Perhaps we could use something like "... see also" or something like this?

densmirn: Yes, we are able to use both formats.

Issue #89, opened 2020-09-29
Build system. Source spell check


shssf:
Need to implement a command to spell check of the text in sources.
- might looks like `python ./setup.py spell`

antonwolfy:
The spell check was implemented as a pre-commit hook in #1798

PR #265, opened 2020-11-18
Asanyarray ascontiguousarray asfarray copy funcs

Alexander-Makaryev:
+21 passed cupy tests
+15 in tests/test_arraymanipulation.py


dpnp/backend/custom_kernels_random.cpp
```diff
@@ -187,7 +187,8 @@ void func_map_init_random(func_map_t& fmap)
     fmap[DPNPFuncName::DPNP_FN_GAUSSIAN][eft_DBL][eft_DBL] = {eft_DBL, (void*)custom_rng_gaussian_c<double>};
     fmap[DPNPFuncName::DPNP_FN_GAUSSIAN][eft_FLT][eft_FLT] = {eft_FLT, (void*)custom_rng_gaussian_c<float>};
 
-    fmap[DPNPFuncName::DPNP_FN_NEGATIVE_BINOMIAL][eft_INT][eft_INT] = {eft_INT, (void*)custom_rng_negative_binomial_c<int>};
+    fmap[DPNPFuncName::DPNP_FN_NEGATIVE_BINOMIAL][eft_INT][eft_INT] = {eft_INT,
```
oleksandr-pavlyk, line 190: Does not look like these changes belong to this PR.

dpnp/dparray.pyx
```diff
@@ -47,6 +47,47 @@ cimport numpy
 cimport dpnp.dpnp_utils as utils
 
 
+# initially copyed from original
```
oleksandr-pavlyk, line 50: copyed->copied

dpnp/dparray.pyx
```diff
@@ -47,6 +47,47 @@ cimport numpy
...
+cdef class _flagsobj:
```
shssf, line 51: I don not understand why we need this. It is better to reuse this from NumPy

dpnp/dparray.pyx
Alexander-Makaryev: done

dpnp/backend/custom_kernels_random.cpp
Alexander-Makaryev: Yes, it is a result of applying of code style tool... The correct way is to apply code style tool during work on changes, but if someone missed it, then these changes occur when I apply code style. Sometimes it is hard to split these changes in new and old code. We should improve our CI to automatic check/apply of code style.

dpnp/dparray.pyx
Alexander-Makaryev: We need to get own behavior while we have no a full enough support of features like f-contiguous memory layout, strides, views (sharing the memory) etc. In fact at this moment we just return a constant values of flags and it will be easier to modify our own class "step by step". We can replace it with original Numpy class in future, when it will be possible to reuse original functionality. Possibly I am wrong)

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -171,14 +175,64 @@ def array(obj, dtype=None, copy=True, order='C', subok=False, ndmin=0):
...
-        checker_throw_value_error("array", "order", order, 'K')
+        checker_throw_value_error("array", "order", order, 'C')
 
     if ndmin != 0:
         checker_throw_value_error("array", "ndmin", ndmin, 0)
 
     return dpnp_array(obj, dtype)
 
 
+def asanyarray(a, dtype=None, order=None):
```
densmirn, line 186: Could you please add block `Limitations` to all the docstring something like this:
```
    Limitations
    -----------
    Parameter ``order`` is supported only with default value `'C'`.
```

dpnp/dparray.pyx
shssf: @Alexander-Makaryev it looks like it is related to #89

dpnp/dparray.pyx
shssf: @Alexander-Makaryev We "return a constant values" because we support one data state only at this time.
It is questionable we will support diffrent states.
Using Numpy Flags is also possible in "one state"- just construct it based on "constant values".
I think this is ambiguous code what is better to handle via "fallback" because it is not related to SYCL

shssf (approved):

PR #268, opened 2020-11-18
DOCS align signatures of dpnp.arange and numpy.arange

densmirn:


dpnp/dpnp_iface_arraycreation.py
```diff
@@ -73,48 +73,64 @@ def arange(*args, **kwargs):
...
+            pass
+        else:
+            if dtype is None:
+                dtype = numpy.float64
+
+            if stop is None:
+                stop = start
+                start = 0
+
+            if step is None:
```
shssf, line 128: What will happend if `arange(100,10)`?
It will described as `stop=100, step=10` in numpy

densmirn: In case of `arange(100,10)` `start=100, stop=10, step=1` as it was before. Results of `numpy` and `dpnp` are the same:
```
>>> import numpy
>>> import dpnp as np
>>> numpy.arange(100, 10)
array([], dtype=int64)
>>> print(np.arange(100, 10))
<__str__ TODO>
```

shssf: it is really strange. Perhaps I cat not find good example for this. as far as I remember, the function `arange` is one of the special function in case of parameters handling. It is not a simply usage of `argc, kwargs`. This signature was implemented intentionally because first parameter is optional.

Are you sure this works properly and all tests non skipped for this?

densmirn: Unskipped all the tests on `arange` and they are passed. BTW the implementation is similar to `cupy` implementation https://github.com/cupy/cupy/blob/v8.1.0/cupy/_creation/ranges.py#L9.

shssf (approved):

PR #270, opened 2020-11-19
REF: random functions in alphabet order

samir-nasibli:
## Description
* changed functions order

### DIFF
--- a/dpnp/random/_random.pyx
+++ b/dpnp/random/_random.pyx
@@ -300,47 +300,6 @@ cpdef dparray dpnp_gumbel(double loc, double scale, size):
     return result
 
 
-cpdef dparray dpnp_negative_binomial(double a, double p, size):
-    """
-    Returns an array populated with samples from negative binomial distribution.
-
-    `negative_binomial` generates a matrix filled with random floats sampled from a
-    univariate negative binomial distribution for a given parameter of the distribution
-    `a` and success probability `p` of a single trial.
-
-    """
-
-    dtype = numpy.int32
-    cdef dparray result
-    cdef DPNPFuncType param1_type
-    cdef DPNPFuncData kernel_data
-    cdef fptr_custom_rng_negative_binomial_c_1out_t func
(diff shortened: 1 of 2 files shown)


dpnp/random/_random.pyx
```diff
@@ -300,114 +300,114 @@ cpdef dparray dpnp_gumbel(double loc, double scale, size):
...
-    univariate negative binomial distribution for a given parameter of the distribution
-    `a` and success probability `p` of a single trial.
+    Returns an array populated with samples from beta distribution.
+    `dpnp_laplace` generates a matrix filled with random floats sampled from a
+    univariate laplace distribution.
 
     """
 
-    dtype = numpy.int32
+    dtype = numpy.float64
```
shssf, line 311: @samir-nasibli Please use DPNP specific types as much as possible
https://github.com/IntelPython/dpnp/blob/4664aac456d8d20f6f115292ea1ccabf3af403d9/dpnp/dpnp_iface_types.py#L46

shssf: I see other same issues below

dpnp/random/dpnp_iface_random.py
```diff
@@ -496,6 +437,65 @@ def gamma(shape, scale=1.0, size=None):
     return call_origin(numpy.random.gamma, shape, scale, size)
 
 
+def geometric(p, size=None):
```
shssf, line 440: it is not good idea to mix "refactoring" with "development".
No changes needed. Please separate such things in future.

samir-nasibli: There is no "development" here. I just changed places of some functions. Just Github is going crazy. There is only refactoring here. The refactoring idea is reflected in the PR title.

dpnp/random/_random.pyx
samir-nasibli: If it is critical, I can change it on other PR.

PR #278, opened 2020-11-19
DOCS align docstring for full with template

densmirn:


dpnp/dpnp_iface_arraycreation.py
```diff
@@ -381,60 +379,52 @@ def full(shape, fill_value, dtype=None, order='C'):
 
 
 # numpy.full_like(a, fill_value, dtype=None, order='K', subok=True, shape=None)
-def full_like(prototype, fill_value, dtype=None, order='C', subok=False, shape=None):
+def full_like(a, fill_value, dtype=None, order='C', subok=False, shape=None):
```
shssf, line 382: Please do not use such short names from data structures. Please use at least 2 symbols for it. `x1` is good candidate

PR #281, opened 2020-11-20
ENH: added `random.bytes`

samir-nasibli:
## Description
Not in our EOY Goal

## TODO
* fixes
* tests
* `tobytes`


samir-nasibli:
@shalnov It is ready for merge, if we need it as is, or we can just simply close it. We don't have this feature on our plan for this year. We don't have also `tobytes` on our plan.

PR #286, opened 2020-11-20
Indexing funcs nonzero

Alexander-Makaryev:
source file for indexing routines
nonzero function implementation
+3 passed tests

### DIFF
--- a/dpnp/dpnp_iface_indexing.py
+++ b/dpnp/dpnp_iface_indexing.py
@@ -1,0 +1,65 @@
+# cython: language_level=3
+# distutils: language = c++
+# -*- coding: utf-8 -*-
+# *****************************************************************************
+# Copyright (c) 2016-2020, Intel Corporation
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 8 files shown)


dpnp/dpnp_iface_indexing.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Notes
+    -----
+    While the nonzero values can be obtained with ``a[nonzero(a)]``, it is
+    recommended to use ``x[x.astype(bool)]`` or ``x[x != 0]`` instead, which
+    will correctly handle 0-d arrays.
+
+    """
+
+    is_a_dparray = isinstance(a, dparray)
```
densmirn, line 105: Please add block `Limitations` to the docstring, something like this:
```
    Limitations
    -----------
        Parameter ``a`` is supported as :obj:`dpnp.ndarray`.
        Otherwise the function will be executed sequentially on CPU.
```

densmirn (commented):
It would be better to have some example in the docstring.

shssf (approved):

PR #287, opened 2020-11-20
ENH: added `random.standard_exponential`

samir-nasibli:
## Description
standard_exponential([size]) Draw samples from the standard exponential distribution.

## Reproduce
```python
>>> np.mean(np.random.standard_exponential(10**6));np.mean(dpnp.random.standard_exponential(10**6))
1.000325845773389
1.0007827968460972
>>> np.var(np.random.standard_exponential(10**6));np.var(dpnp.random.standard_exponential(10**6))
0.9963301885655749
0.9958583407421286

```

## Checklist

- [x] Unit tests:


dpnp/random/dpnp_iface_random.py
```diff
@@ -1243,6 +1244,26 @@ def standard_cauchy(size=None):
     return call_origin(numpy.random.standard_cauchy, size)
 
 
+def standard_exponential(size=None):
+    """Standard exponential distribution.
+
+    """
```
densmirn, line 1250: Please write full docstring for the function.

shssf: @samir-nasibli I will merge this (to unblock you) but please do not forget to put correct docstring into the function as @densmirn recommended.

PR #300, opened 2020-11-24
DEP: update `random.seed` exception handler

samir-nasibli:
deprecate using dpnp own exception handlers/raise

### DIFF
--- a/dpnp/random/dpnp_iface_random.py
+++ b/dpnp/random/dpnp_iface_random.py
@@ -1406,18 +1406,19 @@ def seed(seed=None):
     seed : {None, int}, optional
 
     """
+
     if not use_origin_backend(seed):
-        # TODO:
-        # implement seed default value as is in numpy
-        if seed is None:
-            seed = 1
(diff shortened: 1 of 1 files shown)


dpnp/random/dpnp_iface_random.py
```diff
@@ -1406,20 +1406,21 @@ def seed(seed=None):
...
+            pass
+        else:
+            if seed is None:
+                seed = 1
+            dpnp_srand(seed)
 
     return call_origin(numpy.random.seed, seed)
 
 
+
```
densmirn, line 1423: Excess blank line - https://www.python.org/dev/peps/pep-0008/#blank-lines.

densmirn (commented):
Let's write full docstring (with limitations block) for the function, maybe in separate PR.

samir-nasibli:
> Let's write full docstring (with limitations block) for the function, maybe in separate PR.

This is DEP PR with a specific purpose. All other manipulations mentioned by you will be done in another PRs.

PR #276, opened 2020-11-19
ENH: added `random.multinomial`

samir-nasibli:
## Description
multinomial(n, pvals[, size]) Draw samples from a multinomial distribution.

## TODO
* update docstrings: add limitations (block `limitations` for this function and for other random module funcs will be added later on other PRs).
* fallbacks to numpy will be redesigned on other PR

## Reproduce

```python
>>> pvals = [1 / 6.] * 6
>>> n = 20
>>> np.var(dpnp.random.multinomial(n, pvals, size=10**5));np.var(np.random.multinomial(n, pvals, size=10**5))
2.7855555555563893
2.7705988888888893
>>> np.mean(dpnp.random.multinomial(n, pvals, size=10**5));np.mean(np.random.multinomial(n, pvals, size=10**5))
3.3333333333333335
3.3333333333333335

```

## Checklist

- [x] Docstrings for all functions
- [ ] Gallery example in ./doc/examples (new features only)
- [x] Unit tests:
- [x] Clean style in [the spirit of PEP8](https://www.python.org/dev/peps/pep-0008/)

PR #308, opened 2020-11-25
API: changed `random.multinomial` backend function api

samir-nasibli:
## Description
* using 'simple' interface instead of STL. ( `const double* p_vector, const size_t p_vector_size` instead of `std::vector<double>& p_vector`) -> `random.multinomial` backend function (`custom_rng_multinomial_c` signature was changed)
>Perhaps it might be fixed later.
This is a not quite good to use STL in the functions interface. These containers are good for regular usage but in our case we need to provide "simple" interface. Usually we use Cython or Numba IR as a customer for this intrface. Importing C++ STL is a quite big pain for such usage model.
Please use const double* p_vector, size_t p_vector_size instead. @shssf . See #276

### DIFF
--- a/dpnp/backend/backend_iface.hpp
+++ b/dpnp/backend/backend_iface.hpp
@@ -817,17 +817,23 @@ INP_DLLEXPORT void custom_rng_lognormal_c(void* result, _DataType mean, _DataTyp
  * @ingroup BACKEND_API
  * @brief math library implementation of random number generator (multinomial distribution)
  *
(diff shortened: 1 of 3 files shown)


samir-nasibli:
@shssf currently we do several memcpy: python list -> numpy array (c-contiguous array) on cython level -> std::vector on backend.
If it is bad, we can pass std::vector by a void pointer to the backend, and cast it there: python list -> to std::vector on cython level and passing pointer to the backend with no copy.

PR #314, opened 2020-11-27
DOCS align docstring for trigonometric funcs with template

densmirn:


dpnp/backend_trigonometric.pyx
```diff
@@ -177,6 +177,7 @@ cpdef dparray dpnp_unwrap(dparray array1):
         result = dparray(array1.shape)
 
     for i in range(result.size):
-        result[i] = numpy.unwrap(array1[i])
+        val, = numpy.unwrap([array1[i]])
+        result[i] = val
```
samir-nasibli, line 181: it is not good idea to mix "DOC docstrings align" with "development".
No changes needed in this PR. Please, separate such things in future.

densmirn: @samir-nasibli without the change the example in docstring doesn't work.

samir-nasibli: @densmirn therefore you need to make a separate PR to change the behavior of the function. In this case it is not so critical. Leave it as is. 
My comment was just an advice for the future.

densmirn: @samir-nasibli In general you are right, but in this case I prefer to keep this small fix to this PR if nobody minds.

PR #336, opened 2020-12-02
atleast_1d, atleast_2d, atleast_3d

Rubtsowa:


dpnp/dpnp_iface_manipulation.py
```diff
@@ -60,6 +63,107 @@
...
+    ret : ndarray
+        An array, or list of arrays, each with ``a.ndim >= 1``.
+        Copies are made only if necessary.
+    """
+
+    return call_origin(numpy.atleast_1d, *arys)
+
+
+def atleast_2d(*arys):
+    """
```
shssf, line 88: As far as I understand we have to follow new documentation strings example.
Perhaps it is not the best example but you could see example in https://github.com/IntelPython/dpnp/blob/33013be33accfb0898b3e27c5b28a1e8333ef27b/dpnp/fft/dpnp_iface_fft.py#L68-L77

PR #382, opened 2020-12-08
add tril_indices

Rubtsowa:


dpnp/dpnp_iface_indexing.py
```diff
@@ -198,3 +199,42 @@ def nonzero(a):
...
+    -------
+    inds : tuple of arrays
+        The indices for the triangle. The returned tuple contains two arrays,
+        each with the indices along one dimension of the array.
+
+    """
+
+    if not use_origin_backend():
+        if not isinstance(n, int):
+            raise TypeError("n must be an integer")
```
shssf, line 232: Could you please change this checks a bit.
It is better to reuse fallback for any exception or error handling.
Example of this might be here https://github.com/IntelPython/dpnp/blob/836fe4cd601cb07e785c29a18f3f3f3f1eb54634/dpnp/fft/dpnp_iface_fft.py#L100-L109

PR #387, opened 2020-12-09
add triu_indices()

Rubtsowa:


dpnp/dpnp_iface_indexing.py
```diff
@@ -258,3 +259,37 @@ def tril_indices_from(arr, k=0):
...
+    Returns
+    -------
+    inds : tuple, shape(2) of ndarrays, shape(`n`)
+        The indices for the triangle. The returned tuple contains two arrays,
+        each with the indices along one dimension of the array.  Can be used
+        to slice a ndarray of shape(`n`, `n`).
+    """
+
+    if not use_origin_backend():
+        if isinstance(n, int) and isinstance(k, int) \
```
samir-nasibli, line 291: This will not fallback in otherwise. Please, redesign it. Some example from here:
https://github.com/IntelPython/dpnp/blob/master/dpnp/random/dpnp_iface_random.py#L173:L188

PR #411, opened 2020-12-11
TESTS enable numpy.core tests run

densmirn:
1359 passed, 3043 skipped, 9 xfailed, 1 warning in 18.23s


dpnp/dparray.pyx
```diff
@@ -48,7 +48,7 @@ cimport dpnp.dpnp_utils as utils
 
 
 # initially copied from original
-cdef class _flagsobj:
+class _flagsobj:
```
samir-nasibli, line 51: Do we really need changes related with dparray here? wouldn't it be better to move it to a separate PR, that could be merged before this one.

dpnp/dpnp_iface_trigonometric.py
```diff
@@ -663,14 +662,13 @@ def log(x1):
     [0.0, 1.0, 2.0, -inf]
 
     """
+    if use_origin_backend(x1):
+        if not isinstance(x1, dparray):
+            pass
```
samir-nasibli, line 667: It is better to move this fixes into other PR. Or provide, please, full description for this PR.

dpnp/dparray.pyx
```diff
@@ -531,6 +549,14 @@ cdef class dparray:
         if order is not b'C':
             utils.checker_throw_value_error("dparray::reshape", "order", order, b'C')
 
+        if dn:
```
samir-nasibli, line 552: This additions apply serious changes on our API and may lead big difference from the original `numpy`. Please, move this changes into other PR and add request review also from O. Pavlyk.

shssf:
@densmirn It is looks what we need. Need to continue this work.
Also, I would propose to change subdirectory name from `tests/third_party/numpy` to `tests/third_party/numpy_ext` to have a chance in future to distinguish tests used from external package internally (like cupy now) and direct use of external test system.

It is also possible (optional) to move such test runner outside of DPNP tests to something like `dpnp/tests_external`

densmirn:
> @densmirn It is looks what we need. Need to continue this work.
> Also, I would propose to change subdirectory name from `tests/third_party/numpy` to `tests/third_party/numpy_ext` to have a chance in future to distinguish tests used from external package internally (like cupy now) and direct use of external test system.
> 
> It is also possible (optional) to move such test runner outside of DPNP tests to something like `dpnp/tests_external/numpy`

@shssf how do you think it makes sense to move functionality changes to separate PRs or not? Anyway all the tests are passed.

shssf:
@densmirn I think all things are ok but let me sometime to review it. No need to change this PR at this time.

shssf (approved):

samir-nasibli:
@shssf @densmirn I vote to split this PR into logically smaller ones and request review for some of them from @oleksandr-pavlyk

PR #412, opened 2020-12-11
add diff

Rubtsowa:


dpnp/backend_mathematical.pyx
```diff
@@ -154,6 +155,26 @@ cpdef dparray dpnp_cumsum(dparray x1):
     return result
 
 
+cpdef dparray dpnp_diff(dparray input):
+    size_i = input.size
```
Alexander-Makaryev, line 159: this variable used only once below, why you don't just use input.size instead?

tests/test_mathematical.py
```diff
@@ -0,0 +1,26 @@
...
+                              '[[1, 2], [1, 2]]',
+                              '[[1, 2], [3, 4]]',
+                              '[[[1, 2], [3, 4]], [[1, 2], [2, 1]], [[1, 3], [3, 1]]]',
+                              '[[[[1, 2], [3, 4]], [[1, 2], [2, 1]]], [[[1, 3], [3, 1]], [[0, 1], [1, 3]]]]'
+                              ])
+def test_diff(array):
+    a = numpy.array(array)
+    ia = inp.array(a)
+    result = inp.linalg.det(ia)
+    expected = numpy.linalg.det(a)
```
Alexander-Makaryev, line 25: det?) looks like a copy-paste typo

dpnp/backend_mathematical.pyx
```diff
@@ -154,6 +155,26 @@ cpdef dparray dpnp_cumsum(dparray x1):
...
+    shape_i = input.shape
```
Alexander-Makaryev, line 160: The same as comment with `size_i` variable.
The new variable with unclear name and purpose is defined. It is hard enough to read what happens with it below in code, at least for me. I suggest to avoid such things.

Alexander-Makaryev (commented):
@Rubtsowa please look at my comments.

tests/test_mathematical.py
Alexander-Makaryev: there are test on diff at math_tests\test_sumprod.py, why they are not used?

PR #413, opened 2020-12-11
DEP: removed deprecated tests from skip list

samir-nasibli:
## Description
* removed unnecessary tests cases from skip list.


samir-nasibli:
@shssf please, merge it before #409 , #406, #400 and  #401

PR #424, opened 2020-12-15
Fix diff

Rubtsowa:


tests/test_mathematical.py
```diff
@@ -21,6 +21,8 @@
 def test_diff(array):
     a = numpy.array(array)
     ia = inp.array(a)
-    result = inp.linalg.det(ia)
-    expected = numpy.linalg.det(a)
+    expected = numpy.diff(a)
+    print('EXPECTED ', expected)
```
densmirn, line 25: Let's remove prints.

dpnp/backend_mathematical.pyx
```diff
@@ -155,23 +155,32 @@ cpdef dparray dpnp_cumsum(dparray x1):
...
-    for i in range(size_arr_):
-        for j in range(size_idx):
-            idx = i * size_idx + j
-            input_elem = input.item(idx) - input.item(idx + 1)
-            res.appen(input_elem)
-
-    dpnp_array = dpnp.array(res)
-    dpnp_result_array = dpnp_array.reshape(output_shape)
+cpdef dparray dpnp_diff(dparray input, int n):
+    if n == 0:
```
densmirn, line 159: You could use the following template for the function:
```python
if n == 0:
    return input

if ...:
    ...
    return arr

return dpnp.array(...)
```

dpnp/dpnp_iface_mathematical.py
```diff
@@ -337,22 +337,24 @@ def diff(input, n=1, axis=-1, prepend=None, append=None):
     Limitations
     -----------
     Input array is supported as :obj:`dpnp.ndarray`.
-    Not supported parameters n, axis, prepend, append.
+    Not supported parameters axis, prepend, append.
```
densmirn, line 340: I would write:
```
Parameters ``axis``, ``prepend`` and ``append`` are supported only with default values.
Otherwise the function will be executed sequentially on CPU.
```
because actually the parameters are supported, but in some cases we fallback to `numpy`.

PR #429, opened 2020-12-17
dparray: item, tobytes, tofile, tolist, tostring

Alexander-Makaryev:


shssf (approved):
`__str__` wil be the most visible function with same approach

PR #432, opened 2020-12-18
add ediff1d function

KsanaKozlova:


dpnp/backend_mathematical.pyx
```diff
@@ -201,6 +202,40 @@ cpdef dparray dpnp_divide(dparray x1, dparray x2):
...
+cpdef dparray dpnp_ediff1d(dparray x1, dparray to_end, dparray to_begin):
+
+    types_map = {
+        'int32': dpnp.int64,
+        'int64': dpnp.int64,
+        'float32': dpnp.float32,
+        'float64': dpnp.float64
+    }
+
+    res_type = types_map[x1.dtype.name]
```
shssf, line 214:
```suggestion
    types_map = {
        dpnp.int32: dpnp.int64,
        dpnp.int64: dpnp.int64,
        dpnp.float32: dpnp.float32,
        dpnp.float64: dpnp.float64
    }

    res_type = types_map[x1.dtype]
```
In this case no need to compare strings - pointers will be compared.
Also, what will happen if `dpnp.complex128` come into function?

densmirn: Looks like `KeyError: <dpnp.complex128>` will be raised in case of `x1.dtype == dpnp.complex128`. Maybe better to raise `TypeError` with `checker_throw_type_error(...)` in this case.

shssf: or change this code in the way to use existed mechanism from backend. It might be a step 2. 
As a first step - it is better to avoid string comparison as much as possible.

shssf (approved):

PR #448, opened 2020-12-28
add trapz function

KsanaKozlova:


shssf (approved):
It would be great to use more than 1 letters in variable names

PR #460, opened 2021-01-04
bdist_wheel on windows

shssf:


scripts/azure-pipelines.yml
```diff
@@ -160,10 +160,14 @@ jobs:
   pool:
     vmImage: 'windows-2019' # 'vs2017-win2016'
   steps:
+  - powershell: |
+      Start-Process -Wait -FilePath  "C:\Program Files (x86)\Microsoft Visual Studio\Installer\vs_installer.exe" -ArgumentList "modify --add Microsoft.VisualStudio.Workload.NativeDesktop --add Microsoft.Component.MSBuild --add Microsoft.VisualStudio.Component.Windows10SDK --add Microsoft.VisualStudio.Component.VC.CoreBuildTools --passive --norestart --installpath ""C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise"""
+
   - script: |
```
AndresGuzman-Ballen, line 166: I'm not familiar with how this yaml file is supposed to be setup, but is there supposed to be only one "- script" and not two? I am trying to understand why CI is failing but can't figure out what could possibly be doing so. I see that cmake complains about "XXFLAG" but I don't see that anywhere when I git cloned this repo and grepped for it. I also know that master branch works, so clearly there is "something" in this PR causing a problem. I am having a hard time pinpointing it though :(

shssf: @AndresGuzman-Ballen Yes, I also confused by this error "XXFLAG" because where is no such variable in the project. It looks like this is the result of some pre-scripting or filtering inside TeamCity.
Also, this particular yml file is not used in TeamCity. This file used in azure pipelines only.

PR #472, opened 2021-01-11
Kernel inv

Rubtsowa:


dpnp/backend/include/dpnp_iface.hpp
```diff
@@ -316,6 +316,22 @@ INP_DLLEXPORT void dpnp_cov_c(void* array1_in, void* result1, size_t nrows, size
...
+ *
+ * @param [out] result  Output array.
+ *
+ * @param [in]  shape   Shape of input array.
+ *
+ * @param [in]  ndim    Number of elements in shape.
+ *
+ */
+template <typename _DataType>
+INP_DLLEXPORT void dpnp_inv_c(void* array1_in, void* result1, size_t* shape);
```
shssf, line 333:
```suggestion
INP_DLLEXPORT void dpnp_inv_c(void* array1_in, void* result1, size_t* shape, size_t ndim);
```

dpnp/backend/kernels/dpnp_krnl_linalg.cpp
```diff
@@ -175,6 +175,104 @@ void dpnp_det_c(void* array1_in, void* result1, size_t* shape, size_t ndim)
     return;
 }
 
+template <typename _DataType>
+class dpnp_inv_c_kernel;
+
+template <typename _DataType>
+void dpnp_inv_c(void* array1_in, void* result1, size_t* shape)
```
shssf, line 182: also `ndim` should be here but will not be used at this time

dpnp/backend/kernels/dpnp_krnl_linalg.cpp
```diff
@@ -175,6 +175,104 @@ void dpnp_det_c(void* array1_in, void* result1, size_t* shape, size_t ndim)
...
+{
+    _DataType* array_1 = reinterpret_cast<_DataType*>(array1_in);
+    _DataType* result = reinterpret_cast<_DataType*>(result1);
+
+    size_t n = shape[0];
+
+    _DataType a_arr[n][n];
+    _DataType e_arr[n][n];
+
+    for (size_t i=0; i < n; i++)
```
shssf, line 192: did you do `python ./setup.py style -a`?

dpnp/linalg/dpnp_algo_linalg.pyx
```diff
@@ -167,50 +167,20 @@ cpdef dparray dpnp_eigvals(dparray input):
...
-                e_arr[ind_i, j] = e_arr[ind_i, j] - e_arr[ind_k, j] * temp
-
-    return e_arr
+cpdef dparray dpnp_inv(dparray input_):
+    cdef dparray input = input_.astype(dpnp.float64)
+    cdef DPNPFuncType param1_type = dpnp_dtype_to_DPNPFuncType(input.dtype)
+
+    cdef DPNPFuncData kernel_data = get_dpnp_function_ptr(DPNP_FN_INV, param1_type, param1_type)
+
+    cdef dparray result = dparray(input.size, dtype=dpnp.float64)
```
shssf, line 176: Are you sure you need fixed return type `dpnp.float64`?
Perhaps it is better to use `kernel_data.return_type` as above?
if you kernel must return `float64` it might be better to use `kernel_data.return_type` and put it into `{eft_DBL, (void*)dpnp_inv_c<double>};` for all input types.

dpnp/backend/kernels/dpnp_krnl_linalg.cpp
shssf: Also, please use prefix form for loop variable increment.

shssf:
Nice attempt, keep going!

dpnp/linalg/dpnp_algo_linalg.pyx
Rubtsowa: @shssf I'm sure that this need. Because with this type in result array algorithm for inv in cython start work.

PR #493, opened 2021-01-15
ENH: dpnp backend fallback to classic MKL; update `random.beta`

samir-nasibli:
# Description
* added stream for classic MKL call
* update `radnom.beta` backend func, when GPU selector

# TODO
* backend tests -> TODO pytest_cpp will be added -> other PR
* update example5.cpp -> other PR
* * rename `backend_sycl` -> other PR

### DIFF
--- a/dpnp/backend/kernels/dpnp_krnl_random.cpp
+++ b/dpnp/backend/kernels/dpnp_krnl_random.cpp
@@ -27,6 +27,7 @@
 #include "dpnp_fptr.hpp"
 #include "dpnp_utils.hpp"
 #include "queue_sycl.hpp"
+#include "mkl_vsl.h"
 
 #include <vector>
 
@@ -36,10 +37,7 @@ namespace mkl_blas = oneapi::mkl::blas;
 template <typename _DataType>
 void dpnp_rng_beta_c(void* result, _DataType a, _DataType b, size_t size)
 {
-    if (!size)
-    {
-        return;
-    }
+    int errcode;  // status TODO: using (could be redesigne)
 
     _DataType displacement = _DataType(0.0);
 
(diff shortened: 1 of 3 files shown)


dpnp/backend/src/queue_sycl.cpp
```diff
@@ -142,6 +143,16 @@ void backend_sycl::backend_sycl_rng_engine_init(size_t seed)
     rng_engine = new mkl_rng::mt19937(DPNP_QUEUE, seed);
 }
 
+void backend_sycl::backend_sycl_rng_stream_init(size_t seed)
```
shssf, line 146: Do we really need this stream?
Can we avoid keep it between function calls and create it every time in the function that used it?

dpnp/backend/src/queue_sycl.cpp
```diff
@@ -34,6 +34,7 @@
 cl::sycl::queue* backend_sycl::queue = nullptr;
 #endif
 mkl_rng::mt19937* backend_sycl::rng_engine = nullptr;
+VSLStreamStatePtr backend_sycl::stream = NULL;
```
shssf, line 37: It would be better to avoid global object for the stream (if it possible)

dpnp/backend/src/queue_sycl.cpp
samir-nasibli: Yes, we really need it. It is needed for saving prev. state and reseeding.
The only thing that is needed is rename `backend_sycl` to something like `backend_base`, that contains all these states and queue.

dpnp/backend/src/queue_sycl.cpp
samir-nasibli: We need it to save prev. state.

samir-nasibli:
@shssf this PR is ready for review/merge.
We just need rename `backend_sycl`.
Also we need to gtests for backend, that could be done on other PR.

shssf:
@samir-nasibli It looks like we need to rename `dpnp_srand_c()` to `dpnp_rng_srand_c`. I didn't do this in this PR to be clear.

shssf (approved):

samir-nasibli:
> @samir-nasibli It looks like we need to rename `dpnp_srand_c()` to `dpnp_rng_srand_c()`. I didn't do this in this PR to be clear.

Yes, we can do in on other PR.

samir-nasibli:
> Also, it is not quite clear why you removed `fmap[DPNPFuncName::DPNP_FN_RNG_BETA][eft_FLT][eft_FLT]`. Is it a typo?

No, this is not typo. We do not use `fmap[DPNPFuncName::DPNP_FN_RNG_BETA][eft_FLT][eft_FLT]` at all.

samir-nasibli:
> please note, I didn't check this on GPU.

We can wait with merge, it it is needed. We can do it after adding gtest tests for backend.

dpnp/backend/kernels/dpnp_krnl_random.cpp
```diff
@@ -23,16 +23,51 @@
...
+
+    return rng_stream;
+}
+
+void dpnp_srand_c(size_t seed)
+{
+    backend_sycl::backend_sycl_rng_engine_init(seed);
+    set_rng_stream(seed);
+}
+
```
samir-nasibli, line 70: I don't think that this is simpler.
I vote to save it as it was and revert you changes, except throw exception.

PR #495, opened 2021-01-17
oneapi on windows via conda

shssf:


shssf:
It doesn't provide any executable files. Libraries are installed into conda environment only.
so, useless.

PR #499, opened 2021-01-18
TESTS PERF add black scholes test

densmirn:
`DPNP_QUEUE_GPU=1 pytest tests/tests_perf/math_tests/test_black_scholes.py -s`

### DIFF
--- a/tests/tests_perf/test_perf_base.py
+++ b/tests/tests_perf/test_perf_base.py
@@ -94,6 +94,9 @@ class DPNPTestPerfBase:
         else:
             examine_function = custom_fptr
 
+        # warming up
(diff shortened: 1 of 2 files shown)


tests/tests_perf/test_perf_base.py
```diff
@@ -94,6 +94,9 @@ def dpnp_benchmark(self, name, lib, dtype, size, *args, custom_fptr=None, **kwar
         else:
             examine_function = custom_fptr
 
+        # warming up
+        result = examine_function(*args, **kwargs)
```
shssf, line 98: I would vote to avoid warming-up procedure. This makes us blind in performance assessment.
"madian" value has weak dependency on warmup numbers.

PR #502, opened 2021-01-19
Fix multiply

Rubtsowa:


Alexander-Makaryev:
if it is a fix then where are changes in tests?
what is fixed?

PR #507, opened 2021-01-20
Linalg kron func

Alexander-Makaryev:


dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -194,21 +194,30 @@ def inner(x1, x2, **kwargs):
     return call_origin(numpy.inner, x1, x2, **kwargs)
 
 
-def kron(x1, x2, **kwargs):
+def kron(a, b):
```
shssf, line 197: please don't use single letter variables names

dpnp/dpnp_algo/dpnp_algo.pyx
```diff
@@ -98,6 +98,10 @@ cpdef dparray dpnp_array(obj, dtype=None):
     cdef elem_dtype
     cdef dparray_shape_type obj_shape
 
+    # convert scalar to tuple
+    if dpnp.isscalar(obj):
```
shssf, line 102: maybe it is better to use https://github.com/IntelPython/dpnp/blob/dea4c7a2712bbb1c60b58e7b142cd318ef7a6641/dpnp/dpnp_utils/dpnp_algo_utils.pyx#L296?

shssf (approved):

dpnp/dpnp_iface_linearalgebra.py
Alexander-Makaryev: previously we decided to use the same names (as in original numpy) in python interface to be compliant with original numpy and its documentation, that is why I made these changes

dpnp/dpnp_algo/dpnp_algo.pyx
Alexander-Makaryev: I think that in this place it is better to explicitly show that we convert scalar to tuple tobe able to work with it as with sequence object.
BTW _object_to_tuple(object obj) will not work with all scalars, I made a separate PR for this https://github.com/IntelPython/dpnp/pull/512

PR #519, opened 2021-01-22
cmake parallel jobs

Alexander-Makaryev:


densmirn (approved):

shssf:
on windows:
```
AttributeError: module 'os' has no attribute 'sched_getaffinity'
```

densmirn:
> on windows:
> 
> ```
> AttributeError: module 'os' has no attribute 'sched_getaffinity'
> ```

There is another way to get cpu count that looks cross-platform:
```python
import multiprocessing; multiprocessing.cpu_count()
```

Alexander-Makaryev:
> > on windows:
> > ```
> > AttributeError: module 'os' has no attribute 'sched_getaffinity'
> > ```
> 
> There is another way to get cpu count that looks cross-platform:
> 
> ```python
> import multiprocessing; multiprocessing.cpu_count()
> ```

Yes, I started from this function and got the link to `os.sched_getaffinity()` from its documentation. I replaced one by another, in any way I hope that we will move this inside cmake. This temporal solution will help us to build faster for now.

shssf:
The question is "why we need it"
as far as I remember any parallel builds might brings instability into build process.
do we need one more possible instability?

shssf:
https://cmake.org/cmake/help/latest/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html
If this variable is defined empty the native build tool’s default number is used.

shssf:
Also, the issue with this approach is limitations.
- It works on linux only because "Ninja" generator has no such option.
- It works if run from python setuptools only
- it doesn't work if use `cmake` directly
- it doesn't work if use `make` directly after build directory structure generation

So, it is really better to implement such things inside CMakeList.txt file (in case we really need this in general)

Alexander-Makaryev:
> The question is "why we need it"
> as far as I remember any parallel builds might brings instability into build process.
> do we need one more possible instability?

With 6 cores it saves 2 minutes on compilation of backend library.
If we imagine that developer tries to build only 4 times per hour, then it will save more then 1 hour per day.

Alexander-Makaryev:
> Also, the issue with this approach is limitations.
> 
> * It works on linux only because "Ninja" generator has no such option.
> * It works if run from python setuptools only
> * it doesn't work if use `cmake` directly
> * it doesn't work if use `make` directly after build directory structure generation
> 
> So, it is really better to implement such things inside CMakeList.txt file (in case we really need this in general)

We all are working on Linux for now.
Almost all development tasks require full package, and developers build it through setup.py.
I spent some time on tries to implement it inside CMakeList.txt and left comment about it in the code.
As I wrote in previous comment, this temporary solution will allow us to build faster and save time.

shssf (approved):

PR #521, opened 2021-01-22
change __str__ output similar to Numpy

shssf:


dpnp/dparray.pyx
```diff
@@ -193,10 +193,7 @@ cdef class dparray:
 
         """
 
-        for i in range(self.size):
-            print(self[i], end=' ')
-
-        return "<__str__ TODO>"
+        return str(numpy.asarray(self))
```
samir-nasibli, line 196: 👍

PR #523, opened 2021-01-22
replace dparray with descriptor step1

shssf:


dpnp/dpnp_utils/dpnp_algo_utils.pyx
```diff
@@ -325,3 +326,68 @@ cpdef cpp_bool use_origin_backend(input1=None, size_t compute_size=0):
...
+            return None
+
+        data_tupple = self.descriptor["data"]
+        return data_tupple[0]
+
+    cdef void * get_data(self):
+        cdef long val = self.data
+        return < void * > val
+
+    def __bool__(self):
```
densmirn, line 387: Alternative option:
```python
    def __new__(cls, obj):
        if not hasattr(obj, "__array_interface__"):
            return None
        return super().__new__(cls)
```

dpnp/dpnp_utils/dpnp_algo_utils.pyx
```diff
@@ -325,3 +326,68 @@ cpdef cpp_bool use_origin_backend(input1=None, size_t compute_size=0):
...
+    @property
+    def is_scalar(self):
+        return self.dpnp_descriptor_is_scalar
+
+    @property
+    def data(self):
+        if self.descriptor is None:
+            return None
+
+        data_tupple = self.descriptor["data"]
```
densmirn, line 380:
```suggestion
        data_tuple = self.descriptor["data"]
```

dpnp/dpnp_utils/dpnp_algo_utils.pyx
samir-nasibli: @densmirn string cmp is very expensive

PR #542, opened 2021-01-28
add kernel for take

Rubtsowa:


dpnp/dpnp_algo/dpnp_algo_indexing.pyx
```diff
@@ -57,6 +57,9 @@ __all__ += [
 ]
 
 
+ctypedef void(*custom_linalg_2in_1out_func_ptr_t)(void *, void * , void * , size_t)
```
densmirn, line 60: Why did you use `linalg` in the name?

shssf (approved):

PR #556, opened 2021-02-01
svd: fix conditions in python layer

Alexander-Makaryev:


densmirn (commented):
Better to replace such statements as `if not a == True` with `if not a` or `if a is not True`, but it doesn't look a goal of the PR. So approved!

densmirn (approved):

shssf (approved):

shssf:
need all tests passed

PR #536, opened 2021-01-27
native kernel for funcs: ones, ones_like, zeros, zeros_like, full, full_like

Alexander-Makaryev:

PR #558, opened 2021-02-01
FULL add kernel

densmirn:


Alexander-Makaryev:
I think that there is a more common solution for functions like `full` in my PR https://github.com/IntelPython/dpnp/pull/536

densmirn:
> I think that there is a more common solution for functions like `full` in my PR #536

Almost, except that `full` copies a scalar from one-element input array to resulting array instead of element-wise copying.

dpnp/dpnp_algo/dpnp_algo_arraycreation.pyx
```diff
@@ -84,6 +85,28 @@ cpdef dparray dpnp_diag(v, k):
     return result
 
 
+cpdef dparray dpnp_full(result_shape, value_in, result_dtype):
+    """ Convert string type names (dparray.dtype) to C enum DPNPFuncType """
```
samir-nasibli, line 89:
```suggestion
    # Convert string type names (dparray.dtype) to C enum DPNPFuncType
```
Please use comment sign for comments otherwise Cython converts this to strings from src.

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -612,15 +612,16 @@ def full(shape, fill_value, dtype=None, order='C'):
 
     """
 
-    if (not use_origin_backend()):
-        if order not in ('C', 'c', None):
-            checker_throw_value_error("full", "order", order, 'C')
-
-        _dtype = dtype if dtype is not None else type(fill_value)
+    if not use_origin_backend():
+        if order != 'C':
```
samir-nasibli, line 616:
```suggestion
        if order is not 'C':
```

densmirn: Generally strings in Python are compared by value, not identity. I have one example which shows a potential issue:
```python
a = 'hi'
b = ''.join(['h', 'i'])
a == b  # True
a is b  # False
```

shssf (approved):

samir-nasibli (approved):
@densmirn in this PR you also moving some functional code (without changing) from one src to another, for future please do this kind of refactoring on separate PRs. PR must have one logic.

Alexander-Makaryev (approved):
👍

Issue #569, opened 2021-02-02
Gumbel distribution in random gives SegFault in the end of program execution (freeing resources)


Alexander-Makaryev:
```
$ pytest tests/test_random.py::TestDistributionsGumbel
============================================================================================================ test session starts ============================================================================================================
platform linux -- Python 3.7.7, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: /localdisk/work/amakarye/git/dpnp, inifile: setup.cfg
plugins: hypothesis-6.1.1
collecting ...
DPNP: current queue is CPU: 0
DPNP: version: 0.5.0

collected 4 items

tests/test_random.py ....                                                                                                                                                                                                             [100%]

============================================================================================================= 4 passed in 0.29s =============================================================================================================
Segmentation fault (core dumped)
```

samir-nasibli:
@Alexander-Makaryev there is no need in #570 . I have already PR #560 .

Alexander-Makaryev:
We should use math library but not our own implementations where it is possible. So I think that we want to save this issue and revert changes https://github.com/IntelPython/dpnp/pull/560 when fix in third party components will be provided.

PR #560, opened 2021-02-02
FIX: random.gumbel segfault on GPU

samir-nasibli:
# Description
* fix `random.gumbel` segfault on GPU
temporary solution for #569

PR #570, opened 2021-02-02
skip GPU tests for Gumbel distribution until it is not fixed

Alexander-Makaryev:
https://github.com/IntelPython/dpnp/issues/569


samir-nasibli:
There is no need in this PR. I have already PR #560 . Please, further do not merge in this way and add reviewers.

PR #578, opened 2021-02-07
add kernel for trapz function

KsanaKozlova:


dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -364,5 +388,10 @@ void func_map_init_mathematical(func_map_t& fmap)
     fmap[DPNPFuncName::DPNP_FN_REMAINDER][eft_DBL][eft_DBL] = {eft_DBL,
                                                                (void*)dpnp_remainder_c<double, double, double>};
 
+    fmap[DPNPFuncName::DPNP_FN_TRAPZ][eft_INT][eft_INT] = {eft_DBL, (void*)dpnp_cumprod_c<int, double>};
+    fmap[DPNPFuncName::DPNP_FN_TRAPZ][eft_LNG][eft_LNG] = {eft_DBL, (void*)dpnp_cumprod_c<long, double>};
+    fmap[DPNPFuncName::DPNP_FN_TRAPZ][eft_FLT][eft_FLT] = {eft_DBL, (void*)dpnp_cumprod_c<float, double>};
+    fmap[DPNPFuncName::DPNP_FN_TRAPZ][eft_DBL][eft_DBL] = {eft_DBL, (void*)dpnp_cumprod_c<double, double>};
+
```
samir-nasibli, line 395: this is not related.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -277,6 +277,30 @@ void dpnp_remainder_c(void* array1_in, void* array2_in, void* result1, size_t si
     event.wait();
 }
 
+template <typename _KernelNameSpecialization1, typename _KernelNameSpecialization2, typename _KernelNameSpecialization3>
+class dpnp_trapz_c_kernel;
+
+template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
```
samir-nasibli, line 283: Do we really need output datatype that differs from given two input arrays one?

samir-nasibli: outdated.

dpnp/dpnp_algo/dpnp_algo_mathematical.pyx
```diff
@@ -498,25 +499,25 @@ cpdef dparray dpnp_sum(dparray input, axis=None):
...
 
 
-cpdef dpnp_trapz(dparray y, dparray x, int dx):
+cpdef dpnp_trapz(dparray y1, dparray x1, double dx):
 
-    len = y.size
+    if y1.size <= 1:
+        if y1.dtype == dpnp.float32:
+            return dpnp.array([0], dtype=dpnp.float32)
+        return dpnp.array([0], dtype=dpnp.float64)
```
samir-nasibli, line 507: remove this if this case implemented on backend, or just move this logic on backend.

Leave only getting funcs pointers and mem alloc for dparrays on cython level.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -277,6 +277,43 @@ void dpnp_remainder_c(void* array1_in, void* array2_in, void* result1, size_t si
...
+    _DataType_output* result = reinterpret_cast<_DataType_output*>(result1);
+
+    if (array2_size < 2) {
+        _DataType_input1* sum = reinterpret_cast<_DataType_input1*>(dpnp_memory_alloc_c(1 * sizeof(_DataType_input1)));
+
+        dpnp_sum_c<_DataType_input1>(array1, sum, array1_size);
+
+        result[0] = sum[0];
+
+        dpnp_memory_free_c(sum);
```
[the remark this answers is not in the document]
densmirn: Why not?

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -281,21 +281,34 @@ template <typename _KernelNameSpecialization1, typename _KernelNameSpecializatio
...
     _DataType_input2* array2 = reinterpret_cast<_DataType_input2*>(array2_in);
     _DataType_output* result = reinterpret_cast<_DataType_output*>(result1);
 
-    for (size_t i = 0; i < nrow; ++i) {
-        size_t pos1 = i * row_len;
-        size_t pos2 = i * (row_len - 1);
-        double square = array2[pos2] * array1[pos1] + array2[pos2 + row_len - 2] * array1[pos1 + row_len - 1];
-        for (size_t j = 1; j < row_len - 1; ++j) {
-            square += array1[pos1 + j] * (array2[pos2 + j - 1] + array2[pos2 + j]);
+    if (array2_size < 2) {
```
densmirn, line 290: @KsanaKozlova what do you think about `array1_size != array2_size`? At the least you check the same in Python level.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -281,21 +281,34 @@ template <typename _KernelNameSpecialization1, typename _KernelNameSpecializatio
...
+
+        result[0] = sum[0];
+
+        dpnp_memory_free_c(sum);
+
+        result[0] -= (array1[0] + array1[array1_size - 1]) * 0.5;
+        result[0] *= dx;
+
+    } else {
+        result[0] += array1[0] * (array2[1] - array2[0]) +
```
densmirn, line 303: `+=` -> `=` ?

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -281,21 +281,34 @@ template <typename _KernelNameSpecialization1, typename _KernelNameSpecializatio
...
+                     array1[array1_size - 1] * (array2[array2_size - 1] - array2[array2_size - 2]);
+
+        for (size_t i = 1; i < array1_size - 1; ++i) {
```
densmirn, line 306: Wouldn't you like to implement Sycl kernel here?

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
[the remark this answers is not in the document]
samir-nasibli: because just this is enough, if 1 elem. Just math :)
```cpp
        result[0] = 0;
```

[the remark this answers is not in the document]
densmirn: `result` and `sum` could be of different types. So `result[0] = sum[0];` provides implicit casting.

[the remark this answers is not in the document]
samir-nasibli: > because just this is enough, if 1 elem. Just math :)

at least for cases when `array1_size` == `array2_size`, that implemented here.

> `result` and `sum` could be of different types. So `result[0] = sum[0];` provides implicit casting.

@densmirn no implicit casting for `result[0] = 0;`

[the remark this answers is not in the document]
densmirn: Why do you want to keep `result[0] = 0` if we need to compute sum of `array1` elements?

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -277,6 +277,43 @@ void dpnp_remainder_c(void* array1_in, void* array2_in, void* result1, size_t si
     event.wait();
 }
 
+template <typename _KernelNameSpecialization1, typename _KernelNameSpecialization2, typename _KernelNameSpecialization3>
+class dpnp_trapz_c_kernel;
+
+template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
+void dpnp_trapz_c(void* array1_in, void* array2_in, void* result1, double dx, size_t array1_size, size_t array2_size)
```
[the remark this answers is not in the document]
densmirn: Not always the same.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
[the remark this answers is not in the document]
densmirn: BTW `array1` is not one-element array.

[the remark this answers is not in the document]
samir-nasibli: Due to trapezoidal rule, you can see https://en.wikipedia.org/wiki/Trapezoidal_rule .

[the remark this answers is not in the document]
densmirn: Again `array1` is not one-element array.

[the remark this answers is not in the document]
samir-nasibli: > BTW array1 is not one-element array.

?

```cpp
if (array2_size < 2)
{
    result[0] = 0;
} // array1_size == array2_size 
```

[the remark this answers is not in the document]
densmirn: `array1_size == array2_size` is not always True.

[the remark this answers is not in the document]
samir-nasibli: @densmirn

1)  this backend func implemented for case `array1_size == array2_size`. The check provided on python level.
Also in this case, there is no need for two different size values (array1_size , array2_size) containing the same value.
2) This func is not implemented well if `array1_size` ,` array2_size` are different.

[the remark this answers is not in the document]
samir-nasibli: > Also in this case, there is no need for two different size values (array1_size , array2_size) containing the same value.

But I think we can save this params for API.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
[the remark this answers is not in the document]
samir-nasibli: For current implementation logic of this func they are the same val. But I think we can save this params for API.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -281,37 +281,62 @@ template <typename _KernelNameSpecialization1, typename _KernelNameSpecializatio
...
+    }
+
+    cl::sycl::event event;
+    _DataType_input1* array1 = reinterpret_cast<_DataType_input1*>(const_cast<void*>(array1_in));
+    _DataType_input2* array2 = reinterpret_cast<_DataType_input2*>(const_cast<void*>(array2_in));
     _DataType_output* result = reinterpret_cast<_DataType_output*>(result1);
 
-    if (array2_size < 2) {
-        _DataType_input1* sum = reinterpret_cast<_DataType_input1*>(dpnp_memory_alloc_c(1 * sizeof(_DataType_input1)));
+    if (array1_size < 2) {
```
densmirn, line 296:
```suggestion
    if (array1_size == 1) {
```

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -281,37 +281,62 @@ template <typename _KernelNameSpecialization1, typename _KernelNameSpecializatio
...
         result[0] *= dx;
 
     } else {
-        result[0] += array1[0] * (array2[1] - array2[0]) +
-                     array1[array1_size - 1] * (array2[array2_size - 1] - array2[array2_size - 2]);
 
-        for (size_t i = 1; i < array1_size - 1; ++i) {
-            result[0] += array1[i] * (array2[i + 1] - array2[i - 1]);
-        }
+        _DataType_output* cur_res = reinterpret_cast<_DataType_output*>(dpnp_memory_alloc_c((array1_size - 2) * sizeof(_DataType_output)));
```
densmirn, line 311: Let's move `array1_size - 2` to separate variable.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -281,37 +281,62 @@ template <typename _KernelNameSpecialization1, typename _KernelNameSpecializatio
...
-    _DataType_input1* array1 = reinterpret_cast<_DataType_input1*>(array1_in);
-    _DataType_input2* array2 = reinterpret_cast<_DataType_input2*>(array2_in);
+    if (!array1_size)
+    {
+        return;
+    }
+
+    cl::sycl::event event;
+    _DataType_input1* array1 = reinterpret_cast<_DataType_input1*>(const_cast<void*>(array1_in));
+    _DataType_input2* array2 = reinterpret_cast<_DataType_input2*>(const_cast<void*>(array2_in));
```
samir-nasibli, line 288:
```suggestion
    const _DataType_input1* array1 = reinterpret_cast<const _DataType_input1*>(array1_in);
    const _DataType_input2* array2 = reinterpret_cast<const _DataType_input2*>(array2_in);
```
try this, please.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -281,37 +281,62 @@ template <typename _KernelNameSpecialization1, typename _KernelNameSpecializatio
 class dpnp_trapz_c_kernel;
 
 template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
-void dpnp_trapz_c(void* array1_in, void* array2_in, void* result1, double dx, size_t array1_size, size_t array2_size)
+void dpnp_trapz_c(const void* array1_in, const void* array2_in, void* result1, double dx, size_t array1_size, size_t array2_size)
 {
```
samir-nasibli, line 285: please, add checks if given pointers are not `nullptr`. At least for the first of them, `array1_in`.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -281,37 +281,62 @@ template <typename _KernelNameSpecialization1, typename _KernelNameSpecializatio
...
-        _DataType_input1* sum = reinterpret_cast<_DataType_input1*>(dpnp_memory_alloc_c(1 * sizeof(_DataType_input1)));
+    if (array1_size < 2) {
+        result[0] = 0;
 
-        dpnp_sum_c<_DataType_input1>(array1, sum, array1_size);
+        return;
+    }
 
-        result[0] = sum[0];
+    if (array1_size != array2_size) {
```
samir-nasibli, line 297: You don't use `array2` here at all. How do you verify this algo? Are there tests for checking this?

tests/test_mathematical.py
```diff
@@ -135,39 +135,43 @@ def test_ediff1d_args(self):
...
     @pytest.mark.parametrize("y_array", [[1, 2, 4, 5],
-                                         [1., 2.5, 6., 7., ],
-                                         [2, 4, 6, 8]])
-    @pytest.mark.parametrize("x_array", [[1, 2, 3, 4],
-                                         [2, 4, 6, 8]])
-    def test_trapz_without_params(self, y_array, x_array):
-        y = numpy.array(y_array)
+                                         [1., 2.5, 6., 7., ]])
+    @pytest.mark.parametrize("x_array", [[2, 5, 6, 9]])
+    def test_trapz_with_x_params(self, y_array, x_array, data_type_y, data_type_x):
```
samir-nasibli, line 160: add checks for arrays ndim > 1, please.

tests/test_mathematical.py
```diff
@@ -135,39 +135,43 @@ def test_ediff1d_args(self):
...
 
-        x = numpy.array(x_array)
+        x = numpy.array(x_array, dtype=data_type_x)
         ix = inp.array(x)
 
         result = inp.trapz(iy, x=ix)
         expected = numpy.trapz(y, x=x)
         numpy.testing.assert_array_equal(expected, result)
 
     @pytest.mark.parametrize("y_array", [[1, 2, 4, 5],
```
samir-nasibli, line 171: check also with array ndim > 1, please.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
densmirn: We need to use either `array2` or `dx`. `array2` is used in else-block, `dx` in if-block.

densmirn (approved):

samir-nasibli:
```python
>>> a = [1, 2, 4, 5]
>>> dpnp_a = dpnp.array([1, 2, 4, 5]).reshape((2,2))
>>> dpnp_a
<DPNP DParray:name=dparray: mem=0x55b5bfd8f180: size=4: shape=(2, 2): type=int64>
>>> np_a = np.array([1, 2, 4, 5]).reshape((2,2))
>>> numpy.trapz(np_a)
array([1.5, 4.5])
>>> dpnp.trapz(dpnp_a)
<DPNP DParray:name=dparray: mem=0x55b5bfbfd580: size=2: shape=(2,): type=float64>
>>> print(dpnp.trapz(dpnp_a))
[1.5 4.5]
>>> print(dpnp.trapz(dpnp_a, dpnp_a))
  File "<stdin>", line 1, in <module>
  File "/localdisk/work/snasibli/dpnp_work/tmpr/dpnp/dpnp/dpnp_iface_mathematical.py", line 1488, in trapz
    return call_origin(numpy.trapz, y, x=x, dx=dx, **kwargs)
  ...
    return call_origin(numpy.trapz, y, x=x, dx=dx, **kwargs)
  File "dpnp/dpnp_utils/dpnp_algo_utils.pyx", line 84, in dpnp.dpnp_utils.dpnp_algo_utils.call_origin
    result_origin = function(*args_new, **kwargs)
  File "<__array_function__ internals>", line 6, in trapz
  File "dpnp/dparray.pyx", line 702, in dpnp.dparray.dparray.__array_function__
    return dpnp_func(*args, **kwargs)
RecursionError: maximum recursion depth exceeded while calling a Python object
>>> print(np.trapz(np_a, np_a))
[1.5 4.5]
```

It seems bug for arrays `ndmi` > 1. Please add tests for checking this too. You just can disable that tests and fix this bug on other PR.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
samir-nasibli: In this case it is better update condition and restructured the code
```cpp
if (array1_size == array2_size)
{
    // use array2_size
}
else
{
    // use dx
}
```
This is more readable.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -283,6 +283,11 @@ class dpnp_trapz_c_kernel;
 template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
 void dpnp_trapz_c(const void* array1_in, const void* array2_in, void* result1, double dx, size_t array1_size, size_t array2_size)
 {
+    if ((array1_in == nullptr) || (array2_in == nullptr && array1_size > 1))
```
densmirn, line 286: The condition is `True` if `array1 = np.array([1, 2, 3]); array2 = None`, but we don't need to return in this case. Isn't it?

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -283,7 +283,7 @@ class dpnp_trapz_c_kernel;
 template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
 void dpnp_trapz_c(const void* array1_in, const void* array2_in, void* result1, double dx, size_t array1_size, size_t array2_size)
 {
-    if ((array1_in == nullptr) || (array2_in == nullptr && array1_size > 1))
+    if ((array1_in == nullptr) || (array2_in == nullptr && array2_size > 1))
```
densmirn, line 286: 1. Is `(array2_in == nullptr && array2_size > 1` possible?
2. Can we somehow get `array2_in = nullptr` from cython?

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -283,7 +283,7 @@ class dpnp_trapz_c_kernel;
...
     {
         return;
     }
```
samir-nasibli: It is better separate this checks.

As for `array2_in`, it can be `None`. You have code for that case.

You may provide something like this:
```cpp
if (array1_in == nullptr || !array1_size)
{
    return;
}

if (array2_in == nullptr && (array2_size > 0))
{
    throw std::runtime_error("DPNP Error: dpnp_trapz_c() failed. Invalid arguments for array2_in");
    // throw std::runtime_error("DPNP Error: dpnp_trapz_c() failed.");  // or just this one
}
```

This would be good solution for backend lib users.
Also be aware `array2_in` from cython level can be non `nullptr`.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
samir-nasibli, line 286: It is better separate this checks.

As for `array2_in`, it can be `None`. You have code for that case.

You may provide something like this:
```cpp
if (array1_in == nullptr || !array1_size)
{
    return;
}

if (array2_in == nullptr && (array2_size > 0))
{
    throw std::runtime_error("DPNP Error: dpnp_trapz_c() failed. Invalid arguments for array2_in");
    // throw std::runtime_error("DPNP Error: dpnp_trapz_c() failed.");  // or just this one
}
```

This would be good solution for backend lib users.
Also be aware `array2_in` from cython level can be non `nullptr`.

tests/test_mathematical.py
```diff
@@ -135,39 +135,52 @@ def test_ediff1d_args(self):
...
-        x = numpy.array(x_array)
-        ix = inp.array(x)
+        x = numpy.array(x_array, dtype=data_type_x)
+        ix = inp.array(x_array, dtype=data_type_x)
 
         result = inp.trapz(iy, x=ix)
         expected = numpy.trapz(y, x=x)
         numpy.testing.assert_array_equal(expected, result)
 
+    @pytest.mark.parametrize("array", [[1, 2, 3], [4, 5, 6]])
```
samir-nasibli, line 171: no need for separate test suit, just add this case to prev test suits, please.

shssf (approved):

PR #589, opened 2021-02-11
COPYTO fix casting

densmirn:

PR #590, opened 2021-02-11
copyto alternative impl with macros

Alexander-Makaryev:
This is an alternative solution with macros using for PR
https://github.com/IntelPython/dpnp/pull/589
BTW Looks like we have a bug in macros MACRO_1ARG_2TYPES_OP, look into changes for details and a way to fix.


shssf:
I would not recommend to use the macro.
The macro has a bit diffrent meaning. This macro has INPUT and OUTPUT parameters but copy_to function requires 2 INPUTS and no outputs (by logic).
That is why it didn't allow diffrent output types with same input type - it is impossible in terms of original macro logic.

PR #602, opened 2021-02-15
Kernel put

Rubtsowa:


samir-nasibli:
@Rubtsowa there is no kernel added. Is the name of PR correct?

samir-nasibli:
@Rubtsowa please, add on PR description the passed tests names, that checks your changes.

dpnp/backend/kernels/dpnp_krnl_indexing.cpp
```diff
@@ -143,6 +143,27 @@ void dpnp_diagonal_c(
     return;
 }
 
+template <typename _DataType, typename _IndecesType, typename _ValueType>
```
samir-nasibli, line 146: unused template arguments?

dpnp/dpnp_algo/dpnp_algo_indexing.pyx
```diff
@@ -184,20 +185,38 @@ cpdef dpnp_place(dparray arr, dparray mask, vals):
             counter += 1
 
 
-cpdef dpnp_put(input, ind, v):
+cpdef dpnp_put(dparray input, ind, v):
     ind_is_list = isinstance(ind, list)
```
samir-nasibli, line 189: unused variable.

shssf (approved):

dpnp/dpnp_algo/dpnp_algo_indexing.pyx
```diff
@@ -184,20 +185,38 @@ cpdef dpnp_place(dparray arr, dparray mask, vals):
             counter += 1
 
 
-cpdef dpnp_put(input, ind, v):
+cpdef dpnp_put(dparray input, ind, v):
```
samir-nasibli, line 188:
``` suggestion
cpdef dpnp_put(dparray input_, ind, v):
```
In python/cython `input` is not a keyword. Instead, it is a built-in function. You can create a variable with the name `input`. But doing it is a bad practice because it overshadows the built-in (makes it unusable in the current scope). Use suggested convention, if you must use the name `input`.

samir-nasibli:
@shssf I think we can directly commit this kind of changes on master, if the review is not needed at all.
There is no any description provided for reviewers/other contributors at all.
* no info about new tests or passed old tests for the feature.

PR #603, opened 2021-02-15
Add axis, keepdims, out to sum() operation

shssf:

PR #610, opened 2021-02-18
FIX: typo in func initialization for reduction

samir-nasibli:


dpnp/backend/kernels/dpnp_krnl_reduction.cpp
```diff
@@ -143,7 +143,7 @@ void func_map_init_reduction(func_map_t& fmap)
     // WARNING. The meaning of the fmap is changed. Second argument represents RESULT_TYPE for this function
     // handle "out" and "type" parameters require user selection of return type
     // TODO. required refactoring of fmap to some kernelSelector
-    fmap[DPNPFuncName::DPNP_FN_SUM][eft_INT][eft_INT] = {eft_LNG, (void*)dpnp_sum_c<int, int>};
+    fmap[DPNPFuncName::DPNP_FN_SUM][eft_INT][eft_INT] = {eft_INT, (void*)dpnp_sum_c<int, int>};
```
densmirn, line 146: https://github.com/IntelPython/dpnp/pull/603#discussion_r576712941

samir-nasibli: This is little bit confused.

@densmirn This is bad practice doing smth for python level requirements.
We have only instances for ints here.

@shssf Currently I am working on global solution for `fmap`. Let's leave this PR as is for now.
I will update it or close in future dependently on the solution.

shssf:
tests should not pass because the original code is correct. Changed code is wrong.

samir-nasibli:
> tests should not pass because the original code is correct. Changed code is wrong.

@shssf the original code is correct for the tests but not clear. Provided solution just doesn't fix it. Also see, https://github.com/IntelPython/dpnp/pull/610#discussion_r579105184

PR #620, opened 2021-02-24
MULTIPLY enable broadcasting version 0.1

densmirn:


shssf (changes requested):
It is wrong way in design.
This multiply x3 number of kernels (approximately). 
Growing number of kernel is not big issue itself but from user perspective it is bad idea.
User do not expect diffrent functions in backend for "array+scalar" or "array+array" etc

One of the possible solution (is also a bad idea):
```void dpnp_multiply_array_scalar_c(void* input1_in, void* input2_in, void* result_out, const size_t size)```
replace with
```void dpnp_multiply_array_scalar_c(void* input1_in, void* input2_in, void* result_out, const size_t size_in1, size_t size_in2)```
and analize `size_in1` to understand count of values in input

**Comment by densmirn** [comment]
BTW user is able to use only one kernel `dpnp_multiply_c` to multiply either 2 arrays or array and scalar. Also we execute only `dpnp_maltiply` in Python level.

File: dpnp/backend/kernels/dpnp_krnl_mathematical.cpp [hunk 1]
```diff
@@ -232,6 +232,103 @@ void dpnp_modf_c(void* array1_in, void* result1_out, void* result2_out, size_t s
...
+            result[i] = input1 * input2[i];
+        }
+    };
+    auto kernel_func = [&](cl::sycl::handler& cgh) {
+        cgh.parallel_for<class dpnp_multiply_scalar_array_c_kernel<_DataType_input1, _DataType_input2,
+                                                                   _DataType_output>>(gws, kernel_parallel_for_func);
+    };
+    event = DPNP_QUEUE.submit(kernel_func);
+    event.wait();
+}
```
samir-nasibli, line 297:
```suggestion
template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
void dpnp_multiply_scalar_array_c(void* input1_in, void* input2_in, void* result_out, const size_t size)
{
    dpnp_multiply_array_scalar_c<_DataType_input2, _DataType_input1, _DataType_output>(input2_in, input1_in, result_out, size);
}
```
avoid code duplication, please.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -232,6 +232,103 @@ void dpnp_modf_c(void* array1_in, void* result1_out, void* result2_out, size_t s
...
+    }
+    else if (input1_size == 1)
+    {
+        dpnp_multiply_scalar_array_c<_DataType_input1, _DataType_input2, _DataType_output>(input1_in, input2_in,
+                                                                                           result_out, input2_size);
+    }
+    else if (input2_size == 1)
+    {
+        dpnp_multiply_array_scalar_c<_DataType_input1, _DataType_input2, _DataType_output>(input1_in, input2_in,
+                                                                                           result_out, input1_size);
```
samir-nasibli, line 328: Actually, you don't need different funcs for `array_vs_scalar` and `scalar_vs_array` cases, just reuse one of them.

dpnp/backend/include/dpnp_iface_fptr.hpp
```diff
@@ -127,6 +127,9 @@ enum class DPNPFuncName : size_t
     DPNP_FN_MINIMUM,                  /**< Used in numpy.minimum() implementation  */
     DPNP_FN_MODF,                     /**< Used in numpy.modf() implementation  */
     DPNP_FN_MULTIPLY,                 /**< Used in numpy.multiply() implementation  */
+    DPNP_FN_MULTIPLY_ARRAY_ARRAY,     /**< Used in numpy.multiply(array, array) implementation  */
+    DPNP_FN_MULTIPLY_ARRAY_SCALAR,    /**< Used in numpy.multiply(array, scalar) implementation  */
+    DPNP_FN_MULTIPLY_SCALAR_ARRAY,    /**< Used in numpy.multiply(scalar, array) implementation  */
```
samir-nasibli, line 132: You don't need all of them on public API. Only `DPNP_FN_MULTIPLY` is valuable.

samir-nasibli (changes requested):
Avoid code duplication.

It is better to redesign your functions. Leave `dpnp_multiply_c` only for public API, please. I also agree with https://github.com/IntelPython/dpnp/pull/620#pullrequestreview-597753899 proposal.

You can designe smth like this:
```cpp
// array vs array (for public API)
template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
INP_DLLEXPORT void dpnp_multiply_c(
                                   void* input1_in,
                                   void* input2_in,
                                   void* result_out,
                                   const size_t input1_size,
                                   const size_t input2_size);

// scalar vs array (call it from prev. ones, not for public API)
template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
INP_DLLEXPORT void dpnp_multiply_ls_scal_c(
                                   _DataType_input1 input1_in,
                                   void* input2_in,
                                   void* result_out,
                                   const size_t input1_size,
                                   const size_t input2_size);

// array vs scalar, just reuse (scalar vs array) and call templated prev. one (also is not for public API)
template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
INP_DLLEXPORT void dpnp_multiply_rs_scal_c(
                                   void* input1_in,
                                   _DataType_input2 input2_in,
                                   void* result_out,
                                   const size_t input1_size,
                                   const size_t input2_size);
```

And I think you also just can override your multiply function and try to do something like this:
```cpp
// array vs array (for public API)
template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
INP_DLLEXPORT void dpnp_multiply_c(
                                   void* input1_in,
                                   void* input2_in,
                                   void* result_out,
                                   const size_t input1_size,
                                   const size_t input2_size);

// scalar vs array (call it from prev. ones, not for public API)
template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
INP_DLLEXPORT void dpnp_multiply_c(
                                   _DataType_input1 input1_in,
                                   void* input2_in,
                                   void* result_out,
                                   const size_t input1_size,
                                   const size_t input2_size);

// array vs scalar, just reuse (scalar vs array) and call templated prev. one (also is not for public API)
template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
INP_DLLEXPORT void dpnp_multiply_c(
                                   void* input1_in,
                                   _DataType_input2 input2_in,
                                   void* result_out,
                                   const size_t input1_size,
                                   const size_t input2_size);
```

samir-nasibli:
@densmirn Also it is better split adding this feature and broadcasting feature into different PRs.

shssf:
> BTW user is able to use only one kernel `dpnp_multiply_c` to multiply either 2 arrays or array and scalar. .
is not by the code. I see you added 3 functions into dpnp/backend/include/dpnp_iface.hpp

dpnp/backend/include/dpnp_iface_fptr.hpp
densmirn: Currently I have different types mapping for all the cases `array * array`, `scalar * array`, `array * scalar` to be consistent with `numpy`. So let me know how simultaneously to save the consistency and drop the names.

densmirn:
> @densmirn Also it is better split adding this feature and broadcasting feature into different PRs.

Looks like I don't have broadcasting in this PR except commit message.

dpnp/backend/include/dpnp_iface_fptr.hpp
samir-nasibli: Firstly, there are no `scalar * array` and `array * scalar` cases, you have only one for this (`scaclar * array`).
So you don't have different algos for `scalar * array` and `array * scalar` cases.

Secondly, you have the same "different types mapping" for all `lhs * rhs`. 

I don't understand what exactly you cannot do. Try to describe the problem in more detail.
There is no rocket scinece here, see also prev. comments.

samir-nasibli:
> > @densmirn Also it is better split adding this feature and broadcasting feature into different PRs.
> 
> Looks like I don't have broadcasting in this PR except commit message.

There is no "looks like".
Actually you have no functionality here, which mathes with `broadcating`.
And which is actually correct, just rename currrent PR (remove `broadcating`). Add you brodcasting features in separate PR (s).

dpnp/backend/include/dpnp_iface_fptr.hpp
densmirn: I don't have the same "different types mapping" for all `lhs * rhs`.

densmirn: Moreover `scalar * array` and `array * scalar` cases have different types mapping.

samir-nasibli: > I don't have the same types mapping for all `lhs * rhs`.

Try to give more details for your issue.
This is really no big problem, just extend the general one. (or there a lot of solutions for that)

dpnp/backend/include/dpnp_iface_fptr.hpp
```diff
@@ -127,6 +127,8 @@ enum class DPNPFuncName : size_t
     DPNP_FN_MINIMUM,                  /**< Used in numpy.minimum() implementation  */
     DPNP_FN_MODF,                     /**< Used in numpy.modf() implementation  */
     DPNP_FN_MULTIPLY,                 /**< Used in numpy.multiply() implementation  */
+    DPNP_FN_MULTIPLY_ARRAY_ARRAY,     /**< Used in numpy.multiply(array, array) implementation  */
+    DPNP_FN_MULTIPLY_ARRAY_SCALAR,    /**< Used in numpy.multiply(array, scalar) implementation  */
```
shssf, line 131: The proposal here might be - leave this code is it was before. No addition needed. Everything will be the same from "user" perspective.
One thing should be added - `size_in2`. The parameter to get information about size of each array (size will be size_in1 and one more extra parameter `size_in2`)
Down the call stack, it might be implemented as dispatcher - each combination goes to it's special internal backend function with specific kernel.
From user perspective - minor change and no extra public functions.
Internally - more kernels for each combination (array+scalar, array+array, etc)
Internal implementation might be changed in future without public interface changes.

shssf: This is one of possible "acceptable" implementation. This is not quite beauty but it will work. Also, many other libraries implemented in this way (one public function and many internal specializations).
The main limitation factor I see here - we can't specialize it in compile time. Templates usage might be better solution but we have no information in compile time about parameters (array or scalar).

shssf: So, I see you are going in right direction. Some cleanup of this PR needed

dpnp/backend/include/dpnp_gen_2arg_3type_tbl.hpp
```diff
@@ -66,7 +66,7 @@ MACRO_2ARG_3TYPES_OP(dpnp_fmod_c, cl::sycl::fmod((double)input_elem1, (double)in
 MACRO_2ARG_3TYPES_OP(dpnp_hypot_c, cl::sycl::hypot((double)input_elem1, (double)input_elem2), oneapi::mkl::vm::hypot)
 MACRO_2ARG_3TYPES_OP(dpnp_maximum_c, cl::sycl::max(input_elem1, input_elem2), oneapi::mkl::vm::fmax)
 MACRO_2ARG_3TYPES_OP(dpnp_minimum_c, cl::sycl::min(input_elem1, input_elem2), oneapi::mkl::vm::fmin)
-MACRO_2ARG_3TYPES_OP(dpnp_multiply_c, input_elem1* input_elem2, oneapi::mkl::vm::mul)
+MACRO_2ARG_3TYPES_OP(dpnp_multiply_array_array_c, input_elem1 * input_elem2, oneapi::mkl::vm::mul)
```
shssf, line 69: we lost `dpnp_multiply_c` function here.

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -232,6 +232,71 @@ void dpnp_modf_c(void* array1_in, void* result1_out, void* result2_out, size_t s
...
+    else if (input1_size == 1)
+    {
+        dpnp_multiply_array_scalar_c<_DataType_input2, _DataType_input1, _DataType_output>(input2_in, input1_in,
+                                                                                           result_out, input2_size);
+    }
+    else if (input2_size == 1)
+    {
+        dpnp_multiply_array_scalar_c<_DataType_input1, _DataType_input2, _DataType_output>(input1_in, input2_in,
+                                                                                           result_out, input1_size);
+    }
```
shssf, line 297: This approach I mentioned above in comments

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -232,6 +232,71 @@ void dpnp_modf_c(void* array1_in, void* result1_out, void* result2_out, size_t s
...
+    auto kernel_func = [&](cl::sycl::handler& cgh) {
+        cgh.parallel_for<class dpnp_multiply_array_scalar_c_kernel<_DataType_input1, _DataType_input2,
+                                                                   _DataType_output>>(gws, kernel_parallel_for_func);
+    };
+    event = DPNP_QUEUE.submit(kernel_func);
+    event.wait();
+}
+
+template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
+void dpnp_multiply_c(void* input1_in,
```
shssf, line 268: Also, It might be better to put all required parameters here. Most of them will not be used at this time. These parameters will be passed to all other specialization functions.

dpnp/dpnp_algo/dpnp_algo.pxd
```diff
@@ -200,13 +204,29 @@ cdef extern from "dpnp_iface.hpp":
...
 ctypedef void(*fptr_2in_1out_t)(void * , void*, void*, size_t)
+ctypedef void(*fptr_2in_1out_2sz_t)(const void *, const void *, const void *, const size_t, const size_t)
 ctypedef void(*fptr_blas_gemm_2in_1out_t)(void * , void * , void * , size_t, size_t, size_t)
 ctypedef void(*dpnp_reduction_c_t)(const void *, const size_t, void * , const long*, const size_t, const long*, const size_t, const void * , const long*)
 
 cdef dparray call_fptr_1in_1out(DPNPFuncName fptr_name, dparray x1, dparray_shape_type result_shape)
 cdef dparray call_fptr_2in_1out(DPNPFuncName fptr_name, dparray x1, dparray x2, dparray_shape_type result_shape)
 
 
+ctypedef fused dpnp_numeric:
```
shssf, line 215: Fused types is a bit scaring for me. I played with them sometime ago and reject this solution (select "void*" instead).
Did you try `complex<double>` and bool in this "fused type" approach? Does it work?

dpnp/dpnp_algo/dpnp_algo.pyx
```diff
@@ -255,17 +255,17 @@ Internal functions
 """
 cpdef DPNPFuncType dpnp_dtype_to_DPNPFuncType(dtype):
 
-    if dtype in [numpy.float64, 'float64']:
+    if dtype in [numpy.float64, 'float64', float]:
```
shssf, line 258: type `float` is 32 bit.

dpnp/dpnp_algo/dpnp_algo_mathematical.pyx
```diff
@@ -273,23 +273,37 @@ cpdef tuple dpnp_modf(dparray x1):
...
-        x2_ = dpnp.array([x2])
+cpdef dparray dpnp_multiply(dpnp_input1 x1, dpnp_input2 x2):
+    if dpnp_input1 in dpnp_numeric and dpnp_input2 is dparray:
+        return dpnp_multiply_array_scalar(x2, x1)
+    elif dpnp_input1 is dparray and dpnp_input2 in dpnp_numeric:
+        return dpnp_multiply_array_scalar(x1, x2)
+    elif dpnp_input1 is dparray and dpnp_input2 is dparray:
+        return call_fptr_2in_1out(DPNP_FN_MULTIPLY_ARRAY_ARRAY, x1, x2, x1.shape)
+    else:
+        return x1 * x2
```
shssf, line 284: It looks like this code must be inside DPNPC (not at Cython leyer)

dpnp/dpnp_utils/dpnp_algo_utils.pyx
```diff
@@ -240,6 +240,30 @@ cdef tuple get_shape_dtype(object input_obj):
     return (return_shape, numpy.dtype(type(input_obj)))
 
 
+cdef dparray_shape_type get_common_shape(dparray_shape_type input1_shape, dparray_shape_type input2_shape):
```
shssf, line 243: didn't find where it is used.
Do you expect this algorithm will be the same for all possible variant and functions?
Term "all" - I mean 98% (exceptions might be handled later) of possible variants.

it looks like this is a part of common "broadcast" approach.

dpnp/dpnp_algo/dpnp_algo.pyx
densmirn: Python's `float` is 64 bit
```python
a = 2.0
type(a)  # <class 'float'>
np.array(a).dtype  # dtype('float64')
```

dpnp/backend/include/dpnp_gen_2arg_3type_tbl.hpp
shssf: also, `rng` might be a good example

dpnp/dpnp_algo/dpnp_algo.pyx
shssf: I'm afraid we will have a `trap` here. It might be interpreter specific. Do you think we really need "python type" mentioned in this code? It would be great to avoid it at this code.

shssf:
@densmirn Sorry for so much comments. This is why people prefer "design proposal" before any implementations.

dpnp/backend/include/dpnp_gen_2arg_3type_tbl.hpp
samir-nasibli: > we lost `dpnp_multiply_c` function here. I mean - it looks like we don't need this line of code here. Just delete it. Add dpnp_multiply_c as standalone function declaration. Also, it might be good to create separate `mathematical.hpp` file to collect all function related to this group there. The same as it is done at cython and python layer

Good idea. And it is better to do such refactoring work on separate PR.

dpnp/dpnp_algo/dpnp_algo.pxd
densmirn: Yes both `complex<double>` and `bool` work fine. Currently I don't see alternative way to use a cython function like a c++ function template.

dpnp/backend/include/dpnp_iface_fptr.hpp
densmirn: We can't remove `DPNP_FN_MULTIPLY_ARRAY_SCALAR` because we have different return type for input `array*array` and `array*scalar` of the same type.

dpnp/dpnp_algo/dpnp_algo_mathematical.pyx
densmirn: We have different return type for input `array*array` and `array*scalar` of the same type.

dpnp/dpnp_utils/dpnp_algo_utils.pyx
densmirn: I expect so.

dpnp/backend/include/dpnp_iface.hpp
```diff
@@ -658,6 +658,31 @@ INP_DLLEXPORT void dpnp_floor_divide_c(void* array1_in, void* array2_in, void* r
...
+ * @param [in]  input2_size        Number of elements in input 2.
+ * @param [in]  input1_shape       Shape of input 1.
+ * @param [in]  input2_shape       Shape of input 2.
+ * @param [in]  input1_shape_ndim  Size of shape 1.
+ * @param [in]  input2_shape_ndim  Size of shape 2.
+ */
+template <typename _DataType_input1, typename _DataType_input2, typename _DataType_output>
+INP_DLLEXPORT void dpnp_multiply_c(const void* input1_in,
+                                   const void* input2_in,
+                                   const void* result_out,
```
samir-nasibli, line 678: do we really need const pointer for output array?

dpnp/backend/include/dpnp_iface_fptr.hpp
samir-nasibli: Let the dispatcher function handle this, and remove this from public API.

dpnp/dpnp_algo/dpnp_algo.pxd
```diff
@@ -205,12 +206,32 @@ cdef extern from "dpnp_iface.hpp":
...
+    bint
+    double complex
+
+ctypedef fused dpnp_input1:
+    dparray
+    dpnp_scalar
+
+ctypedef fused dpnp_input2:
+    dparray
+    dpnp_scalar
```
samir-nasibli, line 234: avoid using fused types.
Looks like a workaround, please, redesign it.

tests/skipped_tests.tbl
```diff
@@ -1141,6 +1141,8 @@ tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsDiri
 tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsExponential_param_0_{scale_shape=(), shape=(4, 3, 2)}::test_exponential
 tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsExponential_param_1_{scale_shape=(), shape=(3, 2)}::test_exponential
 tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsExponential_param_2_{scale_shape=(), shape=None}::test_exponential
+tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsGeometric_param_0_{p_shape=(), shape=(4, 3, 2)}::test_geometric
+tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsGeometric_param_1_{p_shape=(), shape=(3, 2)}::test_geometric
```
samir-nasibli, line 1145: Is this needed for checking you function logic?

tests/skipped_tests_gpu.tbl
```diff
@@ -1504,6 +1504,8 @@ tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsDiri
 tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsExponential_param_0_{scale_shape=(), shape=(4, 3, 2)}::test_exponential
 tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsExponential_param_1_{scale_shape=(), shape=(3, 2)}::test_exponential
 tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsExponential_param_2_{scale_shape=(), shape=None}::test_exponential
+tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsGeometric_param_0_{p_shape=(), shape=(4, 3, 2)}::test_geometric
+tests/third_party/cupy/random_tests/test_distributions.py::TestDistributionsGeometric_param_1_{p_shape=(), shape=(3, 2)}::test_geometric
```
samir-nasibli, line 1508: Is this needed for checking you function logic?

dpnp/dpnp_utils/dpnp_algo_utils.pyx
```diff
@@ -240,6 +240,30 @@ cdef tuple get_shape_dtype(object input_obj):
...
+        if input1_shape[it] == input2_shape[it]:
+            result_shape.push_back(input1_shape[it])
+        elif input1_shape[it] == 1:
+            result_shape.push_back(input2_shape[it])
+        elif input2_shape[it] == 1:
+            result_shape.push_back(input1_shape[it])
+        else:
+            err_msg = f"{ERROR_PREFIX} in function get_common_shape()"
+            err_msg += f"operands could not be broadcast together with shapes {input1_shape} {input2_shape}"
+            ValueError(err_msg)
```
samir-nasibli, line 262: leave this logic on python level, please.

densmirn: Do you mean move whole function to Python?

tests/skipped_tests.tbl
densmirn: Function `multiply` is used in the tests and works correct there. To prove that I added separate test with similar check.

dpnp/dpnp_algo/dpnp_algo.pxd
densmirn: If you have redesign ideas let me know.

samir-nasibli: fused types is just a feature for templates. You just don't really need it on cython layer.
And more over you are using this types on `dpnp_multiply` in bad way, where lost all cython templating idea.

tests/skipped_tests.tbl
samir-nasibli: if multiply is used in current tests, then it is ok.

dpnp/dpnp_utils/dpnp_algo_utils.pyx
samir-nasibli: No error handling, error raising and etc. cython level.

samir-nasibli: You are trying to do all in one PR - it is really bad idea.
move not realted and more general (common) changes on seperate PR

dpnp/backend/kernels/dpnp_krnl_mathematical.cpp
```diff
@@ -411,6 +535,87 @@ void func_map_init_mathematical(func_map_t& fmap)
...
+        eft_LNG, (void*)dpnp_multiply_c<long, bool, long>};
+    fmap[DPNPFuncName::DPNP_FN_MULTIPLY_ARRAY_SCALAR][eft_BLN][eft_FLT] = {
+        eft_DBL, (void*)dpnp_multiply_c<double, bool, float>};
+    fmap[DPNPFuncName::DPNP_FN_MULTIPLY_ARRAY_SCALAR][eft_BLN][eft_DBL] = {
+        eft_DBL, (void*)dpnp_multiply_c<double, bool, double>};
+    fmap[DPNPFuncName::DPNP_FN_MULTIPLY_ARRAY_SCALAR][eft_INT][eft_BLN] = {
+        eft_INT, (void*)dpnp_multiply_c<int, int, bool>};
+    fmap[DPNPFuncName::DPNP_FN_MULTIPLY_ARRAY_SCALAR][eft_INT][eft_INT] = {
+        eft_INT, (void*)dpnp_multiply_c<int, int, int>};
+    fmap[DPNPFuncName::DPNP_FN_MULTIPLY_ARRAY_SCALAR][eft_INT][eft_LNG] = {
```
shssf, line 580: Is it really eft_INT * eft_LNG = eft_INT ?
It is not a request for changes. It is just a note. I think we need to find better solution (I hope we can).

densmirn: Yes, it is.
```python
import numpy
a = numpy.array([1, 2, 3], dtype=numpy.int32)
b = numpy.int64(5).item()  # like b = 5
c = numpy.multiply(a, b)
print(a.dtype, type(b), c.dtype)  # int32 <class 'int'> int32
```
Python `int` means `numpy.int64` (`long`).

samir-nasibli:
@densmirn @shssf 

I agree with the [comment](https://github.com/IntelPython/dpnp/pull/620#discussion_r588007504) and it seems to me that the current solution (adding a separate interface for array * scalar) is not good.

I found the following differences here:

|input1 type|input2 type|return for array vs array|return for array vs scalar|
|:---|:---|:---|:---|
|bool|int|int|long|
|bool|float|float|double|
|int|long|long|int|
|float|int|double|float|
|float|long|double|float|


I think that the solution to the current problem will come with the enabling `out` param.

And for these, we will need to implement dtype converter (with different converting kinds). (not in this PR)

In any case, we will need to support something like this in the future for multiply funcs:
```python
In [15]: np.multiply(a,b, out=np.zeros(shape=a.shape, dtype=np.float16))
Out[15]:
array([  0.,  45.,  90., 135., 180., 225., 270., 315., 360.],
      dtype=float16)

In [16]: np.multiply(a,b, out=np.zeros(shape=a.shape, dtype=np.float16)).dtype
Out[16]: dtype('float16')

In [17]: b
Out[17]: 45

In [18]: a
Out[18]: array([0., 1., 2., 3., 4., 5., 6., 7., 8.], dtype=float32)

```

So we will need to have something like this for  `dpnp_multiply_c<output_type><type1><type2>` function flow:
1) type1 \* type2 and save results in `intermediate_result` array with type of type1 or type2, if `output_type` differs from `type1` or `type2`, or just save result in `output` array, if not. We need buffer mem for `intermediate_results`.
2) if `output_type` differs from `type1` or `type2`, due to provided by `out` param or according different multilication type results for array vs scalar or array vs array cases), convert `intermediate_result` output to needed output type and save result in `output`.

This flow can be implemented in future.

Currently in this PR, I propose to have the same output types for all rhs and lhs.

We can for now have in most cases the output as one of the input types, and support these table for all rhs and lhs (where we have diffs for out types for `arr vs scal` and `arr vs arr` cases)

|input1 type|input2 type|return for all rhs\*lhs|
|:---|:---|:---|
|bool|int|int|
|bool|float|float|
|int|long|long|
|float|int|float|
|float|long|float|

And save only `DPNP_FN_MULTIPLY` for pulic API.

Perhaps some tests will have to be disabled for this.

shssf:
intermediate proposal in #641

tests/skipped_tests.tbl
samir-nasibli: > if multiply is used in current tests, then it is ok.

non correct comment here. My fault.

tests/skipped_tests_gpu.tbl
samir-nasibli: What is the reason? Provide some info, please

tests/skipped_tests.tbl
densmirn: I thinks so.

tests/skipped_tests_gpu.tbl
densmirn: Likely due to `types_map` in the old code that multiplies array by scalar https://github.com/IntelPython/dpnp/blob/85aad6be56c4555855da62f0b21c978902448d24/dpnp/dpnp_algo/dpnp_algo_mathematical.pyx#L279-L290

PR #627, opened 2021-02-26
Native triu

Rubtsowa:


dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
```diff
@@ -151,6 +151,85 @@ void dpnp_tril_c(void* array_in, void* result1, const int k, size_t* shape, size
     return;
 }
 
+template <typename _DataType>
+void dpnp_triu_c(void* array_in, void* result1, const int k, size_t* shape, size_t* res_shape, const size_t ndim, const size_t res_ndim)
+{
```
samir-nasibli, line 156: there are no checks for `nullptr`.

dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
```diff
@@ -151,6 +151,85 @@ void dpnp_tril_c(void* array_in, void* result1, const int k, size_t* shape, size
...
+            if (ids[res_ndim - 1] >= diag_idx)
+            {
+                result[i] = array_m[i];
+            }
+            else
+            {
+                result[i] = 0;
+            }
+        }
+    }
```
samir-nasibli, line 229: code duplication - bad practice. It is better to redesign.

tests/test_arraycreation.py
```diff
@@ -136,3 +136,29 @@ def test_tril(m, k):
     expected = numpy.tril(a, k)
     result = dpnp.tril(ia, k)
     numpy.testing.assert_array_equal(expected, result)
+
+
+@pytest.mark.parametrize("k",
+                         [-6, -5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5, 6],
+                         ids=['-6', '-5', '-4', '-3', '-2', '-1', '0', '1', '2', '3', '4', '5', '6'])
```
samir-nasibli, line 143: Do you really need to test all these cases?
Isn't one instance enough for cases k>0, k=0, k<0 for testing?

tests/test_arraycreation.py
```diff
@@ -136,3 +136,29 @@ def test_tril(m, k):
...
+                          [[1, 2], [3, 4]],
+                          [[0, 1, 2], [3, 4, 5], [6, 7, 8]],
+                          [[0, 1, 2, 3, 4], [5, 6, 7, 8, 9]]],
+                         ids=['[0, 1, 2, 3, 4]',
+                              '[1, 1, 1, 1, 1]',
+                              '[[0, 0], [0, 0]]',
+                              '[[1, 2], [1, 2]]',
+                              '[[1, 2], [3, 4]]',
+                              '[[0, 1, 2], [3, 4, 5], [6, 7, 8]]',
+                              '[[0, 1, 2, 3, 4], [5, 6, 7, 8, 9]]'])
```
samir-nasibli, line 158: the same, see as prev. one for test param, please.

dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
```diff
@@ -151,6 +151,85 @@ void dpnp_tril_c(void* array_in, void* result1, const int k, size_t* shape, size
...
+template <typename _DataType>
+void dpnp_triu_c(void* array_in, void* result1, const int k, size_t* shape, size_t* res_shape, const size_t ndim, const size_t res_ndim)
+{
+    _DataType* array_m = reinterpret_cast<_DataType*>(array_in);
+    _DataType* result = reinterpret_cast<_DataType*>(result1);
+
+    size_t res_size = 1;
+    for (size_t i = 0; i < res_ndim; ++i)
+    {
+        res_size *= res_shape[i];
```
samir-nasibli, line 163: It doesn't work correct when shape element is zero. Do checks before multiply.
```python
>>> a = numpy.ones(shape=(1, 2, 0))
>>> a.shape
(1, 2, 0)
```

And add test case for testing with this kind given shape.

dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
```diff
@@ -151,6 +151,85 @@ void dpnp_tril_c(void* array_in, void* result1, const int k, size_t* shape, size
...
+    if (ndim == 1)
+    {
+        for (size_t i = 0; i < res_size; ++i)
+        {
+            size_t n = res_size;
+            size_t val = i;
+            int ids[res_ndim];
+            for (size_t j = 0; j < res_ndim; ++j)
+            {
+                n /= res_shape[j];
```
samir-nasibli, line 175: What if `res_shape[j]` is zero?

dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
```diff
@@ -151,6 +151,85 @@ void dpnp_tril_c(void* array_in, void* result1, const int k, size_t* shape, size
...
+    else
+    {
+        for (size_t i = 0; i < res_size; ++i)
+        {
+            size_t n = res_size;
+            size_t val = i;
+            int ids[res_ndim];
+            for (size_t j = 0; j < res_ndim; ++j)
+            {
+                n /= res_shape[j];
```
samir-nasibli, line 207: the same.What if `res_shape[j]` is zero? (see prev. comment also)

dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
```diff
@@ -151,6 +151,85 @@ void dpnp_tril_c(void* array_in, void* result1, const int k, size_t* shape, size
     return;
 }
 
+template <typename _DataType>
+void dpnp_triu_c(void* array_in, void* result1, const int k, size_t* shape, size_t* res_shape, const size_t ndim, const size_t res_ndim)
+{
+    _DataType* array_m = reinterpret_cast<_DataType*>(array_in);
+    _DataType* result = reinterpret_cast<_DataType*>(result1);
+
```
samir-nasibli, line 159: there are no checks if `ndim` or `res_ndim` is zore.

samir-nasibli (changes requested):
* Avoid code duplication.
* checks for given input values is important.
* Provide numpy original test names, that test this functionality(at least on PR description). (this is significant numpy tests pass.)
* More test coverage, and redesign current test cases (see comments).

Please, provide changes based on suggestions.

dpnp/dpnp_algo/dpnp_algo_arraycreation.pyx
```diff
@@ -262,34 +262,20 @@ cpdef dparray dpnp_tril(dparray m, int k):
...
-        result = dparray(shape=m.shape, dtype=m.dtype)
+    cdef DPNPFuncType param1_type = dpnp_dtype_to_DPNPFuncType(m.dtype)
 
-        for i in range(result.size):
-            ids = get_axis_indeces(i, result.shape)
+    cdef DPNPFuncData kernel_data = get_dpnp_function_ptr(DPNP_FN_TRIU, param1_type, param1_type)
 
-            diag_idx = max(-1, ids[result.ndim - 2] + k)
-            diag_idx = min(diag_idx, result.shape[result.ndim - 1])
+    cdef custom_1in_1out_func_ptr_t func = <custom_1in_1out_func_ptr_t > kernel_data.ptr
```
samir-nasibli, line 277: this pointer type (`custom_1in_1out_func_ptr_t `) is used in several places on this src.
Bad name, should be renamed (not in this PR)

dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
```diff
@@ -110,15 +110,35 @@ void dpnp_tril_c(void* array_in,
...
     size_t res_size = 1;
     for (size_t i = 0; i < res_ndim; ++i)
     {
         res_size *= res_shape[i];
     }
 
+    if (res_size == 0)
+    {
+        return;
+    }
```
samir-nasibli, line 140: this looks like a some solution for  https://github.com/IntelPython/dpnp/pull/627/files#r583932215

But is it correct in terms of algorithm? Is this the behavior of `numpy` for such a given shape?
Do you have any tests for this case?

dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
samir-nasibli: outdated.

samir-nasibli (commented):
Thanks, for the update.

Please, provide numpy origianal tests that check your funcs.

TODOs:
* code duplication can be fixed both in the current RP as well as in a separate.
* add checks for case stated on [comment](https://github.com/IntelPython/dpnp/pull/627/files#r583932215). If it will not work, you can fix  it in the current RP as well as in a separate.

dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
samir-nasibli: outdated.

dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
Rubtsowa: @samir-nasibli  numpy return empty array in this case. For example:

>>> a=np.ndarray((5, 0, 3), dtype=np.int64)
>>> p=np.triu(a)
>>> p
array([], shape=(5, 0, 3), dtype=int64)

samir-nasibli: Nice. Add please tests, that checks the same behaviour for noted case.

shssf (approved):

PR #628, opened 2021-02-28
Reduce iterator. Axes as array support

shssf:


dpnp/backend/src/dpnp_iterator.hpp
```diff
@@ -103,49 +99,88 @@ class DPNP_USM_iterator final
...
     inline bool operator!=(const DPNP_USM_iterator& __rhs) const
     {
-        return (data != __rhs.data);
+        return !(*this == __rhs);
     };
 
     inline bool operator<(const DPNP_USM_iterator& __rhs) const
     {
-        return (data < __rhs.data);
+        return iter_id < __rhs.iter_id;
```
samir-nasibli, line 113:
```suggestion
        assert(base == __rhs.base); // iterators are incomparable if base pointers are different
        return iter_id < __rhs.iter_id;
```
assert is needed.

PR #629, opened 2021-03-01
dpnpc_sum small refactoring

shssf:


densmirn (approved):

dpnp/backend/kernels/dpnp_krnl_reduction.cpp
```diff
@@ -60,11 +78,10 @@ void dpnp_sum_c(const void* input_in,
         return;
     }
 
-    const _DataType_output* initial_ptr = reinterpret_cast<const _DataType_output*>(initial);
-    const _DataType_output init = (initial_ptr == nullptr) ? _DataType_output{0} : *initial_ptr;
+    const _DataType_output init = get_initial_value<_DataType_output>(initial);
 
-    _DataType_input* input = reinterpret_cast<_DataType_input*>(const_cast<void*>(input_in));
-    _DataType_output* result = reinterpret_cast<_DataType_output*>(result_out);
+    _DataType_input* input = get_array_ptr<_DataType_input>(input_in);
```
samir-nasibli, line 83: you cast given const pointer to non-const, is it ok?
the whole point of const pointers is lost.

PR #630, opened 2021-03-01
Fix version of cmake greater than or equal to 3.16.5

densmirn:


shssf:
Why 3.16? The latest version ~3.20

shssf:
Also, please note https://github.com/IntelPython/dpnp/blob/ff2026accf71585c623c3a22a3d3ea804c7b5d6e/dpnp/backend/CMakeLists.txt#L82-L97

densmirn:
> Why 3.16? The latest version ~3.19.2

Took the version from the azure job `make_backend_lin`. Anyway currently I have conda packages conflicts with this and the latest version of cmake in internal CI.  Default version of cmake is 3.14 that breaks testing in internal CI. So I try to select working version of cmake.

shssf:
It looks like internal CI always takes us to stone age :-)

PR #632, opened 2021-03-01
Fix version of cmake greater than or equal to 3.16.5 in conda recipe

densmirn:


densmirn:
Already not required.

densmirn:
Already not required.

PR #634, opened 2021-03-02
Native all

Rubtsowa:


densmirn (commented):
Please let me know where is implementation of the function `dpnp_all_c`?

dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -0,0 +1,64 @@
...
+void dpnp_all_c(void* array1_in, void* result1, const size_t size)
+{
+    _DataType* array_in = reinterpret_cast<_DataType*>(array1_in);
+    _ResultType* result = reinterpret_cast<_ResultType*>(result1);
+
+    bool res = true;
+
+    for (size_t i = 0; i < size; ++i)
+    {
+        if (!bool(array_in[i]))
```
densmirn, line 43: Do we really need to cast input array elements to `bool`?

dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include "dpnp_fptr.hpp"
+#include "dpnp_iface.hpp"
+#include "queue_sycl.hpp"
+
+
+template <typename _DataType, typename _ResultType>
+void dpnp_all_c(void* array1_in, void* result1, const size_t size)
+{
+    _DataType* array_in = reinterpret_cast<_DataType*>(array1_in);
+    _ResultType* result = reinterpret_cast<_ResultType*>(result1);
```
samir-nasibli, line 37:
```suggestion
void dpnp_all_c(const void* array1_in, void* result1, const size_t size)
{
    const _DataType* array_in = reinterpret_cast<const _DataType*>(array1_in);
    _ResultType* result = reinterpret_cast<_ResultType*>(result1);
```

dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <iostream>
+
+#include "dpnp_fptr.hpp"
+#include "dpnp_iface.hpp"
+#include "queue_sycl.hpp"
+
+
+template <typename _DataType, typename _ResultType>
+void dpnp_all_c(void* array1_in, void* result1, const size_t size)
+{
```
samir-nasibli, line 35: Threre are no checks if given ptrs are `nullptr`.

samir-nasibli (changes requested):
Do we have any tests for check?

* list of numpy original tests (passed / disabled).
* also, provide, please, dpnp own tests, if numpy original tests do not cover some important cases for your function. (passed/disabled)

dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -0,0 +1,64 @@
...
+//*****************************************************************************
+
+#include <iostream>
+
+#include "dpnp_fptr.hpp"
+#include "dpnp_iface.hpp"
+#include "queue_sycl.hpp"
+
+
+template <typename _DataType, typename _ResultType>
```
samir-nasibli, line 33: Why do you use template type for the output, if it is `bool` for all instances?
It is no needed.

shssf (approved):

PR #635, opened 2021-03-02
MAINT: avoid cmpl warnings for `dpnp_diag_c`

samir-nasibli:


samir-nasibli:
@Rubtsowa please, do the same refactoring for other funcs.

samir-nasibli:
The another question is do we really need all these unused variables? )

shssf:
> The another question is do we really need all these unused variables? )

yes we do. And I would vote to all functions parameters variables to be implemented in API.

shssf (approved):

PR #636, opened 2021-03-03
add parameters to dpnp.prod

shssf:


dpnp/backend/kernels/dpnp_krnl_reduction.cpp
```diff
@@ -114,47 +114,90 @@ void dpnp_sum_c(void* result_out,
...
     {
         return;
     }
 
-    _DataType* array_1 = reinterpret_cast<_DataType*>(array1_in);
-    _DataType* result = reinterpret_cast<_DataType*>(result1);
+    const _DataType_output init = get_initial_value<_DataType_output>(initial, 1);
+
+    _DataType_input* input = get_array_ptr<_DataType_input>(input_in);
+    _DataType_output* result = get_array_ptr<_DataType_output>(result_out);
```
densmirn, line 148: We can avoid `const_cast` for `result_out`. Wouldn't we like to move `get_array_ptr` to utils maybe in separate PR?

dpnp/dpnp_iface_mathematical.py
```diff
@@ -93,6 +93,15 @@
 ]
 
 
+def convert_result_scalar(result, keepdims):
```
densmirn, line 96: I would add `to` or `2` to the function name, `convert_result_to_scalar` or `convert_result2scalar`.

densmirn (approved):
Good code!

Rubtsowa (approved):

dpnp/backend/kernels/dpnp_krnl_reduction.cpp
```diff
@@ -114,47 +114,90 @@ void dpnp_sum_c(void* result_out,
...
+    if ((input_in == nullptr) || (result_out == nullptr))
     {
         return;
     }
 
-    _DataType* array_1 = reinterpret_cast<_DataType*>(array1_in);
-    _DataType* result = reinterpret_cast<_DataType*>(result1);
+    const _DataType_output init = get_initial_value<_DataType_output>(initial, 1);
+
+    _DataType_input* input = get_array_ptr<_DataType_input>(input_in);
```
samir-nasibli, line 147: casting const pointer to non-const?

samir-nasibli (approved):
LGTM.

dpnp/backend/kernels/dpnp_krnl_reduction.cpp
shssf: "yes" and "yes" :-)
`const_cast` for `result` has no meaning but I use the function `get_array_ptr` to consistency.

dpnp/dpnp_iface_mathematical.py
shssf: I would vote for `convert_result2scalar` :-)

dpnp/backend/kernels/dpnp_krnl_reduction.cpp
shssf: yes. I understand It is bad. I didn't find a method how to use `const ptr` for MKL and PSTL usage

PR #639, opened 2021-03-05
Swap output and input parameter types in sum() and prod()

shssf:


samir-nasibli:
LGTM

PS by the way, which is better left- or right-hand traffic? :)

samir-nasibli (approved):

PR #642, opened 2021-03-09
Native astype

Rubtsowa:


dpnp/backend/kernels/dpnp_krnl_common.cpp
```diff
@@ -35,6 +35,42 @@
 namespace mkl_blas = oneapi::mkl::blas;
 namespace mkl_lapack = oneapi::mkl::lapack;
 
+template <typename _DataType, typename _TypeNew, typename _ResultType>
```
samir-nasibli, line 38: what is `_TypeNew` for?

dpnp/backend/kernels/dpnp_krnl_common.cpp
```diff
@@ -35,6 +35,42 @@
...
+class dpnp_astype_c_kernel;
+
+template <typename _DataType, typename _TypeNew, typename _ResultType>
+void dpnp_astype_c(void* array1_in, void* result1, size_t size)
```
samir-nasibli, line 42: remember this interface only for `unsafe` casting kind.

dpnp/backend/kernels/dpnp_krnl_common.cpp
```diff
@@ -35,6 +35,42 @@
...
+{
+    cl::sycl::event event;
+
+    _DataType* array_in = reinterpret_cast<_DataType*>(array1_in);
+    _ResultType* result = reinterpret_cast<_ResultType*>(result1);
```
samir-nasibli, line 47:
```suggestion
void dpnp_astype_c(const void* array1_in, void* result1, const size_t size)
{
    cl::sycl::event event;

    const _DataType* array_in = reinterpret_cast<const _DataType*>(array1_in);
    _ResultType* result = reinterpret_cast<_ResultType*>(result1);
```
Try not to ignore suggestions from previous PRs.

dpnp/dparray.pyx
```diff
@@ -788,19 +790,21 @@ cdef class dparray:
...
+                pass
+            elif copy is not True:
+                pass
+            elif order is not 'K':
+                pass
+            else:
+                return dpnp_astype(self, dtype)
 
-        return dpnp_astype(self, dtype)
+        return call_origin(numpy.ndarray.astype, self, dtype, order, casting, subok, copy)
```
samir-nasibli, line 807: I think doing fallback on `numpy.ndarray` here is bad idea.
It is better just throw error.

tests_external/skipped_tests_numpy.tbl
```diff
@@ -2881,6 +2881,7 @@ tests/test_umath.py::test_rint_big_int
 tests/test_umath.py::TestRoundingFunctions::test_object_direct
 tests/test_umath.py::TestRoundingFunctions::test_object_indirect
 tests/test_umath.py::test_signaling_nan_exceptions
+tests/test_umath.py::TestSign::test_sign_dtype_nan_object
```
samir-nasibli, line 2884: what is the reason?
Provide info on description, please.

tests/test_dparray.py
```diff
@@ -0,0 +1,20 @@
...
+import pytest
+
+
+@pytest.mark.parametrize("res_dtype",
+                         [numpy.float64, numpy.float32, numpy.int64, numpy.int32, numpy.bool, numpy.bool_, numpy.complex],
+                         ids=['float64', 'float32', 'int64', 'int32', 'bool', 'bool_', 'complex'])
+@pytest.mark.parametrize("arr_dtype",
+                         [numpy.float64, numpy.float32, numpy.int64, numpy.int32, numpy.bool, numpy.bool_, numpy.complex],
+                         ids=['float64', 'float32', 'int64', 'int32', 'bool', 'bool_', 'complex'])
+@pytest.mark.parametrize("arr",
```
samir-nasibli, line 12: what if array is empty?
Corner cases.

tests/skipped_tests_gpu.tbl
```diff
@@ -1,3 +1,17 @@
...
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-bool]
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-bool_]
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-complex]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-float64]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-float32]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-int64]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-int32]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-bool]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-bool_]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-complex]
```
samir-nasibli, line 14: please, provide some info why are you skipping all this tests?

tests/skipped_tests.tbl
```diff
@@ -1,3 +1,17 @@
...
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-bool]
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-bool_]
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-complex]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-float64]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-float32]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-int64]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-int32]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-bool]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-bool_]
+tests/test_dparray.py::test_astype[[[-2, -1], [1, 2]]-complex-complex]
```
samir-nasibli, line 14: what is the reason?

samir-nasibli (changes requested):
Provide also info about tests, that you are disabling.

tests/skipped_tests.tbl
Rubtsowa: because 

error: assigning to 'bool' from incompatible type 'const std::complex<double>'
error: assigning to 'int' from incompatible type 'const std::complex<double>'
error: assigning to 'long' from incompatible type 'const std::complex<double>'
error: assigning to 'float' from incompatible type 'const std::complex<double>'
error: assigning to 'double' from incompatible type 'const std::complex<double>'

samir-nasibli:
So seems that, you are disabling even more tests for changes proposed by you.
You need to investigate them.

tests/skipped_tests.tbl
samir-nasibli: Seems like they have the same issue. Investigation and fix is needed.

Rubtsowa:
@samir-nasibli  you sure that I am disabling even more tests?
I added 196 tests, 20 of them skipped(not converted many types in complex in c++ ). And I uskip 1 test and 1 skip in existing tests.

tests/skipped_tests_gpu.tbl
```diff
@@ -1,3 +1,24 @@
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-float64]
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-float32]
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-int64]
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-int32]
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-bool]
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-bool_]
+tests/test_dparray.py::test_astype[[-2, -1, 0, 1, 2]-complex-complex]
```
samir-nasibli, line 7: Complex to complex doesn't works?

tests_external/skipped_tests_numpy.tbl
Rubtsowa: This test failed with error:
DPNP error: in function dpnp_dtype_to_DPNPFuncType() type '<class 'object'>' is not supported

samir-nasibli: `object` is a (global) variable. By default it is bound to a built-in class which is the root of the type hierarchy.

`dpnp_dtype_to_DPNPFuncType` can not convert `object` to `dpnpc` function types (DPNP_FT_DOUBLE, DPNP_FT_FLOAT, DPNP_FT_LONG, ...). 
See this internal func impl [here (https://github.com/IntelPython/dpnp/blob/master/dpnp/dpnp_algo/dpnp_algo.pyx#L256:L271)

You need fallback on `numpy` for case when dtype is object.

samir-nasibli: > You can fallback on numpy for case when dtype is object.

Sorry, my fault. It is bad proposal, just leave it as is.
No need doing fallback on `numpy` for dparray own methods.

tests/skipped_tests.tbl
```diff
@@ -154,7 +168,6 @@ tests/test_linalg.py::test_svd[(5,3)-complex128]
 tests/test_random.py::TestPermutationsTestShuffle::test_shuffle1[ lambda x: x]
 tests/test_random.py::TestPermutationsTestShuffle::test_shuffle1[ lambda x: dpnp.asarray(x).astype(dpnp.int8)]
 tests/test_random.py::TestPermutationsTestShuffle::test_shuffle1[ lambda x: dpnp.asarray(x).astype(dpnp.complex64)]
-tests/test_random.py::TestPermutationsTestShuffle::test_shuffle1[ lambda x: dpnp.asarray(x).astype(object)]
```
samir-nasibli, line 157: :

tests/test_dparray.py
samir-nasibli: outdated.

samir-nasibli (approved):
Looks pretty good. Thanks!

samir-nasibli:
Issues with `complex` type can be solved on other PR.

dpnp/dpnp_algo/dpnp_algo.pyx
```diff
@@ -125,10 +126,16 @@ cpdef dparray dpnp_array(obj, dtype=None):
 
 
 cpdef dparray dpnp_astype(dparray array1, dtype_target):
-    cdef dparray result = dparray(array1.shape, dtype=dtype_target)
+    cdef DPNPFuncType param1_type = dpnp_dtype_to_DPNPFuncType(array1.dtype)
+    cdef DPNPFuncType param2_type = dpnp_dtype_to_DPNPFuncType(dtype_target)
```
shssf, line 130: `dtype_target` might be diffrent. It can contain some object with no `dtype`. I think this is the cause of the error you saw

shssf (approved):

PR #651, opened 2021-03-16
add trace

Rubtsowa:


samir-nasibli (commented):
your tests do not match the function logic.
Please, address the comments above

tests/test_arraycreation.py
```diff
@@ -138,9 +138,6 @@ def test_loadtxt(type):
         numpy.testing.assert_array_equal(dpnp_res, np_res)
 
 
-@pytest.mark.parametrize("dtype",
```
samir-nasibli: Please, don't remove your test cases.
Just disable them by adding in skip list.

They may be enabled in the future.

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -1059,7 +1059,9 @@ def trace(arr, offset=0, axis1=0, axis2=1, dtype=None, out=None):
             pass
         elif axis2 != 1:
             pass
-        elif out != None:
+        elif dtype is not None:
```
samir-nasibli, line 1062: You discard cases where `dtype` matches with `arr.dtype`.

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -1059,7 +1059,9 @@ def trace(arr, offset=0, axis1=0, axis2=1, dtype=None, out=None):
...
+            pass
+        elif out is not None:
```
samir-nasibli, line 1064: You discard cases where `out.dtype` matches with `arr.dtype`.

samir-nasibli (changes requested):
Duplicating. see https://github.com/IntelPython/dpnp/pull/651#discussion_r596397873

`TeamCity Build SAT` is failed.

dpnp/dpnp_iface_arraycreation.py
shssf: I think it is acceptable for this stage of development

shssf (approved):

PR #652, opened 2021-03-17
add identity function

KsanaKozlova:


dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
```diff
@@ -110,6 +110,25 @@ void dpnp_full_like_c(void* array_in, void* result, const size_t size)
...
+        for (size_t j = 0; j < n; ++j)
+        {
+            if (i == j)
+            {
+                result[i * n + j] = 1;
+            } else {
+                result[i * n + j] = 0;
+            }
+        }
+    }
```
densmirn, line 129: We could parallelize it with SYCL if we rolled up nested loops to a single one as below
```suggestion
    for (size_t i = 0; i < n * n; ++i)
    {
        if (i / n == i % n)
        {
            result[i] = 1;
        }
        else
        {
            result[i] = 0;
        }
    }
```

dpnp/backend/include/dpnp_iface.hpp
```diff
@@ -485,6 +485,16 @@ template <typename _DataType>
 INP_DLLEXPORT void dpnp_diagonal_c(
     void* array1_in, void* result1, const size_t offset, size_t* shape, size_t* res_shape, const size_t res_ndim);
 
+/**
+ * @ingroup BACKEND_API
+ * @brief Implementation of identity function
+ *
+ * @param [out] result1   Output array.
+ * @param [in]  n         Number of rows (and columns) in n x n output..
```
samir-nasibli, line 493:
```suggestion
 * @param [in]  n         Number of rows (and columns) in n x n output.
```

dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
samir-nasibli: We can use 2-dim range here if it is needed.

samir-nasibli: kernel might look like this:
```cpp
    cl::sycl::range<2> gws(size, size);
    auto kernel_parallel_for_func = [=](cl::sycl::id<2> global_id) {
        size_t i = global_id[0];
        size_t j = global_id[1];
        result[i * n + j] = (i == j) ? 1 : 0;
    };
```

But before this you need check on `size`. It should be non-zero.

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -710,6 +711,38 @@ def geomspace(start, stop, num=50, endpoint=True, dtype=None, axis=0):
...
+    >>> import dpnp as np
+    >>> np.identity(3)
+    array([[1.,  0.,  0.],
+           [0.,  1.,  0.],
+           [0.,  0.,  1.]])
+
+    """
+    if not use_origin_backend():
+        if like is not None:
+            pass
```
samir-nasibli, line 737:
```suggestion
        if n < 0:
            pass
        elif like is not None:
            pass
```
`n` should be non-negative.

dpnp/backend/kernels/dpnp_krnl_arraycreation.cpp
```diff
@@ -110,6 +110,45 @@ void dpnp_full_like_c(void* array_in, void* result, const size_t size)
...
+            if (i / n == i % n)
+            {
+                result[i] = 1;
+            }
+            else
+            {
+                result[i] = 0;
+            }
+        }
+    };
```
samir-nasibli, line 141: Try this.
```suggestion
    cl::sycl::range<2> gws(n, n);
    auto kernel_parallel_for_func = [=](cl::sycl::id<2> global_id) {
        size_t i = global_id[0];
        size_t j = global_id[1];
        result[i * n + j] = (i == j) ? 1 : 0;
    };
```
```

**Review by densmirn — APPROVED** [review]

File: dpnp/dpnp_iface_arraycreation.py (hunk 3 above)
Comment (samir-nasibli) [reply dpnp/dpnp_iface_arraycreation.py:738]: outdated.

**Review by samir-nasibli — APPROVED** [review]

File: dpnp/dpnp_iface_arraycreation.py [hunk 5]
```diff
@@ -711,6 +712,40 @@ def geomspace(start, stop, num=50, endpoint=True, dtype=None, axis=0):
...
+        if n < 0:
```
densmirn, line 739:
```suggestion
        elif n < 0:
```
?

shssf (approved):

PR #653, opened 2021-03-19
Partition

Rubtsowa:


dpnp/dpnp_algo/dpnp_algo_sorting.pyx
```diff
@@ -39,13 +39,34 @@ from dpnp.dpnp_utils cimport *
...
 
 
+ctypedef void(*fptr_dpnp_partition_t)(const void * , const void * , void * , const size_t , const size_t * , const size_t)
+
+
 cpdef dparray dpnp_argsort(dparray in_array1):
     return call_fptr_1in_1out(DPNP_FN_ARGSORT, in_array1, in_array1.shape)
 
 
+cpdef dparray dpnp_partition(dparray arr, int kth, axis=-1, kind='introselect', order=None):
```
samir-nasibli, line 54: we can use enums or just int vals for `kind` param in our cython layer

dpnp/dpnp_iface_sorting.py
```diff
@@ -101,6 +106,36 @@ def argsort(in_array1, axis=-1, kind=None, order=None):
     return numpy.argsort(in_array1, axis, kind, order)
 
 
+def partition(arr, kth, axis=-1, kind='introselect', order=None):
```
samir-nasibli, line 109: use `self` for accessing the attributes and methods of the instance of the class.

dpnp/dparray.pyx
```diff
@@ -942,6 +942,34 @@ cdef class dparray:
         """
         return argsort(self, axis, kind, order)
 
+    def partition(arr, kth, axis=-1, kind='introselect', order=None):
```
samir-nasibli, line 945: use `self` for accessing the attributes and methods of the instance of the class.

dpnp/dpnp_iface_sorting.py
```diff
@@ -40,15 +40,20 @@
 """
 
 
+import collections
+
```
samir-nasibli, line 44: remove it.

dpnp/dparray.pyx
```diff
@@ -942,6 +942,34 @@ cdef class dparray:
         """
         return argsort(self, axis, kind, order)
 
+    def partition(self, kth, axis=-1, kind='introselect', order=None):
```
shssf, line 945: It is not quite clear for me, why we need this code here. Why we can't call dpnp.partition() instead implementation it here in dparray.

dpnp/backend/kernels/dpnp_krnl_sorting.cpp
```diff
@@ -83,6 +83,55 @@ struct _sort_less
...
+    const _DataType* sort_arr = reinterpret_cast<const _DataType*>(sort_array1_in);
+    _DataType* result = reinterpret_cast<_DataType*>(result1);
+
+    size_t size_ = 1;
+    for (size_t i = 0; i < ndim - 1; ++i)
+    {
+        size_ *= shape[i];
+    }
+
+    cl::sycl::range<2> gws(size_, kth+1);
```
shssf, line 104: In general case, the `shape[]` can contains zero as one of the dimension. It cause `size_` be zero. It will failed with GPU.
Please put "if()" to check GWS dimensions non zero in front of this line.

dpnp/backend/kernels/dpnp_krnl_sorting.cpp
```diff
@@ -83,6 +83,55 @@ struct _sort_less
...
+void dpnp_partition_c(const void* sort_array1_in, void* result1, const size_t kth, const size_t* shape, const size_t ndim)
+{
+
+    cl::sycl::event event;
+
+    const _DataType* sort_arr = reinterpret_cast<const _DataType*>(sort_array1_in);
+    _DataType* result = reinterpret_cast<_DataType*>(result1);
+
+    size_t size_ = 1;
+    for (size_t i = 0; i < ndim - 1; ++i)
```
shssf, line 99: `ndim - 1` is suspicious for me. It looks like `size_` represents number of elements in the array. If `ndim==1` the variable `size` will have `1` instead `shape[0]`

PR #654, opened 2021-03-19
Wheel changes to master

Vyacheslav-Smirnov:
- Update README
 - Set RPATH to dpctl and dpnp itself
 - Fix in setup.py


conda-recipe/build.sh
```diff
@@ -25,6 +25,10 @@ if [ -n "${TBBROOT}" ]; then
     . ${TBBROOT}/env/vars.sh
 fi
 
+# Set RPATH for wheels
+export CFLAGS="-Wl,-rpath,\$ORIGIN/../dpctl,-rpath,\$ORIGIN $CFLAGS"
```
shssf, line 29: How it will work in Windows?

setup.py
```diff
@@ -183,13 +180,13 @@
...
       long_description=__readme_file__,
+      long_description_content_type="text/markdown",
       author="Intel Corporation",
-      author_email="Intel Corporation",
       maintainer="Intel Corp.",
       maintainer_email="scripting@intel.com",
       url="https://intelpython.github.io/dpnp/",
       download_url="https://github.com/IntelPython/dpnp",
-      license=__license_file__,
+      license='BSD',
```
shssf, line 189: Is this change related to "wheel testing"?

conda-recipe/build.sh
Vyacheslav-Smirnov: The "build.sh" file is used only for unix build. For windows there should be bld.bat with build commands.
As there is no bld.bat file, windows build now is just fake)

setup.py
Vyacheslav-Smirnov: Yes, changing setup.py format is required to build a good wheel which can be uploaded to Pypi

Vyacheslav-Smirnov:
@shssf , Are you ok with the changes?

Vyacheslav-Smirnov:
Will be merged in scope of https://github.com/IntelPython/dpnp/pull/657

PR #655, opened 2021-03-19
MULTIPLY enable broadcasting

densmirn:
This PR shows current status of broadcasting implementation through USM iterator.

Currently issue below is under investigation:
```
DPNP_QUEUE_GPU=1 python -c "import dpnp; a = dpnp.array([7]); b = dpnp.array([7]); c = dpnp.multiply(a, b); print(c)"
Available SYCL devices:
 - id=4466, type=8, gws=67108864, cu=12, name=Intel(R) FPGA Emulation Device
 - id=32902, type=2, gws=8192, cu=12, name=Intel(R) Core(TM) i7-10710U CPU @ 1.10GHz
 - id=32902, type=4, gws=256, cu=24, name=Intel(R) Graphics [0x9bca]
 - id=32902, type=4, gws=256, cu=24, name=Intel(R) Graphics [0x9bca]
 - id=32902, type=18, gws=1, cu=1, name=SYCL host device
Running on: Intel(R) Graphics [0x9bca]
queue initialization time: 0.000503686 (sec.)
SYCL kernels link time: 0.112728 (sec.)

terminate called after throwing an instance of 'cl::sycl::runtime_error'
  what():  Native API failed. Native API returns: -30 (CL_INVALID_VALUE) -30 (CL_INVALID_VALUE)
Aborted (core dumped)
```

### DIFF
--- a/dpnp/backend/kernels/dpnp_krnl_elemwise.cpp
+++ b/dpnp/backend/kernels/dpnp_krnl_elemwise.cpp
@@ -28,6 +28,7 @@
 
 #include <dpnp_iface.hpp>
 #include "dpnp_fptr.hpp"
+#include "dpnp_iterator.hpp"
 #include "dpnp_utils.hpp"
 #include "queue_sycl.hpp"
 
@@ -353,10 +354,6 @@ static void func_map_init_elemwise_1arg_1type(func_map_t& fmap)
                   const size_t* where)                                                                                 \
     {                                                                                                                  \
         /* avoid warning unused variable*/                                                                             \
-        (void)input1_shape;                                                                                            \
-        (void)input1_shape_ndim;                                                                                       \
-        (void)input2_shape;                                                                                            \
-        (void)input2_shape_ndim;                                                                                       \
         (void)where;                                                                                                   \
                                                                                                                        \
         if (!input1_size || !input2_size)                                                                              \
@@ -364,18 +361,31 @@ static void func_map_init_elemwise_1arg_1type(func_map_t& fmap)
             return;                                                                                                    \
         }                                                                                                              \
                                                                                                                        \
-        const size_t result_size = (input2_size > input1_size) ? input2_size : input1_size;                            \
-                                                                                                                       \
-        const _DataType_input1* input1_data = reinterpret_cast<const _DataType_input1*>(input1_in);                    \
-        const _DataType_input2* input2_data = reinterpret_cast<const _DataType_input2*>(input2_in);                    \
+        _DataType_input1* input1_data = reinterpret_cast<_DataType_input1*>(const_cast<void*>(input1_in));             \
+        _DataType_input2* input2_data = reinterpret_cast<_DataType_input2*>(const_cast<void*>(input2_in));             \
         _DataType_output* result = reinterpret_cast<_DataType_output*>(result_out);                                    \
                                                                                                                        \
+        std::vector<size_t> result_shape = get_common_shape(input1_shape, input1_shape_ndim,                           \
+                                                            input2_shape, input2_shape_ndim);                          \
+                                                                                                                       \
+        DPNPC_id<_DataType_input1> input1(input1_data, input1_shape, input1_shape_ndim);                               \
+        input1.broadcast(result_shape);                                                                                \
+                                                                                                                       \
+        DPNPC_id<_DataType_input2> input2(input2_data, input2_shape, input2_shape_ndim);                               \
+        input2.broadcast(result_shape);                                                                                \
+                                                                                                                       \
+        const size_t result_size = input1.get_output_size();                                                           \
+                                                                                                                       \
         cl::sycl::range<1> gws(result_size);                                                                           \
+        const DPNPC_id<_DataType_input1>* input1_it = &input1;                                                         \
+        const DPNPC_id<_DataType_input2>* input2_it = &input2;                                                         \
         auto kernel_parallel_for_func = [=](cl::sycl::id<1> global_id) {                                               \
-            size_t i = global_id[0]; /*for (size_t i = 0; i < result_size; ++i)*/                                      \
-            const _DataType_output input1_elem = (input1_size == 1) ? input1_data[0] : input1_data[i];                 \
-            const _DataType_output input2_elem = (input2_size == 1) ? input2_data[0] : input2_data[i];                 \
-            result[i] = __operation1__;                                                                                \
+            const size_t i = global_id[0]; /*for (size_t i = 0; i < result_size; ++i)*/                                \
+            {                                                                                                          \
+                const _DataType_output input1_elem = (*input1_it)[i];                                                  \
+                const _DataType_output input2_elem = (*input2_it)[i];                                                  \
+                result[i] = __operation1__;                                                                            \
+            }                                                                                                          \
         };                                                                                                             \
         auto kernel_func = [&](cl::sycl::handler& cgh) {                                                               \
             cgh.parallel_for<class __name__##_kernel<_DataType_output, _DataType_input1,                               \
@@ -390,9 +400,7 @@ static void func_map_init_elemwise_1arg_1type(func_map_t& fmap)
                            std::is_same<_DataType_input1, float>::value) &&                                            \
                           std::is_same<_DataType_input2, _DataType_input1>::value)                                     \
             {                                                                                                          \
-                _DataType_input1* input1 = const_cast<_DataType_input1*>(input1_data);                                 \
-                _DataType_input2* input2 = const_cast<_DataType_input2*>(input2_data);                                 \
-                event = __operation2__(DPNP_QUEUE, result_size, input1, input2, result);                               \
+                event = __operation2__(DPNP_QUEUE, result_size, input1_data, input2_data, result);                     \
             }                                                                                                          \
             else                                                                                                       \
             {                                                                                                          \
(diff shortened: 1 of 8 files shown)


dpnp/backend/kernels/dpnp_krnl_elemwise.cpp
```diff
@@ -353,29 +354,38 @@ static void func_map_init_elemwise_1arg_1type(func_map_t& fmap)
...
-        const size_t result_size = (input2_size > input1_size) ? input2_size : input1_size;                            \
-                                                                                                                       \
-        const _DataType_input1* input1_data = reinterpret_cast<const _DataType_input1*>(input1_in);                    \
-        const _DataType_input2* input2_data = reinterpret_cast<const _DataType_input2*>(input2_in);                    \
+        _DataType_input1* input1_data = reinterpret_cast<_DataType_input1*>(const_cast<void*>(input1_in));             \
+        _DataType_input2* input2_data = reinterpret_cast<_DataType_input2*>(const_cast<void*>(input2_in));             \
         _DataType_output* result = reinterpret_cast<_DataType_output*>(result_out);                                    \
                                                                                                                        \
+        std::vector<size_t> result_shape = get_common_shape(input1_shape, input1_shape_ndim,                           \
```
shssf, line 368: maybe "get_result_shape" is better for this function name? not sure...

dpnp/backend/kernels/dpnp_krnl_elemwise.cpp
```diff
@@ -353,29 +354,38 @@ static void func_map_init_elemwise_1arg_1type(func_map_t& fmap)
...
+                                                            input2_shape, input2_shape_ndim);                          \
+                                                                                                                       \
+        DPNPC_id<_DataType_input1> input1(input1_data, input1_shape, input1_shape_ndim);                               \
+        input1.broadcast(result_shape);                                                                                \
```
shssf, line 372: "broadcast_to_shape" func name?

shssf:
gtests needs to be implemented to test "broadcast" feature in iterator.
Perhaps, it is better to start with gtests and later use "broadcast iterator" in real algorithms

dpnp/backend/tests/test_broadcast_iterator.cpp
```diff
@@ -0,0 +1,202 @@
...
+// THE POSSIBILITY OF SUCH DAMAGE.
+//*****************************************************************************
+
+#include <gtest/gtest.h>
+#include <numeric>
+#include <vector>
+
+#include "dpnp_iterator.hpp"
+
+#define DPNP_LOCAL_QUEUE 1 // TODO need to fix build procedure and remove this workaround. Issue #551
```
shssf, line 32: @samir-nasibli it looks like it requires your attention

dpnp/backend/tests/test_broadcast_iterator.cpp
```diff
@@ -0,0 +1,202 @@
...
+#include "queue_sycl.hpp"
+
+using namespace std;
+using dpnpc_it_t = DPNPC_id<size_t>::iterator;
+using dpnpc_value_t = dpnpc_it_t::value_type;
+using dpnpc_index_t = dpnpc_it_t::size_type;
+
+template <typename _DataType>
+vector<_DataType> get_input_data(const vector<dpnpc_index_t>& shape)
```
shssf, line 41: code duplication. It looks like it is better to move this function into some common place for test suite.

dpnp/backend/tests/test_broadcast_iterator.cpp
```diff
@@ -0,0 +1,202 @@
...
+    DPNPC_id<dpnpc_value_t> result_obj(input_data.data(), {3, 1});
+    result_obj.broadcast_to_shape({3, 4});
+
+    for (dpnpc_index_t output_id = 0; output_id < result_obj.get_output_size(); ++output_id)
+    {
+        EXPECT_EQ(result_obj[output_id], expected_data[output_id]);
+    }
+}
+
+TEST(TestBroadcastIterator, take_value_broadcast_loop_3D)
```
shssf, line 65: It is not clear why these tests are not with `IteratorParameters` variadic tests

dpnp/backend/tests/test_broadcast_iterator.cpp
```diff
@@ -0,0 +1,202 @@
...
+    result_obj.broadcast_to_shape({2, 3, 4});
+
+    const dpnpc_index_t output_size = result_obj.get_output_size();
+    EXPECT_EQ(output_size, 24);
+}
+
+struct IteratorParameters
+{
+    vector<dpnpc_it_t::size_type> input_shape;
+    vector<dpnpc_it_t::size_type> output_shape;
```
shssf, line 92: Is it defined or standardized somewhere? I mean, `output_shape` as an input.
I see, at least, two approaches:
1. we keep input internally: `input_shape` and `output_shape`. Also we internally calculate step distances and other things to do iteration.
2. we keep input internally: `input_shape` and `broatcast_axis`. Also we internally calculate `output_shape`, step distances and other things to do iteration.

#2 will match existed interface code (to function `set_axes`) and, perhaps, required some extra parameter in ctor() like:
`DPNPC_id<dpnpc_value_t> result_obj(input_data.data(), {3, 4}, BROADCAST);`

dpnp/backend/tests/test_broadcast_iterator.cpp
```diff
@@ -0,0 +1,202 @@
...
+ * output = np.ones(output_shape, dtype=np.int64)
+ * print(f"output shape={output.shape}")
+ *
+ * result = input * output
+ * print(f"result={np.array2string(result.reshape(result.size), separator=', ')}\n", sep=", ")
+ */
+INSTANTIATE_TEST_SUITE_P(
+    TestBroadcastIterator,
+    IteratorBroadcasting,
+    testing::Values(
```
shssf, line 186: 👍 
Do you think it is good to add some combinations like "{1, 0}. {0}"?

dpnp/backend/tests/test_broadcast_iterator.cpp
shssf: I don't know what complexity you are ready to implement and NumPy is required.
as far as I understand, we have two input shapes ("several" in general). Let's take:
```
shape1={2, 3, 4}
shape2={3, 4}
```
it looks like we can broadcast `shape2` into `shape1` by two steps
```
1. [3, 4] => [1, 3, 4]
2. [1, 3, 4] => [2, 3, 4]
```
Also, as I understand, shapes [3, 4] and [3, 1, 4, 1] have the same layout in memory.
I think everything depends on input parameters. If we have shapes as input - it is ok. I think having `axis` to broadcast is more flexible and scalable solution.

shssf: From another perspective, it is NOT good idea.
May be better leave it as is but make function names more clear.

dpnp/backend/tests/test_broadcast_iterator.cpp
densmirn: I wanted to add something like this `IteratorParameters{{2, 0, 1}, {2, 0, 4}, {}}));` as test parameter, but in this case we have empty result. So SYCL kernel doesn't make sense here, because we iterate by result size that is equal to 0 in this case. In real case we should return from the function before submitting such kernel to the queue. Moreover we can see below issue on GPU when submitting such kernel to the queue
`C++ exception with description "Native API failed. Native API returns: -30 (CL_INVALID_VALUE) -30 (CL_INVALID_VALUE)" thrown in the test body.`
That is why 2 tests on reduction fail https://github.com/IntelPython/dpnp/pull/661#issuecomment-810387036.

dpnp/backend/kernels/dpnp_krnl_elemwise.cpp
```diff
@@ -405,6 +417,9 @@ static void func_map_init_elemwise_1arg_1type(func_map_t& fmap)
         }                                                                                                              \
                                                                                                                        \
         event.wait();                                                                                                  \
+                                                                                                                       \
+        input1_it->~DPNPC_id();                                                                                        \
+        input2_it->~DPNPC_id();                                                                                        \
```
samir-nasibli, line 422: No needed here.

dpnp/backend/tests/test_broadcast_iterator.cpp
```diff
@@ -158,8 +105,12 @@ TEST_P(IteratorBroadcasting, sycl_broadcast)
 
     for (dpnpc_index_t i = 0; i < result_size; ++i)
     {
-        EXPECT_EQ(result.at(i), param.result.at(i));
+        EXPECT_EQ(result[i], param.result.at(i));
     }
+
+    input_it->~DPNPC_id();
```
samir-nasibli, line 111: no need to explicitly call the destructor.

densmirn: Needed for `placement new`.

dpnp/backend/kernels/dpnp_krnl_elemwise.cpp
samir-nasibli: These obvious things that you are saying, but a little bit not about that.

The comment was not about freeing resources in general, it is about that we should avoid such explicitly call the destructor.
I see that current iterator implementation has flaws, updates for to the iterator interface are required for this.
In anyway this is not the current PR problem.

shssf (approved):

PR #658, opened 2021-03-22
ITERATOR free memory when reset axes

densmirn:


densmirn:
@shssf do you think it's valid case or not?

shssf:
yes, but no "if" clause needed. Also, other variables are better to revert to default.

densmirn:
All other variables will be recalculated with axes reset, so I'm not sure we need to revert these ones.

shssf (approved):

PR #681, opened 2021-04-14
ENH: kernels for `random.vonmisses`

samir-nasibli:
## Description
[WIP]

### DIFF
--- a/dpnp/backend/kernels/dpnp_krnl_random.cpp
+++ b/dpnp/backend/kernels/dpnp_krnl_random.cpp
@@ -1234,6 +1234,9 @@ void dpnp_rng_uniform_c(void* result, const long low, const long high, const siz
 #define M_PI 3.141592653589793238462643383279502884197
 #endif
 
+template <typename _KernelNameSpecialization>
+class dpnp_acceptance_vonmises_large_kappa_c_kernel;
+
 template <typename _DataType>
 void dpnp_rng_vonmises_large_kappa_c(void* result, const _DataType mu, const _DataType kappa, const size_t size)
 {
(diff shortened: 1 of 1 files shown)


dpnp/backend/kernels/dpnp_krnl_random.cpp
```diff
@@ -1242,65 +1243,70 @@ void dpnp_rng_vonmises_large_kappa_c(void* result, const _DataType mu, const _Da
 
     Uvec = reinterpret_cast<_DataType*>(dpnp_memory_alloc_c(size * sizeof(_DataType)));
     Vvec = reinterpret_cast<_DataType*>(dpnp_memory_alloc_c(size * sizeof(_DataType)));
-
-    for (size_t n = 0; n < size;)
+    n = reinterpret_cast<size_t*>(dpnp_memory_alloc_c(sizeof(size_t)));
```
shssf, line 1246: this is quite strange (Make scalar as a array with one element).
I think it should be a scalar, not an array.

dpnp/backend/kernels/dpnp_krnl_random.cpp
```diff
@@ -1242,65 +1243,70 @@ void dpnp_rng_vonmises_large_kappa_c(void* result, const _DataType mu, const _Da
...
+                Y = kappa * (s_minus_one + neg_W_minus_one);
 
+                if ((Y * (2 - Y) >= V) || (cl::sycl::log(Y / V) + 1 >= Y))
+                {
+                    Y = neg_W_minus_one * (2 - neg_W_minus_one);
+                    if (Y < 0)
+                        Y = 0.0;
+                    else if (Y > 1.0)
+                        Y = 1.0;
+                    n[0] = n[0] + 1;
```
shssf, line 1281: This is a mistake. This is parallel environment (SYCL kernel). Writing inside the kernel into same memory cause https://en.wikipedia.org/wiki/Race_condition

dpnp/backend/kernels/dpnp_krnl_random.cpp
```diff
@@ -1242,65 +1243,70 @@ void dpnp_rng_vonmises_large_kappa_c(void* result, const _DataType mu, const _Da
...
-            _DataType sn, cn, sn2, cn2;
-            _DataType neg_W_minus_one, V, Y;
+        cl::sycl::range<1> diff_gws(diff_size);
 
-            sn = sin(Uvec[i]);
-            cn = cos(Uvec[i]);
-            V = Vvec[i];
-            sn2 = sn * sn;
-            cn2 = cn * cn;
+        auto paral_kernel_some = [&](cl::sycl::handler& cgh) {
```
shssf, line 1257: Kernel inside the loop with bigger trip count. It would be more efficient to parallelize (make kernel) the algorithm by bigger value `size` instead `size-n`. So, it will require a loop inside the kernel.
It is questionable what will be more performant
1. loop with a kernels queue
2. kernel with a loop

It is hard to predict it with no perf measurements but I would vote that parallelization with bigger number of threads should be better.

dpnp/backend/kernels/dpnp_krnl_random.cpp
samir-nasibli: scalar and this is the same. You just can not pass n to sycl region in other way.

shssf:
@samir-nasibli Is this PR ready to review or still in development stage?

samir-nasibli:
> @samir-nasibli Is this PR ready to review or still in development stage?

I will update this PR or move some part of this changes to another PR with closing this.

dpnp/backend/kernels/dpnp_krnl_random.cpp
```diff
@@ -1276,49 +1277,50 @@ void dpnp_rng_vonmises_large_kappa_c(void* result, const _DataType mu, const _Da
...
-        }
+                if ((Y * (2 - Y) >= V) || (cl::sycl::log(Y / V) + 1 >= Y))
+                {
+                    Y = neg_W_minus_one * (2 - neg_W_minus_one);
+                    if (Y < 0)
+                        Y = 0.0;
+                    else if (Y > 1.0)
+                        Y = 1.0;
+                    *n = *n + 1;
+                    result1[*n] = cl::sycl::asin(cl::sycl::sqrt(Y));
```
Alexander-Makaryev, line 1315: Looks like here we are getting race condition, that is why we are getting wrong results. To prevent it we should calculate `n` (index of result) from `i`.

dpnp/backend/kernels/dpnp_krnl_random.cpp
```diff
@@ -1371,36 +1374,39 @@ void dpnp_rng_vonmises_small_kappa_c(void* result, const _DataType mu, const _Da
...
+                size_t i = global_id[0];
+                _DataType Z, W, Y, V;
+                Z = cl::sycl::cos(Uvec[i]);
+                V = Vvec[i];
+                W = (kappa + s_kappa * Z) / (s_kappa + kappa * Z);
+                Y = s_kappa - kappa * W;
+                if ((Y * (2 - Y) >= V) || (cl::sycl::log(Y / V) + 1 >= Y))
+                {
+                    *n = *n + 1;
+                    result1[*n] = cl::sycl::acos(W);
```
Alexander-Makaryev, line 1401: Looks like here we are getting race condition, that is why we are getting wrong results. To prevent it we should calculate n (index of result) from i.

samir-nasibli:
@LukichevaPolina
The use of additional memory with the amount of data is not a good practice in optimization. We must avoid this cases.
We have to remove the possibilities for a potential race condition in the algorithm.

oleksandr-pavlyk:
Stale PR?

PR #683, opened 2021-04-21
Enable tests on special GPU device

densmirn:


samir-nasibli:
Do we really need this resolution for failed tests?
@densmirn provide, please, more detailed description for this case.
It is better create issue ticket for that.

tests/conftest.py
```diff
@@ -49,13 +48,12 @@ def pytest_collection_modifyitems(config, items):
...
 
-    current_queue_is_cpu = dpnp.dpnp_queue_is_cpu()
-    print("")
-    print(f"DPNP: current queue is CPU: {current_queue_is_cpu}")
-    print(f"DPNP: version: {dpnp.version.version}")
-    print("")
-    if not current_queue_is_cpu or os.getenv('DPNP_QUEUE_GPU') == '1':
+    if os.getenv('DPNP_QUEUE_GPU') == '1':
         excluded_tests.extend(get_excluded_tests(test_exclude_file_gpu))
+        # TODO: remove below logic when all the tests passed on special GPU device
```
samir-nasibli, line 53: We do not need special cases.
It is just better add some notes on README.md about this device.

samir-nasibli: And create issue ticket for that.

densmirn:
`import dpnp` in pytest config makes most tests fail on special GPU device. So dropped usage of dpnp in the config.

densmirn:
> `import dpnp` in pytest config makes most tests fail on special GPU device. So dropped usage of dpnp in the config.

And it's not clear why tests fail.

samir-nasibli:
@densmirn Do we have CI jobs for testing on that "special GPU device"?
If not, then you shouldn't add changes to the working code.

PR #685, opened 2021-04-23
STY: style fix 4/23/2021

samir-nasibli:


dpnp/backend/kernels/dpnp_krnl_sorting.cpp
```diff
@@ -172,7 +172,7 @@ void dpnp_partition_c(void* array1_in, void* array2_in, void* result1, const siz
 
     event.wait();
 
-    dpnp_memory_free_c(shape);
+        dpnp_memory_free_c(shape);
```
samir-nasibli, line 175: @shssf 
it seems that stylechecker has problems

dpnp/backend/kernels/dpnp_krnl_sorting.cpp
```diff
@@ -121,29 +122,29 @@ void dpnp_partition_c(void* array1_in, void* array2_in, void* result1, const siz
...
+        std::partial_sort(matrix, matrix + shape_[ndim - 1], matrix + shape_[ndim - 1]);
         for (size_t j = ind_begin; j < ind_end + 1; ++j)
         {
             size_t ind = j - ind_begin;
             arr2[j] = matrix[ind];
         }
     }
 
-    size_t* shape = reinterpret_cast<size_t*>(dpnp_memory_alloc_c(ndim * sizeof(size_t)));
+        size_t* shape = reinterpret_cast<size_t*>(dpnp_memory_alloc_c(ndim * sizeof(size_t)));
```
samir-nasibli, line 142: the same comment as above

shssf (approved):

dpnp/backend/kernels/dpnp_krnl_sorting.cpp
shssf: it depends on tools version because diffrent versions has diffrent default parameters. Previously, I tried to fix stylechecker parameters in configuration file but number of differences are grow with time. Looks like we have some new parameter which needs to be fixed.

samir-nasibli:
@shssf 
CI fails due to problems with `Ubuntu 20.04 External Tests`. We need to fix it.
The rest is OK. We can merge it.

PR #688, opened 2021-04-27
rebuild with numpy-devel 1.19.2 for NDA release

xaleryb:


Alexander-Makaryev:
This PR was used to get custom build from CI, it is not actual any more.

PR #698, opened 2021-05-06
add kernel for max with axis

Rubtsowa:


dpnp/dpnp_algo/dpnp_algo_statistics.pyx
```diff
@@ -142,89 +142,62 @@ cpdef dparray dpnp_cov(dparray array1):
...
     cdef Py_ssize_t axis_size = 0
+    cdef dparray_shape_type axis_ = axis
 
-    func(input.get_data(), result.get_data(), < size_t * > input._dparray_shape.data(), input.ndim, < size_t * > axis.data(), axis_size)
+    if _axis_ is not None:
+        axis = _axis_
+        axis_.reserve(len(axis))
+        for shape_it in axis:
+            if shape_it < 0:
+                raise ValueError("DPNP dparray::__init__(): Negative values in 'shape' are not allowed")
```
samir-nasibli, line 163: avoid raising error on cython level. Move this, please, on python level.

dpnp/dpnp_iface_statistics.py
```diff
@@ -301,8 +301,28 @@ def max(input, axis=None, out=None, keepdims=numpy._NoValue, initial=numpy._NoVa
     """
 
     if not use_origin_backend(input):
+        isaxis = True
```
shssf, line 304: What cases you are trying to handle here? Is it a code to validate axis as input?
It would be better to write small comment in the code about it.
If it a user input validation procedure - it is better to put it in some "utils" place. Perhaps it is already where.

Rubtsowa: @shssf 
DPNP dparray::__init__(): Negative values in 'shape' are not allowed

To avoid causing this error at the Cython level, this check has been added. The error call removal was requested in the PR above.

shssf (approved):

PR #700, opened 2021-05-11
Revert "add trace"

densmirn:
Reverts IntelPython/dpnp#651

We can close the PR after providing other one where tests for `trace` are passed on GPU. It seems `trace` doesn't work on GPU due to memory for `shape` is allocated on CPU but used on GPU.


densmirn:
Provided fix https://github.com/IntelPython/dpnp/pull/701. So close the PR.

PR #722, opened 2021-05-24
support parameter out for matmul func

KsanaKozlova:


dpnp/dpnp_algo/dpnp_algo.pyx
```diff
@@ -235,8 +236,19 @@ cpdef dparray dpnp_matmul(dparray in_array1, dparray in_array2):
...
+
+    cdef dparray result
+
+    if out is not None:
+        if out.dtype != result_type:
+            raise TypeError("Inappropriate type of ``out`` variable")
+        elif out.shape != shape_result:
+            raise ValueError("Inappropriate shape of ``out`` variable")
+        else:
+            result = out
```
samir-nasibli, line 248: Do not raise error here.
It is better move this on python level and make this as fallback condition.

dpnp/dpnp_algo/dpnp_algo.pyx
```diff
@@ -235,8 +236,19 @@ cpdef dparray dpnp_matmul(dparray in_array1, dparray in_array2):
...
-    # ceate result array with type given by FPTR data
-    cdef dparray result = dparray(shape_result, dtype=result_type)
+
+    cdef dparray result
+
+    if out is not None:
+        if out.dtype != result_type:
+            raise TypeError("Inappropriate type of ``out`` variable")
+        elif out.shape != shape_result:
+            raise ValueError("Inappropriate shape of ``out`` variable")
```
densmirn, line 246: I would use `checker_throw_value_error` in these cases, e.g.:
`checker_throw_value_error('matmul', 'out.dtype', out.dtype, result_type)`

densmirn: You could replace `elif` with `if` and remove line with `else`.

samir-nasibli: No error raising on cython layer.

densmirn: We don't know `result_type` in python level.

densmirn (approved):

dpnp/dpnp_algo/dpnp_algo.pyx
```diff
@@ -235,8 +236,18 @@ cpdef dparray dpnp_matmul(dparray in_array1, dparray in_array2):
...
-    # ceate result array with type given by FPTR data
-    cdef dparray result = dparray(shape_result, dtype=result_type)
+
+    cdef dparray result
+
+    if out is not None:
+        if out.dtype != result_type:
+            checker_throw_value_error('matmul', 'out.dtype', out.dtype, result_type)
+        if out.shape != shape_result:
+            checker_throw_value_error('matmul', 'out.shape', out.shape, shape_result)
```
samir-nasibli, line 246: This will not work. Avoid using this kind of error throwing. It is better avoiding error raising in cython level. redesign is needed.

samir-nasibli (commented):
Please, avoid using quoted comments, use `#` sign for them.

shssf (approved):

PR #727, opened 2021-06-07
support parameter out for sin function

KsanaKozlova:


dpnp/dpnp_algo/dpnp_algo_trigonometric.pyx
```diff
@@ -142,8 +142,32 @@ cpdef dparray dpnp_radians(dparray x1):
...
+
+    result_type = dpnp_DPNPFuncType_to_dtype(< size_t > kernel_data.return_type)
+
+    shape_result = x1.shape
+
+    cdef dparray result
+
+    if out is not None:
+        if out.dtype != result_type:
+            checker_throw_value_error('sin', 'out.dtype', out.dtype, result_type)
```
samir-nasibli, line 159: Do you have any tests for checking this branch?

It seems that this error raise will not work.

shssf: Are you sure we need to throw an exception if types are diffrent? Does NumPy do the same?

KsanaKozlova: If the out type is smaller than the input type, then NumPy does the same (throws an error), if out type is larger than the input type, then Numpy casts without an error.

>>> import numpy as np
>>> x_float = np.array([1, 2, 3, 4], dtype=np.float32)
>>> out_double = np.empty(4, dtype=np.float64)
>>> out_long = np.empty(4, dtype=np.int64)
>>> np.sin(x_float, out=out_long)
  File "<stdin>", line 1, in <module>
TypeError: ufunc 'sin' output (typecode 'f') could not be coerced to provided output parameter (typecode 'l') according to the casting rule ''same_kind''
>>> np.sin(x_float, out=out_double)
array([ 0.84147102,  0.90929741,  0.14112   , -0.7568025 ])

samir-nasibli: > If the out type is smaller than the input type, then NumPy does the same (throws an error), if out type is larger than the input type, then Numpy casts without an error.

The problem is not in the smaller bit size of the output, `out_long` is bigger than `x_float`.
Issue in the casting with `same kind` rule.

samir-nasibli: In any way, this not a solution.

Your code here will simply not work correctly.
To check, add tests that process this case.
Smth like this: https://github.com/IntelPython/dpnp/blob/master/tests/test_random.py#L29:L31

shssf: @KsanaKozlova I think the problem is a bit deeper than we think.
we have 4 output result types configuration knobs:
- default result type
- defined by `out` parameter
- defined by `dtype` parameter
- defined (determined) by `casting` parameter

This comment was about first two cases. But the problem bigger.
I don't understand (do you?) what does it mean "bigger type". Bigger by number of machine bits representation or something else? 
What is "bigger" type `int32` or `uint32`? Number of bits are the same.

It looks like we need to think (in other PR work flow) about some common procedure which "match" or "collate" types.

densmirn (approved):

PR #738, opened 2021-06-24
CHOLESKY enable on GPU

densmirn:


shssf (approved):
I assume this is an optimization which uses preallocated result array instead intermediate scratch array.

PR #754, opened 2021-06-29
make example compile lines easier

shssf:
Several examples could not be compiled due changed interface functions


samir-nasibli (approved):

shssf:
I assume test are failed because intel One API update.

PR #765, opened 2021-07-05
implement convert_single_elem_array_to_scalar() function

shssf:


shssf:
Have no idea for better name for this function

PR #768, opened 2021-07-07
simple verbose mode impl

Alexander-Makaryev:
This is a simple initial implementation of "verbose mode" for DPNP like it is done with MKL_VERBOSE in MKL.
The goal is to be able to detect calls of MKL/Kernels/etc and get some info like execution time.
This approach with "empty" barrier event can be applicable to any SYCL functionality.
Verbose mode is enabled by environment variable DPNP_VERBOSE=1, behaviour looks like in MKL.
Example of prints with a call of linalg.qr:
```
DPNP_VERBOSE oneapi::mkl::lapack::geqrf Time: 30.7181 ms
DPNP_VERBOSE oneapi::mkl::lapack::orgqr Time: 30.8357 ms
execution time: 0.0724216050002724 (sec.)
```
The last one row is a printing of time measured on Python side in users script.


dpnp/backend/src/queue_sycl.cpp
```diff
@@ -170,7 +170,15 @@ void backend_sycl::backend_sycl_queue_init(QueueOptions selector)
...
-    queue = new cl::sycl::queue(dev, exception_handler);
+    if (is_verbose_mode())
+    {
+        cl::sycl::property_list properties{sycl::property::queue::enable_profiling()};
+        queue = new cl::sycl::queue(dev, exception_handler, properties);
+    }
+    else
+    {
+        queue = new cl::sycl::queue(dev, exception_handler);
+    }
```
shssf, line 181:
```suggestion
    cl::sycl::property_list properties;
    if (is_verbose_mode())
    {
        properties.push_back(sycl::property::queue::enable_profiling());
    }
    queue = new cl::sycl::queue(dev, exception_handler, properties);
```

PR #770, opened 2021-07-08
add allclose function

KsanaKozlova:


dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -71,6 +71,74 @@ void dpnp_all_c(const void* array1_in, void* result1, const size_t size)
...
+
+    const _DataType1* array1 = reinterpret_cast<const _DataType1*>(array1_in);
+    const _DataType2* array2 = reinterpret_cast<const _DataType2*>(array2_in);
+    _ResultType* result = reinterpret_cast<_ResultType*>(result1);
+
+    if (!array1_in || !result1)
+    {
+        return;
+    }
+
```
densmirn, line 90:
```suggestion
    if (!array1_in || !array2_in)
    {
        return;
    }

    const _DataType1* array1 = reinterpret_cast<const _DataType1*>(array1_in);
    const _DataType2* array2 = reinterpret_cast<const _DataType2*>(array2_in);
    _ResultType* result = reinterpret_cast<_ResultType*>(result1);

```

densmirn (commented):
Can we return either `True` or `False` from the native function without using one-element array for result?

dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -71,6 +71,74 @@ void dpnp_all_c(const void* array1_in, void* result1, const size_t size)
     event.wait();
 }
 
+template <typename _DataType1, typename _DataType2, typename _ResultType>
+class dpnp_allclose_c_kernel;
+
+template <typename _DataType1, typename _DataType2, typename _ResultType>
+void dpnp_allclose_c(const void* array1_in, const void* array2_in, void* result1, const size_t size, double rtol, double atol)
```
samir-nasibli, line 78: We already have `rtol` and `atol` in std namespace.
For better naming please rename them.

dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -71,6 +71,74 @@ void dpnp_all_c(const void* array1_in, void* result1, const size_t size)
...
+    auto kernel_parallel_for_func = [=](cl::sycl::id<1> global_id) {
+        size_t i = global_id[0];
+
+        if (std::abs(array1[i] - array2[i]) <= (atol + rtol * std::abs(array2[i]))){
+            close[i] = 1;
+        }
+        else
+        {
+            close[i] = 0;
+        }
```
samir-nasibli, line 111: it is better to use ternary operator here.

dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -71,6 +71,74 @@ void dpnp_all_c(const void* array1_in, void* result1, const size_t size)
...
+    event = DPNP_QUEUE.submit(kernel_func);
+
+    event.wait();
+
+    dpnp_sum_c<int, int>(sumclose, close, &size, 1, NULL, 0, NULL, NULL);
+
+    dpnp_memory_free_c(close);
+
+
+    if (sumclose[0] == (int)size){
```
samir-nasibli, line 128: it is better to use ternary operator here.

samir-nasibli: casting size_t to int be careful

dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -71,6 +71,74 @@ void dpnp_all_c(const void* array1_in, void* result1, const size_t size)
...
+    cl::sycl::range<1> gws(size);
+    auto kernel_parallel_for_func = [=](cl::sycl::id<1> global_id) {
+        size_t i = global_id[0];
+
+        if (std::abs(array1[i] - array2[i]) <= (atol + rtol * std::abs(array2[i]))){
+            close[i] = 1;
+        }
+        else
+        {
+            close[i] = 0;
```
shssf, line 110:
```suggestion
        if (std::abs(array1[i] - array2[i]) > (atol + rtol * std::abs(array2[i])))
        {
            result[0] = false;
        }
```

shssf: no `dpnp_sum_c` call needed in this way and no extra memory allocation required

Alexander-Makaryev:
@KsanaKozlova 
What about more tests from, as an example,
https://github.com/IntelPython/dpnp/blob/master/tests/third_party/cupy/logic_tests/test_comparison.py#L119
It is fine that you implement your own tests, but it is easy to miss something in this case. Third party tests are very helpful.

shssf:
> t?

shssf:
> Can we return either `True` or `False` from the native function without using one-element array for result?

It brake common functions API style in DPNPC

dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -71,6 +71,52 @@ void dpnp_all_c(const void* array1_in, void* result1, const size_t size)
...
+        return;
+    }
+
+    cl::sycl::event event;
+
+    cl::sycl::range<1> gws(size);
+    auto kernel_parallel_for_func = [=](cl::sycl::id<1> global_id) {
+        size_t i = global_id[0];
+
+        if (std::abs(array1[i] - array2[i]) > (atol_val + rtol_val * std::abs(array2[i]))){
```
samir-nasibli, line 102: Just for thought.
`std::abs` works only on host. That means you do memcpy every time between host and GPU dev when you call std::abs. This can have a significant impact on performance.
We have `cl::sycl::fabs` for genfloats, and you can implement your own abs kernel for gen integers.

samir-nasibli (commented):
It would be fine having `cupy` external tests enabled.

dpnp/backend/kernels/dpnp_krnl_logic.cpp
Alexander-Makaryev: Per my understanding, kernel function is compiled and executed on device as one whole. So it is impossible that part of it is executed on device and another part is executed on host. It means that compiler was able to compile `std::abs` in kernel function.
On another hand should be a reason why `cl::sycl::fabs` exists, possibly its implementation is optimized for kernels. It is a good point to compare performance of implementations of kernel with `std::abs` and with `cl::sycl::fabs`.
Maybe we should create a separate issue/task about it.

shssf (approved):

PR #775, opened 2021-07-09
add parameter out to exp, log and cos funcs

KsanaKozlova:


dpnp/dpnp_algo/dpnp_algo_trigonometric.pyx
```diff
@@ -116,8 +162,31 @@ cpdef dparray dpnp_expm1(utils.dpnp_descriptor x1):
     return call_fptr_1in_1out(DPNP_FN_EXPM1, x1, x1.shape)
 
 
-cpdef dparray dpnp_log(utils.dpnp_descriptor x1):
-    return call_fptr_1in_1out(DPNP_FN_LOG, x1, x1.shape)
+cpdef dparray dpnp_log(utils.dpnp_descriptor x1, dparray out=None):
```
shssf, line 165: I think it is better to change `call_fptr_1in_1out` to add one more parameter `out`.

dpnp/dpnp_algo/dpnp_algo_trigonometric.pyx
```diff
@@ -116,8 +162,31 @@ cpdef dparray dpnp_expm1(utils.dpnp_descriptor x1):
...
+    cdef DPNPFuncData kernel_data = get_dpnp_function_ptr(DPNP_FN_LOG, param1_type, param1_type)
+
+    result_type = dpnp_DPNPFuncType_to_dtype(< size_t > kernel_data.return_type)
+
+    shape_result = x1.shape
+
+    cdef dparray result
+
+    if out is not None:
+        if out.dtype != result_type:
```
shssf, line 177: I think it is better to move these checks to python level and go to fallback instead throwing exceptions.

dpnp/dpnp_iface_trigonometric.py
```diff
@@ -395,9 +395,9 @@ def cos(x1):
 
     x1_desc = dpnp.get_dpnp_descriptor(x1)
     if x1_desc:
-        return dpnp_cos(x1_desc)
+        return dpnp_cos(x1_desc, out=out)
```
shssf, line 398:
```suggestion
        return dpnp_cos(x1_desc, out)
```
It looks like we use positional parameters and do not use "named" (keyword) parameters in Cython functions.

dpnp/dpnp_algo/dpnp_algo.pyx
```diff
@@ -258,22 +258,32 @@ cdef dparray call_fptr_1out(DPNPFuncName fptr_name, dparray_shape_type result_sh
...
+    else:
+        result = utils.create_output_array(result_shape, kernel_data.return_type, None)
 
     cdef fptr_1in_1out_t func = <fptr_1in_1out_t > kernel_data.ptr
-    """ Call FPTR function """
+
     func(x1.get_data(), result.get_data(), x1.size)
 
-    return result
+    return resultlt
```
Alexander-Makaryev, line 286: @KsanaKozlova 
What is resultlt? I don't see definition, looks like a typo... Did you tried to run this code?)

Alexander-Makaryev:
@KsanaKozlova 
Just to double check.
I see only self-made tests, we really have no any third-party tests for `out` param in these functions?

shssf (approved):

PR #779, opened 2021-07-12
ENH: kernels for `random.vonmisses`; part 1

samir-nasibli:
## Description
Enable computations on devices [CPU/GPU].

## Tests
* DPNP own:
```shell
tests/test_random.py::TestDistributionsVonmises::test_moments[large_kappa] PASSED
tests/test_random.py::TestDistributionsVonmises::test_moments[small_kappa] PASSED
tests/test_random.py::TestDistributionsVonmises::test_invalid_args PASSED
tests/test_random.py::TestDistributionsVonmises::test_seed[large_kappa] PASSED
tests/test_random.py::TestDistributionsVonmises::test_seed[small_kappa] PASSED
```
* \+ numpy external

### DIFF
--- a/dpnp/backend/kernels/dpnp_krnl_random.cpp
+++ b/dpnp/backend/kernels/dpnp_krnl_random.cpp
@@ -1244,7 +1244,7 @@ void dpnp_rng_uniform_c(void* result, const long low, const long high, const siz
 template <typename _DataType>
 void dpnp_rng_vonmises_large_kappa_c(void* result, const _DataType mu, const _DataType kappa, const size_t size)
 {
-    if (!size)
+    if (!size || !result)
     {
         return;
     }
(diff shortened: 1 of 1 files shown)


samir-nasibli:
@shssf moved some part of changes from #681 here. This PR is ready to merge.

dpnp/backend/kernels/dpnp_krnl_random.cpp
```diff
@@ -1391,20 +1394,22 @@ void dpnp_rng_vonmises_small_kappa_c(void* result, const _DataType mu, const _Da
...
-    {
-        double mod, resi;
-
-        resi = (Vvec[i] < 0.5) ? mu - result1[i] : mu + result1[i];
-        mod = fabs(resi);
-        mod = (fmod(mod + M_PI, 2 * M_PI) - M_PI);
-        result1[i] = (resi < 0) ? -mod : mod;
-    }
+    cl::sycl::range<1> gws(size);
+    auto paral_kernel_acceptance = [&](cl::sycl::handler& cgh) {
```
shssf, line 1400: it looks like this lambda is mostly the same as in other part of code. I suspect, this will be used for many other part of code.
If it is not - please drop off this message.
So, perhaps it is better to make it as a macro or, that is better, as inline function in this C++ source. Just to avoid code duplication.
This is not quite important. It might be leaved as is, at least for now.

shssf (approved):

dpnp/backend/kernels/dpnp_krnl_random.cpp
samir-nasibli: @shssf There is no code duplication here. Please read messages carefully)

samir-nasibli: @shssf Actually you are right. But it is valuable, when we reusing the same code more than 2 times. I vote leave it as is.

shssf (approved):

PR #782, opened 2021-07-13
FIX: `vstack`, `hstack` fallback and tests

samir-nasibli:
## Description
* fixed `vstack` and `hstack` fallback
* numpy like tests for `vstack` and `hstack`

### DIFF
--- a/tests/test_arraymanipulation.py
+++ b/tests/test_arraymanipulation.py
@@ -34,24 +34,24 @@ class TestHstack:
         numpy.testing.assert_array_equal(res, desired)
 
     def test_1D_array(self):
-        a = array([1])
-        b = array([2])
-        res = hstack([a, b])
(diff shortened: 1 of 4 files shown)


tests/test_arraymanipulation.py
```diff
@@ -16,3 +17,78 @@ def test_asfarray(type, input):
...
+
+    def test_0D_array(self):
+        b = dpnp.array(2)
+        a = dpnp.array(1)
+        res = dpnp.hstack([a, b])
+        desired = dpnp.array([1, 2])
+        numpy.testing.assert_array_equal(res, desired)
+
+    def test_1D_array(self):
+        a = array([1])
```
shssf, line 37: Could you please use `dpnp.` prefix as above to make it clear what exactly module used?

shssf (approved):

PR #802, opened 2021-07-30
ENH: `dpnp.histogram` fallback; tests

samir-nasibli:
## Description
* fallback for `dpnp.histogram`
* tests (some numpy original tests, while we have issue with them)

## Tests
* pasta some numpy original tests to `tests/test_histograms.py`, while we have issue with numpy testsing
```bash
tests/test_histograms.py::TestHistogram::test_simple PASSED
tests/test_histograms.py::TestHistogram::test_one_bin PASSED
tests/test_histograms.py::TestHistogram::test_density PASSED
tests/test_histograms.py::TestHistogram::test_arr_weights_mismatch PASSED 
```


dpnp/dpnp_iface_histograms.py
```diff
@@ -0,0 +1,82 @@
...
+import dpnp
+import numpy
+
+
+__all__ = [
+    'histogram',
+]
+
+
+def histogram(a, bins=10, range=None, normed=None, weights=None, density=None):
```
shssf, line 55: Could you please move this code into `dpnp_iface_statistics.py`?

samir-nasibli: Logically and due the the `numpy`'s division by modules `histogram` should not be in `dpnp_iface_statistics.py`.
However, if it's required I will add it there.

shssf (approved):

PR #819, opened 2021-08-05
TEST: `dpnp` own tests for `fft` modul

samir-nasibli:


tests/test_fft.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+
+import dpnp
+
+import numpy
+
+@pytest.mark.parametrize("type",
+                         [numpy.complex128, numpy.complex64, numpy.float32, numpy.float64, numpy.int32, numpy.int64],
```
shssf, line 8:
```suggestion
                         [dpnp.complex128, dpnp.complex64, dpnp.float32, dpnp.float64, dpnp.int32, dpnp.int64],

```

tests/test_fft.py
```diff
@@ -0,0 +1,22 @@
...
+import dpnp
+
+import numpy
+
+@pytest.mark.parametrize("type",
+                         [numpy.complex128, numpy.complex64, numpy.float32, numpy.float64, numpy.int32, numpy.int64],
+                         ids=['complex128', 'complex64', 'float32', 'float64', 'int32', 'int64'])
+def test_fft(type):
+    # 1 dim array
+    data = numpy.arange(100, dtype=type)
```
shssf, line 12: Why you test foreign data type for this FFT test suite? it is not interoperability testing suite.

tests/test_fft.py
```diff
@@ -0,0 +1,22 @@
...
+    # TODO:
+    # doesn't work correct with `complex64` (not supported)
+    # dpnp_data = dpnp.arange(100, dtype=type)
+    dpnp_data = dpnp.array(data)
+
+    np_res = numpy.fft.fft(data)
+    dpnp_res = dpnp.fft.fft(dpnp_data)
+
+    numpy.testing.assert_array_equal(dpnp_res, np_res)
```
shssf, line 21:
```suggestion
    numpy.testing.assert_allclose(dpnp_res, np_res, rtol=1e-4, atol=1e-7)
```
you can not compare floating point numbers as binary after complex floating point operations. You always need to use  rtol/atol

samir-nasibli: Of course you are right. Floating point numbers are not compared this way.
I specified the function incorrectly.  =)
It seemed to me that `assert_array_equal` works like `assert_allclose` with parameters `rtol = 1e-07, atol = 0` for floating point numbers.
Thank you!

tests/test_fft.py
samir-nasibli: In this case this types are foreign for `numpy` in test suit.
It is better use strings for both lib `dtype` constructions.

shssf: This test is in DPNP test suite. I don't understand what does it mean "foreign".

tests/test_fft.py
```diff
@@ -0,0 +1,20 @@
...
+import numpy
+
+@pytest.mark.parametrize("type", ['complex128', 'complex64', 'float32', 'float64', 'int32', 'int64'])
+def test_fft(type):
+    # 1 dim array
+    data = numpy.arange(100, dtype=numpy.dtype(type))
+    # TODO:
+    # doesn't work correct with `complex64` (not supported)
+    # dpnp_data = dpnp.arange(100, dtype=dpnp.dtype(type))
+    dpnp_data = dpnp.array(data)
```
shssf, line 14:
```suggestion
    dpnp_data = dpnp.arange(100, dtype=numpy.dtype(type))

```
it would be better to use same creation procedure for both (expected and test data). To avoid issues in other function which are not related to this test.

samir-nasibli: I have had dilemma for `complex64`, see comment below. But in any way I agree with you, actually `complex64` are disabled now. So I will change.

PR #822, opened 2021-08-05
ENH: example for DPNPC fft

samir-nasibli:
## Description
* time test example for dpnpc fft vs mkl fft.

### DIFF
--- a/dpnp/backend/examples/example11.cpp
+++ b/dpnp/backend/examples/example11.cpp
@@ -1,0 +1,134 @@
+//*****************************************************************************
+// Copyright (c) 2016-2020, Intel Corporation
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


dpnp/backend/examples/example11.cpp
```diff
@@ -0,0 +1,134 @@
...
+#include <iostream>
+#include <time.h>
+
+#include <CL/sycl.hpp>
+#include <oneapi/mkl.hpp>
+
+#include <dpnp_iface.hpp>
+
+template <typename T>
+void print_dpnp_array(T* arr, size_t size)
```
shssf, line 43: It is better to use https://github.com/IntelPython/dpnp/blob/1f0c8804997508d548d833489ac91887c6abbe3d/dpnp/backend/src/dpnp_utils.hpp#L295 instead

shssf:
What is the purpose of this example?

samir-nasibli:
> What is the purpose of this example?

PR in progress.

PR #828, opened 2021-08-06
Fix issue discovered by test tests/test_statistics.py

densmirn:


dpnp/backend/kernels/dpnp_krnl_statistics.cpp
```diff
@@ -653,6 +653,17 @@ template <typename _DataType, typename _ResultType>
 void dpnp_var_c(
     void* array1_in, void* result1, const size_t* shape, size_t ndim, const size_t* axis, size_t naxis, size_t ddof)
 {
+    size_t size = 1;
+    for (size_t i = 0; i < ndim; ++i)
+    {
+        size *= shape[i];
```
samir-nasibli, line 659: undef behavior for shape with zero val:
For example:
```python
>>> dpnp.ones(shape=(12,0,3)).shape
(12, 0, 3)
```

densmirn: PR decides other task and I didn't touch this code, just moved it, but perhaps you are right and it could be fixed in separate PR.

samir-nasibli: Actually, I see that `numpy` has the same behavior for similar cases:
```python
>>> numpy.ones(shape=(12,0,3)).size
0
>>> numpy.ones(shape=(12,0,3)).shape
(12, 0, 3)
>>> numpy.empty(shape=(12,0,3)).shape
(12, 0, 3)
>>> numpy.empty(shape=(12,0,3)).size
0
>>> numpy.var(numpy.ones(shape=(12,0,3)))
nan
>>> dpnp.var(dpnp.ones(shape=(12,0,3)))
nan
```
It could be fixed in separate PR, perhaps no fix is needed.

shssf: > undef behavior for shape with zero val:
> For example:
> 
> ```python
> >>> dpnp.ones(shape=(12,0,3)).shape
> (12, 0, 3)
> ```

If you have shape with zero in ant position - size of the array must be 0.

shssf (approved):

PR #872, opened 2021-08-20
SYCL memory type detector. experimental

shssf:


dpnp/backend/src/queue_sycl.hpp
```diff
@@ -148,4 +148,47 @@ class backend_sycl
...
+              << "\n\t device.is_accelerator=" << queue.get_device().is_accelerator()
+              << std::endl;
+
+    if (queue.is_host() && (ptr_type == cl::sycl::usm::alloc::device))
+    { // move from Device memory to Host
+        // dst_ptr = alloc()
+        queue.memcpy(dst_ptr, src_ptr, 1);
+
+    }
+    else if(!queue.is_host() && ((ptr_type == cl::sycl::usm::alloc::host) || (ptr_type == cl::sycl::usm::alloc::unknown)))
```
samir-nasibli, line 185: Different conditions for the same block of code. 
If there are no typos, conditions can be combined by the logical-OR. But it can be left as is for more readability.

samir-nasibli, line 185: Different conditions for the same block of code.
If there are no typos, conditions can be combined by the logical-OR. But it can be left as is for more readability.

PR #912, opened 2021-08-30
Fix TBB issue

shssf:


oleksandr-pavlyk (approved):
Use of `["-static-libstdc++", "-static-libgcc"]` actually breaks `libstdc++`. Use of these flags results in some _cxa_* symbols being exported by Python native extensions, like dparray. 

When stdc++ was also loaded dynamically one ended one having two statically allocated global structures in the process (_cxa_global structs use to track exceptions).

TBB broke because it raised an exception by calling into one part of stdc++, and was trying to catch the exception by calling into another copy of stdc++ which was not aware of the exception being thrown.

oleksandr-pavlyk:
I confirmed locally that this changes fixes the crash reported in SAT-4417

Alexander-Makaryev (approved):
🥇

utils/command_build_clib.py
```diff
@@ -64,7 +64,7 @@
 _project_linker = "dpcpp"
 _project_cmplr_flag_sycl_devel = ["-fsycl-device-code-split=per_kernel"]
 _project_cmplr_flag_sycl = ["-fsycl"]
-_project_cmplr_flag_stdcpp_static = ["-static-libstdc++", "-static-libgcc"]
+_project_cmplr_flag_stdcpp_static = [] # This brakes TBB ["-static-libstdc++", "-static-libgcc"]
```
alexey-katranov, line 67: The comment is slightly misleading. It does not brake TBB but it brakes C++ runtime behavior because exception are improperly handled due to symbols conflicts. I.e. the exception is created and handled in one instance of C++ runtime but `std::current_exception` (called by TBB in a catch section) is called from another instance C++ runtime. Therefore, the `std::current_exception` returns empty exception (because the second C++ runtime does not know about exception created by the first C++ runtime) that crashes C++ runtime if passed to `std::rethrow_exception`.

PR #915, opened 2021-08-30
use sycl_adapter in krnl_random

shssf:


samir-nasibli:
Will this adapter be used only for specific functions or for whole module?

shssf:
> Will this adapter be used only for specific functions or for whole module?
This adapter needed to copy memory that can not be used with current SYCL queue. So, it will be used for particular functions which requires external arrays handling

PR #921, opened 2021-09-01
Add converting usm_array to ndarray in fallback

densmirn:


shssf (changes requested):
Why we can't implement "to_numpy" as "asnumpy" in doctl?

shssf (approved):

Alexander-Makaryev (approved):
LGTM

PR #923, opened 2021-09-06
Fix test test_asfarray and add test test_asfarray2

densmirn:
We can't pass `usm_ndarray` as parameter to `numpy.asfarray()`.


shssf (approved):

Alexander-Makaryev (approved):
👍

PR #928, opened 2021-09-09
Move choose from cython to backend

LukichevaPolina:
Move choose from cython to backend

### DIFF
--- a/dpnp/backend/kernels/dpnp_krnl_indexing.cpp
+++ b/dpnp/backend/kernels/dpnp_krnl_indexing.cpp
@@ -86,8 +86,13 @@ template <typename _DataType>
 class dpnp_diagonal_c_kernel;
 
 template <typename _DataType>
-void dpnp_diagonal_c(
-    void* array1_in, const size_t input1_size, void* result1, const size_t offset, size_t* shape, size_t* res_shape, const size_t res_ndim)
+void dpnp_diagonal_c(void* array1_in,
+                     const size_t input1_size,
+                     void* result1,
+                     const size_t offset,
+                     size_t* shape,
+                     size_t* res_shape,
+                     const size_t res_ndim)
 {
     const size_t res_size = std::accumulate(res_shape, res_shape + res_ndim, 1, std::multiplies<size_t>());
     if (!(res_size && input1_size))
@@ -237,7 +242,12 @@ void dpnp_fill_diagonal_c(void* array1_in, void* val_in, size_t* shape, const si
 }
 
 template <typename _DataType>
-void dpnp_nonzero_c(const void* in_array1, void* result1, const size_t result_size, const size_t* shape, const size_t ndim, const size_t j)
+void dpnp_nonzero_c(const void* in_array1,
+                    void* result1,
+                    const size_t result_size,
+                    const size_t* shape,
+                    const size_t ndim,
+                    const size_t j)
 {
     if ((in_array1 == nullptr) || (result1 == nullptr))
     {
@@ -255,7 +265,6 @@ void dpnp_nonzero_c(const void* in_array1, void* result1, const size_t result_si
     DPNPC_ptr_adapter<long> result_ptr(result1, result_size, true, true);
     const _DataType* arr = input1_ptr.get_ptr();
     long* result = result_ptr.get_ptr();
-
 
     size_t idx = 0;
     for (size_t i = 0; i < input1_size; ++i)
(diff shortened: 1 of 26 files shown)


dpnp/dpnp_iface_indexing.py
```diff
@@ -81,35 +81,29 @@ def choose(x1, choices, out=None, mode='raise'):
...
-            pass
-        elif not isinstance(choices, list):
+    x1_desc = dpnp.get_dpnp_descriptor(x1)
+    
+    choices_list =[]
+    for choice in choices:
+        choices_list.append(dpnp.get_dpnp_descriptor(choice))
+
+    size = x1_desc.size
+    choices_size = choices_list[0].size
```
shssf, line 91: `dpnp.get_dpnp_descriptor` can return `False` if no conversion possible. In this case, it should go to fallback

dpnp/dpnp_algo/dpnp_algo_indexing.pyx
```diff
@@ -65,12 +66,27 @@ ctypedef void(*custom_indexing_6in_func_ptr_t)(void *, void * , void * , const s
 ctypedef void(*fptr_dpnp_nonzero_t)(const void * , void * , const size_t, const size_t * , const size_t , const size_t)
 
 
-cpdef utils.dpnp_descriptor dpnp_choose(object input, list choices):
-    cdef shape_type_c obj_shape = utils._object_to_tuple(len(input))
-    cdef utils.dpnp_descriptor res_array = utils_py.create_output_descriptor_py(obj_shape, choices[0].dtype, None)
+cpdef utils.dpnp_descriptor dpnp_choose(utils.dpnp_descriptor input, list choices1):
+    cdef shape_type_c input_shape = input.shape  
+
+    cdef vector[void * ] choices
```
shssf, line 72:
```suggestion
    cdef vector[utils.dpnp_descriptor] choices
```

dpnp/backend/kernels/dpnp_krnl_indexing.cpp
```diff
@@ -26,12 +26,46 @@
...
+void dpnp_choose_c(void* result1, void* array1_in, std::vector<void*> choices1, size_t size)
+{
+    if ((array1_in == nullptr) || (result1 == nullptr) || !choices1.size() || !size)
+    {
+        return;
+    }
+    DPNPC_ptr_adapter<_DataType1> input1_ptr(array1_in, size);
+    _DataType1* array_in = input1_ptr.get_ptr();
+
+    DPNPC_ptr_adapter<_DataType2*> choices_ptr(choices1.data(), choices1.size());
```
shssf, line 49: Wrong usage.
In this code, potentially, the container itself could be copied. The main idea is to copy input memory from `device` to the kernel space. So, it looks like a loop needed here over input data stored in the container

dpnp/backend/kernels/dpnp_krnl_indexing.cpp
```diff
@@ -26,12 +26,46 @@
...
 #include <dpnp_iface.hpp>
 #include "dpnp_fptr.hpp"
 #include "dpnpc_memory_adapter.hpp"
 #include "queue_sycl.hpp"
 
+template <typename _DataType1, typename _DataType2>
+class dpnp_choose_c_kernel;
+
+template <typename _DataType1, typename _DataType2>
+void dpnp_choose_c(void* result1, void* array1_in, std::vector<void*> choices1, size_t size)
```
shssf, line 40: please do not use STL in the API. Change it as `void* choses, size_t choices_size`

shssf:
Thank you. Many things are done right. Few of them needs to be addressed.

dpnp/dpnp_algo/dpnp_algo_indexing.pyx
```diff
@@ -67,7 +67,9 @@ ctypedef void(*fptr_dpnp_nonzero_t)(const void * , void * , const size_t, const
 
 
 cpdef utils.dpnp_descriptor dpnp_choose(utils.dpnp_descriptor input, list choices1):
-    cdef shape_type_c input_shape = input.shape  
+    cdef shape_type_c input_shape = input.shape 
+    cdef shape_type_c choices_shape = (len(choices1),) 
+    cdef shape_type_c choice_shape = choices1[0].shape
```
densmirn, line 72: Looks like we don't need the `shapes`, because we use only first element from them.

densmirn (commented):
This PR shouldn't have code that doesn't related to the function `choose`.

shssf (approved):

PR #941, opened 2021-09-17
Copy inputs if they have strides

Alexander-Makaryev:
We need a workaround for inputs with strides because current implementations of functions in dpnp assume that data is c-contiguous without strides.

### DIFF
--- a/dpnp/dpnp_iface.py
+++ b/dpnp/dpnp_iface.py
@@ -150,7 +150,7 @@ def convert_single_elem_array_to_scalar(obj, keepdims=False):
     return obj
 
 
-def get_dpnp_descriptor(ext_obj):
+def get_dpnp_descriptor(ext_obj, copy_when_strides=True):
     """
     Return True:
       never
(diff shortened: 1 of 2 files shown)


densmirn (approved):

dpnp/dpnp_iface.py
```diff
@@ -169,6 +169,14 @@ def get_dpnp_descriptor(ext_obj):
     if use_origin_backend():
         return False
 
+    # while dpnp functions have no implementation with strides support
+    # we need to create a non-strided copy
+    # if function get implementation for strides case
+    # then this behavior can be disabled with setting "copy_when_strides"
+    if copy_when_strides and hasattr(ext_obj, "strides") and ext_obj.strides is not None:
```
oleksandr-pavlyk, line 176:
```suggestion
    if copy_when_strides and getattr(ext_obj, "strides", None) is not None:
```

oleksandr-pavlyk (approved):

PR #945, opened 2021-09-20
Fix "ufunc 'isnan' not supported for the input types..."

LukichevaPolina:
Fix "ufunc 'isnan' not supported for the input types..."

### DIFF
--- a/dpnp/dpnp_iface_counting.py
+++ b/dpnp/dpnp_iface_counting.py
@@ -75,9 +75,6 @@ def count_nonzero(x1, axis=None, *, keepdims=False):
     5
 
     """
-    if config.__DPNP_OUTPUT_DPCTL__:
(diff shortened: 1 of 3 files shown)


densmirn (approved):

dpnp/dpnp_iface_counting.py
```diff
@@ -72,6 +75,8 @@ def count_nonzero(x1, axis=None, *, keepdims=False):
     5
 
     """
+    if config.__DPNP_OUTPUT_DPCTL__:
+        return call_origin(numpy.count_nonzero, x1, axis, keepdims=keepdims)
```
Alexander-Makaryev, line 79: We made changes in count_nonzero func implementation. Is this fallback still required?

densmirn: I don't think so.

PR #950, opened 2021-09-21
BUG: disabled `non_zero` func tests

samir-nasibli:
## Description
Disabled tests that uses `dpnp_count_non_zero` due the bug in this func.

Fix is on #893


Alexander-Makaryev:
@samir-nasibli Could you please double check it with master? https://github.com/IntelPython/dpnp/pull/893 is merged.

Alexander-Makaryev:
https://github.com/IntelPython/dpnp/pull/893 fixed these tests, there are no such fails on master branch

PR #954, opened 2021-09-22
Skip Results are not equal

LukichevaPolina:


densmirn (approved):

LukichevaPolina:
Tests fail with error "Results are not equal" because None as the result of work dpnp function is compared to the result of numpy function in numpy_cupy_equal. None returns from _call_func (in helper.py), which handles the exception from the dpnp function. The exception happened because usmarray hasn`t method item().

Alexander-Makaryev:

PR #958, opened 2021-09-22
Skip "is not an instance of <class 'dpnp.dparray.dparray'>"

LukichevaPolina:


densmirn (approved):

LukichevaPolina:
The tests are skipped because self.assertIsInstance(r, xp.ndarray)is called in test_diag_construction, which  that compares usmarray and dpnp.ndarray and we cannot change cupy tests.

Alexander-Makaryev:

PR #959, opened 2021-09-22
Skip "is out of bounds for array of dimension"

LukichevaPolina:


samir-nasibli (commented):
fix is not provided.

LukichevaPolina:
The tests are skipped because function get_array_module in test_sort.py which compares usmarray and dpnp.ndarray works incorrectly and we cannot change cupy tests.

Alexander-Makaryev:

PR #961, opened 2021-09-22
Implement dpnp.astype

densmirn:


tests/test_random.py
```diff
@@ -965,21 +965,21 @@ def test_no_miss_numbers(self, dtype):
...
                                       lambda x: dpnp.asarray([(i, i) for i in x],
                                                              [("a", object), ("b", dpnp.int32)])],
                              ids=['lambda x: dpnp.array([])',
-                                  'lambda x: dpnp.asarray(x).astype(dpnp.int8)',
-                                  'lambda x: dpnp.asarray(x).astype(dpnp.float32)',
+                                  'lambda x: dpnp.astype(asarray(x), dpnp.int8)',
+                                  'lambda x: dpnp.astype(asarray(x), dpnp.float32)',
                                   # 'lambda x: dpnp.asarray(x).astype(dpnp.complex64)',
-                                  'lambda x: dpnp.asarray(x).astype(object)',
+                                  # 'lambda x: dpnp.astype(asarray(x), object)',
```
samir-nasibli, line 982: It's better to add the test cases into skip list for disabled tests instead of adding it to the comments block in the test src.

Alexander-Makaryev: Looks like usm array does not support 'object' type, that it why this line is commented...

Alexander-Makaryev (approved):
🥇

PR #966, opened 2021-09-23
fix diff function

Alexander-Makaryev:
Replaced `.item(i+1)` with use of slice and multi-indexes to iterative over elements.


oleksandr-pavlyk:
The change is fine. The code before and code after are equivalent in logic, but that logic may not be correct to begin with.

oleksandr-pavlyk (approved):
Approving, with reservations.

PR #982, opened 2021-09-29
Use Python 3.9 [public CI]

PokhodenkoSA:
- [x] Update GitHub CI for building development package
- [x] Update conda recipe
- [x] Dependencies
  - [x] https://github.com/IntelPython/dpctl/pull/599
- [x] Install oneAPI 2021.3 via APT
 
Closes #984.

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -8,6 +8,11 @@ env:
 jobs:
   build:
     runs-on: ubuntu-20.04
+
+    strategy:
+      matrix:
+        python: [3.8, 3.9]
+
     steps:
       - name: Install Intel repository
         run: |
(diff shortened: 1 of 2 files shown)


Alexander-Makaryev (approved):
LGTM

PokhodenkoSA:
oneAPI 2021.4 installed from apt packages, but also oneAPI 2021.3 packages installed in conda build environment. I think there is conflict. I will try to remove oneAPI from system and use only conda packages.

Alexander-Makaryev:
> oneAPI 2021.4 installed from apt packages, but also oneAPI 2021.3 packages installed in conda build environment. I think there is conflict. I will try to remove oneAPI from system and use only conda packages.

The root cause is in new version of oneDPL/PSTL (one of components in oneAPI), it has incompatible changes.
This component is not present yet as a Python package, that is why using only packages is impossible.

Alexander-Makaryev:
Downgrading of oneDPL/PSTL to previous version (made by downgrading of oneAPI) is a temporary way of dealing with a problem, it should be fixed in future. Created issue https://github.com/IntelPython/dpnp/issues/986

PR #987, opened 2021-10-03
Build of backend library with custom math lib and dpl

Alexander-Makaryev:
Modifications in CI configuration need to be applied before these changes will take effect.


Alexander-Makaryev:
@xaleryb @Vyacheslav-Smirnov @ekomarova I created a ticket number SAT-4526 and need your help)

PR #989, opened 2021-10-04
Changing kernel names to be compatible with latest dpl/pstl

Alexander-Makaryev:
As DPL/PSTL developers say, the correct way is to provide such unique classes for naming in each kernel. They can be removed at all, but in this case name for kernel will be generated automatically and it will be very long and impossible to read.
I tried to do so and debug become a hell.)

### DIFF
--- a/dpnp/backend/kernels/dpnp_krnl_statistics.cpp
+++ b/dpnp/backend/kernels/dpnp_krnl_statistics.cpp
@@ -72,7 +72,10 @@ void dpnp_correlate_c(void* result_out,
 }
 
 template <typename _DataType>
-class dpnp_cov_c_kernel;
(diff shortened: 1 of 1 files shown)


densmirn (commented):
We could use word `dpl` in one kernel class name `dpnp_cov_c_dpl_kernel`.

Alexander-Makaryev:
> We could use word `dpl` in one kernel class name `dpnp_cov_c_dpl_kernel`.

I think that adding number in classs name is more clear and common way. It is not only DPL or not DPL functions but about kernel naming in summary.
What wee will do if we have many DPL functions and many own kernels in one backend function?

densmirn:
np, let it be

densmirn (approved):

PR #991, opened 2021-10-04
Implement strides for 7 element-wise functions part 2

densmirn:
Waiting for https://github.com/IntelPython/dpnp/pull/985 before merge.

PR #985, opened 2021-09-30
Implement strides for 11 element-wise functions

densmirn:

PR #992, opened 2021-10-05
Implement strides for 5 bitwise functions part 3

densmirn:
Waiting for both https://github.com/IntelPython/dpnp/pull/985 and https://github.com/IntelPython/dpnp/pull/991 before merge.


Alexander-Makaryev (approved):
LGTM

PR #993, opened 2021-10-05
dpnp array container, init impl

Alexander-Makaryev:
This PR is created in purpose to share approach and start a discussion, it is not ready for merge.

### DIFF
--- a/dpnp/backend/kernels/dpnp_krnl_common.cpp
+++ b/dpnp/backend/kernels/dpnp_krnl_common.cpp
@@ -84,7 +84,7 @@ cl::sycl::event dot(cl::sycl::queue& queue,
                     size_t input1_strides,
                     size_t input2_strides,
                     size_t size,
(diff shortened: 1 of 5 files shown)


Alexander-Makaryev:
@oleksandr-pavlyk @densmirn Could you please look at `dpnp/dpnp_array.py`? What do you think about it? Is it what we want to have?

dpnp/dpnp_array.py
```diff
@@ -0,0 +1,477 @@
...
+        >>> a = np.arange(6).reshape(2,3)
+        >>> a.argmax()
+        5
+        >>> a.argmax(0)
+        array([1, 1, 1])
+        >>> a.argmax(1)
+        array([2, 2])
+
+        """
+        return argmax(self, axis, out)
```
densmirn, line 250:
```suggestion
        return dpnp.argmax(self, axis, out)
```

densmirn (approved):

PR #1004, opened 2021-10-14
Fix numpy-dpnp import order

densmirn:


oleksandr-pavlyk:
Another example of a well-documented PR. Given the age I surmise this PR has gone stale. Closing

PR #953, opened 2021-09-21
Skip tests "object has no attribute'"

KsanaKozlova:

PR #1037, opened 2021-10-28
FIX: random shuffle for ndim arrays

samir-nasibli:
## Description
* fixed setting `high_dim_size` variable on cython level

## Root cause
Setting `high_dim_size` variable value was accidentally changed on #953, see https://github.com/IntelPython/dpnp/pull/953/files#diff-848ee382ceb656b32b8b6a920adc5a66bf65c04af2789d34beca1af7209cf898R839

`len(multidim_array)` shouldn't be equal to its whole size `multidim_array.size()` for ndim array. so #953 bad fix for this.

## TODO
update/reimplement tests for shuffle


samir-nasibli:
@Alexander-Makaryev will this pr be merged?

antonwolfy:
Closed since no plans to implement that.

PR #1041, opened 2021-11-05
Relax build dependency to numpy

PokhodenkoSA:
The workflow uses following channels for build:
https://github.com/IntelPython/dpnp/blob/457a5984712d28e73531fd70e482c099fe87d2c8/.github/workflows/conda-package.yml#L48

At the same time recipe requires concrete numpy version:
https://github.com/IntelPython/dpnp/blob/457a5984712d28e73531fd70e482c099fe87d2c8/conda-recipe/meta.yaml#L11 

I have relaxed requirements for numpy version in conda recipe because `defaults` channel does not have `numpy 1.19.5` (only `1.19.2`).

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -8,7 +8,7 @@ requirements:
     host:
       - python
       - setuptools
-      - numpy 1.19.5
+      - numpy 1.19
       - cython
       - cmake 3.19
       - dpctl >=0.10


PokhodenkoSA:
@xaleryb @oleksandr-pavlyk 
This PR is not for removing numpy pinning at all. It is separate topic. This PR only for enabling public CI and do not break internal CI at the same time.

oleksandr-pavlyk:
It is hard to imagine that precise pinning to 1.19.5 is warranted. Major/minor granularity is more appropriate if needed at all. 

Ideally version pinning should be done in `conda_build_config.yaml`.

oleksandr-pavlyk (approved):

xaleryb:
Previous variant was realized due to special request from TF to be compatible with 1.19.5 numpy because TF used this version only. I think 1.19 will work for us as well because we have only one version of numpy 1.19 on channel (is numpy 1.19.5).

xaleryb:
@Alexander-Makaryev as component owner please decide if we need to merge it.

Alexander-Makaryev (approved):
LGTM

PR #1052, opened 2021-11-25
prep for dpctl/include reorg (IntelPython/dpctl#685)

oleksandr-pavlyk:
Use `__has_include` to minimize disruption that the said reorg may cause.


densmirn (approved):

Alexander-Makaryev (approved):
LGTM

PR #1054, opened 2021-11-29
Implement SYCL kernels in noncentral_chisquare

LukichevaPolina:


dpnp/backend/kernels/dpnp_krnl_random.cpp
```diff
@@ -562,14 +566,24 @@ void dpnp_rng_noncentral_chisquare_c(void* result, const _DataType df, const _Da
         event_out.wait();
 
         shape = 0.5 * df;
-
+
```
densmirn, line 569:
```suggestion

```

densmirn (approved):

dpnp/backend/kernels/dpnp_krnl_random.cpp
```diff
@@ -594,13 +607,20 @@ void dpnp_rng_noncentral_chisquare_c(void* result, const _DataType df, const _Da
...
+                auto kernel_parallel_for_func2 = [=](cl::sycl::id<1> global_id) {
+                    size_t index = global_id[0];
+                    result1[idx[index + i]] = tmp[index];
+                };
+                auto kernel_func2 = [&](cl::sycl::handler& cgh) {
+                    cgh.parallel_for<class dpnp_rng_noncentral_chisquare_c_kernel2<_DataType>>(gws2, 
+                                                                                               kernel_parallel_for_func2);
+                };
+                event_out = DPNP_QUEUE.submit(kernel_func2);
+                event_out.wait();
```
samir-nasibli, line 620: How many kernels are you submitting here?

samir-nasibli (commented):
Is this an optimization? If so, how is this proven? Is there any measurement data?

LukichevaPolina:
> Is this an optimization? If so, how is this proven? Is there any measurement data?

The purpose of these changes is not optimization, but to implement the execution on the device to prevent copying data to the host.

samir-nasibli:
> > Is this an optimization? If so, how is this proven? Is there any measurement data?
> 
> The purpose of these changes is not optimization, but to implement the execution on the device to prevent copying data to the host.

Is it really copying data to the host? how was it found?

Even if this is not an optimization, and if we assume that copying occurs here, do these changes really do not degrade the execution time?

Alexander-Makaryev:
> > > Is this an optimization? If so, how is this proven? Is there any measurement data?
> > 
> > 
> > The purpose of these changes is not optimization, but to implement the execution on the device to prevent copying data to the host.
> 
> Is it really copying data to the host? how was it found?
> 
> Even if this is not an optimization, and if we assume that copying occurs here, do these changes really do not degrade the execution time?

`DPNPC_ptr_adapter<_DataType> result1_ptr(result, size, true, true);` copy happens here because we explicitly provide `true` value for the third parameter.
We have to do the copy while we have a 'host' code that works with the pointer.

samir-nasibli:
> > > > Is this an optimization? If so, how is this proven? Is there any measurement data?
> > > 
> > > 
> > > The purpose of these changes is not optimization, but to implement the execution on the device to prevent copying data to the host.
> > 
> > 
> > Is it really copying data to the host? how was it found?
> > Even if this is not an optimization, and if we assume that copying occurs here, do these changes really do not degrade the execution time?
> 
> `DPNPC_ptr_adapter<_DataType> result1_ptr(result, size, true, true);` copy happens here because we explicitly provide `true` value for the third parameter. We have to do the copy while we have a 'host' code that works with the pointer.

Pay attention please, here a lot of kernels are added to the queue.
Is there any confirmation/proven data that these changes, even taking into account the copies, do not degrade the execution time?

Alexander-Makaryev (approved):
👍

PR #1057, opened 2021-12-06
build with setuptools

Alexander-Makaryev:


utils/command_build_clib.py
```diff
@@ -270,8 +279,10 @@ def build_libraries(self, libraries):
                     link_command += " /link"  # start linker options
                     link_command += " " + " ".join(extra_link_preargs)
                     link_command += " " + ".lib ".join(libraries) + ".lib"  # libraries
+                    link_command += " /LIBPATH:" + " /LIBPATH:".join(library_dirs)
                     link_command += " /OUT:" + c_library_filename  # output file name
                     link_command += " " + " ".join(extra_link_postargs)
+                    print(link_command)
```
densmirn, line 285: We don't have to keep prints in the code.

Alexander-Makaryev: This print helps to understand what happens in the log because command itself is not printed during execution with `os.system`.

densmirn (approved):

PR #1063, opened 2021-12-11
Pass queue from cython to backend for matmul

densmirn:
Introduce a backend function for matrix multiplication which accepts the execution queue as an argument.


densmirn:
@oleksandr-pavlyk could you please review PR?

Alexander-Makaryev (approved):
lgtm

oleksandr-pavlyk (approved):

Alexander-Makaryev:
We already discussed that doc strings for old interface should be removed.
The test `test_2in_1out` is added but not working right now because parametrization list `func` is empty.
Expected that it will be fixed in next PR.

PR #1083, opened 2021-12-14
Pin mkl version with hash

PokhodenkoSA:
Based on #1082.
This PR adds:
- [x] Package have hash (i.e. `dpnp-0.9.0rc1-py39h4188336_38.tar.bz2`) as variable from conda_build_config.yml used in recipe. The same variable name used in internal CI. It is possible to built the same dpnp version but different mkl version.

Fixes #1079.

PR #1084, opened 2021-12-15
Add `dpl-include` and remove oneAPI installation

PokhodenkoSA:
- [x] based on #1083 
- [x] Add `dpl-include` in build dependencies
- [x] Remove oneAPI installation

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -25,8 +25,6 @@ jobs:
         run: conda install conda-build
       - name: Build conda package
         run: |
-          . /opt/intel/oneapi/setvars.sh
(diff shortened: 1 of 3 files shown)


Alexander-Makaryev:
Changes related to MKL are equivalent to #1090
Also compiler is pinned in #1091
Name of artifact / package from build of last one PR is dpnp-0.9.0rc1-py37h840d394_48.tar.bz2, "hash" is included.

We already have a PR with attempt to use DPL package https://github.com/IntelPython/dpnp/pull/1035 that was created 2 months ago. We are tracking it.
Unfortunately, dpl-include package is not stable at this moment. For example, latest available version in CI is broken and CI fails. That is why we are not ready to use it.

PR #1090, opened 2021-12-23
Pin math lib version to x.x in runtime

Alexander-Makaryev:


PokhodenkoSA (approved):

xaleryb:
I have checked result of this PR ("mkl-dpcpp >=2022.0,<2022.1.0a0",) looks good but I think restriction for min pin can be useless in this case.

PR #1108, opened 2022-01-11
Fix issue where passing dpnp_array to dpctl API as parameter

densmirn:


KsanaKozlova (approved):

Alexander-Makaryev:
Looks like the only one place where new function "unwrap_array" used is call it for parameter of function "copy_from_origin". So, why we need a new function, maybe we should add this logic into "copy_from_origin" func?

densmirn:
> Looks like the only one place where new function "unwrap_array" used is call it for parameter of function "copy_from_origin". So, why we need a new function, maybe we should add this logic into "copy_from_origin" func?

At least we have one more code place where we could use function `unwrap_array`: https://github.com/IntelPython/dpnp/blob/master/dpnp/dpnp_container.py#L74-L77
Perhaps in future we could use the function in other places.
I thought about direct changing of function `copy_from_origin`, but decided that the function shouldn't handle inputs depending on its type and created separate function for it.
We could call `unwrap_array` in `copy_from_origin` directly. @Alexander-Makaryev what do you think about it?

Alexander-Makaryev:
> > Looks like the only one place where new function "unwrap_array" used is call it for parameter of function "copy_from_origin". So, why we need a new function, maybe we should add this logic into "copy_from_origin" func?
> 
> At least we have one more code place where we could use function `unwrap_array`: https://github.com/IntelPython/dpnp/blob/master/dpnp/dpnp_container.py#L74-L77 Perhaps in future we could use the function in other places. I thought about direct changing of function `copy_from_origin`, but decided that the function shouldn't handle inputs depending on its type and created separate function for it. We could call `unwrap_array` in `copy_from_origin` directly. @Alexander-Makaryev what do you think about it?

Sounds good to me

Alexander-Makaryev (approved):
👍

PR #1114, opened 2022-01-21
Add property flags to dpnp_array

densmirn:


PokhodenkoSA (approved):
Possibly, adding a test would be great too.

Alexander-Makaryev (changes requested):
Tests should check the correctness/expected results of the functionality.
But these tests check something else... maybe they check that we can create 2 objects with the equal property, is not it?)
The example below demonstrates this.
As you can see, the current implementation does not provide the expected functionality, but tests pass.
```
In [1]: import dpctl.tensor as dpt

In [2]: shape = (5, 2)

In [3]: X = dpt.usm_ndarray(shape)

In [4]: X.flags
Out[4]: 1

In [5]: type(X.flags)
Out[5]: int

In [6]: import numpy

In [7]: Y = numpy.ndarray((2,5))

In [8]: Y.flags
Out[8]:
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False

In [9]: type(Y.flags)
Out[9]: numpy.flagsobj
```

densmirn:
> Tests should check the correctness/expected results of the functionality. But these tests check something else... maybe they check that we can create 2 objects with the equal property, is not it?) The example below demonstrates this. As you can see, the current implementation does not provide the expected functionality, but tests pass.
> 
> ```
> In [1]: import dpctl.tensor as dpt
> 
> In [2]: shape = (5, 2)
> 
> In [3]: X = dpt.usm_ndarray(shape)
> 
> In [4]: X.flags
> Out[4]: 1
> 
> In [5]: type(X.flags)
> Out[5]: int
> 
> In [6]: import numpy
> 
> In [7]: Y = numpy.ndarray((2,5))
> 
> In [8]: Y.flags
> Out[8]:
>   C_CONTIGUOUS : True
>   F_CONTIGUOUS : False
>   OWNDATA : True
>   WRITEABLE : True
>   ALIGNED : True
>   WRITEBACKIFCOPY : False
>   UPDATEIFCOPY : False
> 
> In [9]: type(Y.flags)
> Out[9]: numpy.flagsobj
> ```

dpnp is based on dpctl in case of property flags where flags differ numpy. Tests check dpnp flags are equal to dpctl flags.

densmirn:
In case if we dpnp should work like numpy in case of flags, I vote for closing the PR.

antonwolfy:
Done in #1265

PR #1115, opened 2022-01-25
Fix test_density for histogram

LukichevaPolina:


densmirn (approved):

Alexander-Makaryev (changes requested):
Looks like dpnp.sum returns array instead scalar (as in original Numpy) and this is a reason why we have to change assert functions. Could you please check it and fix if it is so instead of using another functions?

antonwolfy:
#1785 covers that PR

PR #1120, opened 2022-02-03
support real input in fft with mkl

KsanaKozlova:


densmirn (approved):

dpnp/backend/kernels/dpnp_krnl_fft.cpp
```diff
@@ -167,35 +167,39 @@ void dpnp_fft_fft_sycl_c(const void* array1_in,
...
-                                    const size_t result_size,
-                                    _Descriptor_type& desc,
-                                    const size_t norm)
+void dpnp_fft_fft_mathlib_cmplx_to_cmplx_c(const void* array1_in,
+                                          void* result_out,
+                                          const shape_elem_type* input_shape,
+                                          const size_t shape_size,
+                                          const size_t result_size,
+                                          _Descriptor_type& desc,
+                                          const size_t norm)
```
samir-nasibli, line 176: indentation.

densmirn (commented):
Ideally when C-interfaces in DPNP repeats relevant numpy interfaces. So I'm not sure it's a good idea to extend C-functions with extra parameters. It seems parameter `real` complicates logic of fft.

dpnp/backend/kernels/dpnp_krnl_fft.cpp
```diff
@@ -215,98 +221,184 @@ void dpnp_fft_fft_mathlib_compute_c(DPCTLSyclQueueRef q_ref,
...
-    else if (std::is_same<_DataType_input, std::complex<float>>::value &&
-             std::is_same<_DataType_output, std::complex<float>>::value)
-    {
-        desc_sp_cmplx_t desc(input_shape[shape_size - 1]);
 
-        dpnp_fft_fft_mathlib_compute_c<_DataType_input, _DataType_output, desc_sp_cmplx_t>(
-            q_ref, array1_in, result1, input_shape, shape_size, result_size, desc, norm);
+    sycl::event::wait(fft_events);
+
+    if (!real) {
```
densmirn, line 275: Could we return from function if real is 1 and add comment why? Then shift else-code to the left.

PR #1122, opened 2022-02-08
Extend backend functions interfaces with queue and events

densmirn:


Alexander-Makaryev:
For some reason tests on Windows become extremely slow on this PR. We need to pay attention on this during next iterations of testing.

Alexander-Makaryev (approved):
LGTM

PR #1124, opened 2022-02-09
Skipping tests with fails on Windows after switching shapes/strides d…

Alexander-Makaryev:
…atatype


densmirn:
Wouldn't we like to create separate tbl-file with list of tests failed only on Windows like addition to existing tbl-file? Content from both tbl-files would be merged in runtime on Windows.

Alexander-Makaryev:
> Wouldn't we like to create separate tbl-file with list of tests failed only on Windows like addition to existing tbl-file? Content from both tbl-files would be merged in runtime on Windows.

I think no. This is not a regular behavior that something works on one OS and fails on another.
Right now we have to skip these tests to track CI. The next step is to fix these issues and work with common functionality on all OS.

PR #1127, opened 2022-02-11
fix python layer in fft module

KsanaKozlova:


densmirn (commented):
Let's create variable with value `0` to see what it means. E.g. `norm`.

dpnp/fft/dpnp_iface_fft.py
```diff
@@ -86,10 +86,12 @@ def fft(x1, n=None, axis=-1, norm=None):
...
     if x1_desc:
-        # if norm is None or norm is 'backward':
-        #     norm_val = 0
-        # else:
-        #     norm_val = 1
+        norm_val = 0
+        if norm is "forward":
+            norm_val = 1
+        elif norm is "ortho":
+            norm_val = 2
```
densmirn, line 93: I would propose to create `enum` for `norm`. What do you think about that?

dpnp/fft/dpnp_iface_fft.py
```diff
@@ -69,6 +70,22 @@
 ]
 
 
+class Norm(Enum):
+    backward = 0
+    forward = 1
+    ortho = 2
+
+def _norm_value(norm):
```
densmirn, line 78: Maybe rename it to `get_validated_norm`?

dpnp/fft/dpnp_iface_fft.py
```diff
@@ -69,6 +70,22 @@
...
+    if norm is None or norm == "backward":
```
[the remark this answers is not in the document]
densmirn: My bad, let's keep the check.

densmirn (approved):

Alexander-Makaryev (approved):
LGTM

PR #1143, opened 2022-03-14
to_device array method impl

Alexander-Makaryev:


tests/test_sycl_queue.py
```diff
@@ -83,3 +83,18 @@ def test_2in_1out(func, device):
...
+                         valid_devices,
+                         ids=[device.filter_string for device in valid_devices])
+@pytest.mark.parametrize("device_to",
+                         valid_devices,
+                         ids=[device.filter_string for device in valid_devices])
+def test_to_device(device_from, device_to):
+    data = [1., 1., 1., 1., 1.]
+
+    x = dpnp.array(data, device=device_from)
+    y =  x.to_device(device_to)
```
densmirn, line 98:
```suggestion
    y = x.to_device(device_to)
```

densmirn (approved):

dpnp/dpnp_array.py
```diff
@@ -97,6 +97,13 @@ def T(self):
         else:
             return dpnp.transpose(self)
 
+    def to_device(self, target_device):
+        """
+        Transfer array to target device
+        """
+
+        return dpnp_array(shape=self.shape, buffer=self.get_array().to_device(target_device))
```
densmirn, line 105: Other option is usage of method `__new__` to create `dpnp_array`. As an example there is method `flatten`.

oleksandr-pavlyk (approved):
Looks good.

PR #1148, opened 2022-03-24
Function "linalg.qr" raised an exception with mode='reduced'

antonwolfy:
DPNP function "linalg.qr" raised an exception when running in mode='reduced'.
DPNP backend kernel for "linalg.qr" crashed with a core dump for m*n matrix when m>n.

### DIFF
--- a/dpnp/backend/kernels/dpnp_krnl_linalg.cpp
+++ b/dpnp/backend/kernels/dpnp_krnl_linalg.cpp
@@ -662,17 +662,17 @@ DPCTLSyclEventRef dpnp_qr_c(DPCTLSyclQueueRef q_ref,
         }
     }
 
-    DPNPC_ptr_adapter<_ComputeDT> result1_ptr(q_ref, result1, size_m * size_m, true, true);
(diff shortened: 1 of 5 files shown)


densmirn (approved):

Alexander-Makaryev:
We say that `reduced` mode is supported, but current tests use `complete` mode:
https://github.com/IntelPython/dpnp/blob/224a683e0cc7c0d2e0d75276adaffd429a527530/tests/test_linalg.py#L226
Could you please modify tests to cover `reduced` mode?
(BTW looks like in current tests `complete` mode works too.)

antonwolfy:
> We say that `reduced` mode is supported, but current tests use `complete` mode:
> 
> https://github.com/IntelPython/dpnp/blob/224a683e0cc7c0d2e0d75276adaffd429a527530/tests/test_linalg.py#L226
> 
> 
> Could you please modify tests to cover `reduced` mode?
> (BTW looks like in current tests `complete` mode works too.)

New patch uploaded to cover that. Additionally spontaneous crash was fixed observed on GPU linalg.qr tests.

Alexander-Makaryev (approved):
LGTM

PR #1153, opened 2022-04-01
real and imag funcs impl

Alexander-Makaryev:


dpnp/dpnp_iface_mathematical.py
```diff
@@ -908,6 +910,17 @@ def gradient(x1, *varargs, **kwargs):
...
 
+def imag(val):
+    """
+    Return the imaginary part of the complex argument.
+
+    For full documentation refer to :obj:`numpy.imag`.
+
+    """
+
+    return dpnp.ndarray(val.get_array().shape, buffer=val.get_array().imag)
```
densmirn, line 921: Better to wrap `val` by `dpnp_descriptor`, then call method `get_array()` of `dpnp_descriptor` object. The method will check type of input object and be able to raise understandable exception.

Alexander-Makaryev: done

dpnp/dpnp_array.py
```diff
@@ -585,7 +585,19 @@ def flatten(self, order='C'):
...
+        """
+        Return the imaginary part of the complex argument.
+
+        """
+
+        return dpnp.imag(self)
+
+    @imag.setter
+    def imag(self, value):
+        dpnp.imag(self)[:] = value
```
oleksandr-pavlyk, line 600: This setter only makes sense when `dpnp.imag` is a view into original array's allocation. This is clearly not true for arrays with non-complex types.

PR #1154, opened 2022-04-01
remove mixed host\dev impl from dpnp_all dpnp_any

vladislav-markin:


dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -51,37 +51,34 @@ DPCTLSyclEventRef dpnp_all_c(DPCTLSyclQueueRef q_ref,
...
 
     if (!size)
     {
+        init_mem_event.wait();
         return event_ref;
     }
 
     sycl::range<1> gws(size);
-    auto kernel_parallel_for_func = [=](sycl::id<1> global_id) {
+    auto parallel_for_kernel_func = [=](sycl::id<1> global_id) {
```
Alexander-Makaryev, line 69: What is purpose of renaming `kernel_parallel_for_func` -> `parallel_for_kernel_func`?
It will be better to leave old variable name because the same name is used in other functions implementations. It makes code more unified.

vladislav-markin: I though it would be more convenient name, like "parallel_for" in the beginning right away says that it is parallel for related stuff (we don't really care that it is kernel because parallel_for->kernel_func, that was my idea). But OK, i'll revert this change

dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -51,17 +51,17 @@ DPCTLSyclEventRef dpnp_all_c(DPCTLSyclQueueRef q_ref,
     }
 
     sycl::queue q = *(reinterpret_cast<sycl::queue*>(q_ref));
-    sycl::event event;
 
     DPNPC_ptr_adapter<_DataType> input1_ptr(q_ref, array1_in, size);
     DPNPC_ptr_adapter<_ResultType> result1_ptr(q_ref, result1, 1, true, true);
```
Alexander-Makaryev, line 56: Could you please try to switch the first one `true` to `false` (parameter `target_no_sycl` = `false`)?
Then try to remove code with `DPNPC_ptr_adapter` at all.

PR #1155, opened 2022-04-04
remove mixed host\dev impl from dpnp_all dpnp_any

vladislav-markin:
Remove mixed host\dev implementation for 3 functions: dpnp_all_c(), dpnp_allclose_c() and dpnp_any_c().
Keep away from excess memcpy() calls between host and device memory for "result" pointer.
The "result" pointer has to be set with an initial value in kernel code rather on host one.
Add missing DPCTLEvent_Delete() invocations for absolute versions of function calls without SYCL arguments and return type.


oleksandr-pavlyk (approved):
It would be great if `dpnp_all_c` has a static_assert to check that `_ResultType` is `bool`.

oleksandr-pavlyk:
So I assume this PR has been superseded by gh-1301 and gh-1302. Please close if so.

PR #1168, opened 2022-08-17
Get rid of "Improper Null Termination" issue

antonwolfy:
Add a null-terminated symbol at the end of char array to avoid "Improper Null Termination" issue reported by Checkmarx scan.

### DIFF
--- a/dpnp/backend/kernels/dpnp_krnl_random.cpp
+++ b/dpnp/backend/kernels/dpnp_krnl_random.cpp
@@ -1691,7 +1691,7 @@ DPCTLSyclEventRef dpnp_rng_shuffle_c(DPCTLSyclQueueRef q_ref,
     {
         // Multidimensional ndarrays require a bounce buffer.
         size_t step_size = (size / high_dim_size) * itemsize; // size in bytes for x[i] element
-        void* buf = sycl::malloc_device<void *>(step_size, q);
+        void* buf = sycl::malloc_device(step_size, q);
         for (size_t i = uvec_size; i > 0; i--)
         {
             size_t j = (size_t)(floor((i + 1) * Uvec[i - 1]));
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
This change is not correct. Even though the pointer type is `char *` it is really meant as typeless pointer. The proper way to resolve the warning is to do `static_cast<void *>(char_ptr)` in the argument of `memcpy`.

The buffer is allocated of the size `itemsize` to store an item content, not a string.

antonwolfy:
> This change is not correct. Even though the pointer type is `char *` it is really meant as typeless pointer. The proper way to resolve the warning is to do `static_cast<void *>(char_ptr)` in the argument of `memcpy`.
> 
> The buffer is allocated of the size `itemsize` to store an item content, not a string.

The solution is reworked in a bit different way than suggested. The buffer initially is declared as a pointer on void, since no pointer arithmetic required. Also the code around was a bit simplified with direct q.memcpy() call instead of submitting the kernel with single mempy() to queue.

dpnp/backend/kernels/dpnp_krnl_random.cpp
```diff
@@ -1674,23 +1674,15 @@ DPCTLSyclEventRef dpnp_rng_shuffle_c(DPCTLSyclQueueRef q_ref,
         // Fast, statically typed path: shuffle the underlying buffer.
         // Only for non-empty, 1d objects of class ndarray (subclasses such
         // as MaskedArrays may not support this approach).
-        char* buf = reinterpret_cast<char*>(sycl::malloc_shared(itemsize * sizeof(char), q));
+        void* buf = sycl::malloc_device<void *>(itemsize, q);
```
oleksandr-pavlyk, line 1677:
```suggestion
        void* buf = sycl::malloc_device(itemsize, q);
```

dpnp/backend/kernels/dpnp_krnl_random.cpp
```diff
@@ -1699,23 +1691,15 @@ DPCTLSyclEventRef dpnp_rng_shuffle_c(DPCTLSyclQueueRef q_ref,
     {
         // Multidimensional ndarrays require a bounce buffer.
         size_t step_size = (size / high_dim_size) * itemsize; // size in bytes for x[i] element
-        char* buf = reinterpret_cast<char*>(sycl::malloc_shared(step_size * sizeof(char), q));
+        void* buf = sycl::malloc_device<void *>(step_size, q);
```
oleksandr-pavlyk, line 1694:
```suggestion
        void* buf = sycl::malloc_device(step_size, q);
```

oleksandr-pavlyk (approved):

antonwolfy:
@oleksandr-pavlyk , may I ask you to merge without passing the checks, no new issues observed there? (I personally don't have such permissions.)

PR #1172, opened 2022-08-25
dpnp_take failed on Windows due to memory corruption

antonwolfy:
DPNP backend had an implicit memory conversion from int32 to int64 type for incoming "indices" array when calling dpnp_take().
On Windows default integer type is int32, while on Linux int64. That is when calling numpy.array() with array of integer values but without dtype argument passed, elements of the resulting array will have type int32 - on Windows and int64 - on Linux hosts where the CI is executed.
Thus on Windows a memory for the "indices" array is allocated as for int32 data type. In DPNP backend the pointer of "indices" array data is casted to a pointer on int64 data type. This leads to a memory corruption.
The issue wasn't visible in Linux host, because "indices" array here had been allocated for int64 data type.

The previously skipped tests for dpnp_take running Windows host (potentially failed due to resolved here issue) are restored and included back to the verification scope.

The existing test was extended to explicitly pass int32 and int64 data type on both Linux and Windows hosts and in scope of checks for both CPU and GPU devices.


dpnp/backend/kernels/dpnp_krnl_indexing.cpp
```diff
@@ -1074,15 +1072,23 @@ void func_map_init_indexing_func(func_map_t& fmap)
                                                                         (void*)dpnp_put_along_axis_ext_c<double>};
 
     fmap[DPNPFuncName::DPNP_FN_TAKE][eft_BLN][eft_BLN] = {eft_BLN, (void*)dpnp_take_default_c<bool, int64_t>};
-    fmap[DPNPFuncName::DPNP_FN_TAKE][eft_INT][eft_INT] = {eft_INT, (void*)dpnp_take_default_c<int32_t, int64_t>};
+    fmap[DPNPFuncName::DPNP_FN_TAKE][eft_INT][eft_INT] = {eft_INT, (void*)dpnp_take_default_c<int32_t, int32_t>};
```
oleksandr-pavlyk, line 1075: Clearly these are not all combinations. What's happening in `dpnp` in case when the type combination is not found in `fmap`? 

I also think that `[eft_C128][eft_C128]` was really intended for the use-case where function pointer lookup depended only on the source array type, and not on the type of indices.

oleksandr-pavlyk (approved):

dpnp/backend/kernels/dpnp_krnl_indexing.cpp
antonwolfy: > Clearly these are not all combinations. What's happening in `dpnp` in case when the type combination is not found in `fmap`?

It will result in runtime exception, like:
`RuntimeError: DPNP Error: Function ID with unsupported second parameter type.`

> I also think that `[eft_C128][eft_C128]` was really intended for the use-case where function pointer lookup depended only on the source array type, and not on the type of indices.

Yep, agree. In case of numpy there will be a type error, like:
`TypeError: Cannot cast array data from dtype('complex256') to dtype('int64') according to the rule 'safe'`

oleksandr-pavlyk (approved):

PR #1185, opened 2022-09-20
[Build] setuptools 63.4.1 breaks build for Windows

antonwolfy:
DPNP build on Windows was previously successful in internal CI where setuptools=58.0.4. The same setuptools version is currently used in intel channel of Anaconda.
Switching to newer version setuptools=63.4.1 in internal CI broke the build flow due to non-backward compatible changed released in setuptools=60.0.0.
To prevent this and keep the legacy setuptools behavior, a workaround with setting environment variable SETUPTOOLS_USE_DISTUTILS="stdlib" is implemented.

### DIFF
--- a/conda-recipe/bld.bat
+++ b/conda-recipe/bld.bat
@@ -2,6 +2,11 @@ REM A workaround for activate-dpcpp.bat issue to be addressed in 2021.4
 set "LIB=%BUILD_PREFIX%\Library\lib;%BUILD_PREFIX%\compiler\lib;%LIB%"
 SET "INCLUDE=%BUILD_PREFIX%\include;%INCLUDE%"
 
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk:
This fix works because vendored `distutils` does not have changes we apply to cpython's distutils. Proper solution should transition windows build to use `cmake` which would render this change no longer needed.

oleksandr-pavlyk (approved):
Perhaps add a TODO note to remove setting of `SETUPTOOLS_USE_DISTUTILS` once transition to build backend on Windows to cmake is complete.

antonwolfy:
> Perhaps add a TODO note to remove setting of `SETUPTOOLS_USE_DISTUTILS` once transition to build backend on Windows to cmake is complete.

Done

PR #1189, opened 2022-09-26
Fix import dpnp on Win & python 3.10

antonwolfy:
In case of Windows OS t is required to explicitly define a path where to search for DLLs towards both DPNP backend and DPCTL Sycl interface, otherwise the DPNP import will fail. This is because the libraries are not installed under any of default paths where Python is looking for.

Python 3.10 modifies the behavior on how it looks for DLLs to make it more secure. An environment variable PATH isn't longer used as default DLL search location. Library provides must add it to DLL search path using 'add_dll_directory' utility for Python's internal dynamic library opener to be able to find it.

The solution working with all supported pythons is supposed to be a combination of two approaches with extending PATH env variable and with calling os.add_dll_directory() if available.

### DIFF
--- a/dpnp/__init__.py
+++ b/dpnp/__init__.py
@@ -39,8 +39,7 @@ if system() == 'Windows':
     if hasattr(os, "add_dll_directory"):
         os.add_dll_directory(mypath)
         os.add_dll_directory(dpctlpath)
-    else:
-        os.environ["PATH"] += os.pathsep + mypath + os.pathsep + dpctlpath
+    os.environ["PATH"] += os.pathsep + mypath + os.pathsep + dpctlpath
 
 
 from dpnp.dpnp_array import dpnp_array as ndarray
--- a/dpnp/__init__.py
+++ b/dpnp/__init__.py
@@ -30,10 +30,10 @@ mypath = os.path.dirname(os.path.realpath(__file__))
 import dpctl
 dpctlpath = os.path.dirname(dpctl.__file__)
 
-# For Windows OS, with Python >3.7, it is required to explicitly define a path where to search
-# for DLLs towards both DPNP backend and DPCTL Sycl interface, otherwise
-# DPNP import will be failing. This is because the libraries are not installed
-# under any of default paths where Python is searching.
+# For Windows OS with Python >= 3.7, it is required to explicitly define a path
+# where to search for DLLs towards both DPNP backend and DPCTL Sycl interface,
+# otherwise DPNP import will be failing. This is because the libraries
+# are not installed under any of default paths where Python is searching.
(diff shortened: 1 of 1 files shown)


dpnp/__init__.py
```diff
@@ -30,7 +30,16 @@
 import dpctl
 dpctlpath = os.path.dirname(dpctl.__file__)
 
-os.environ["PATH"] += os.pathsep + mypath + os.pathsep + dpctlpath
+# For Windows OS it is required to explicitly define a path where to search
+# for DLLs towards both DPNP backand and DPCTL Sycl interface, otherwise
```
oleksandr-pavlyk, line 34:
```suggestion
# for DLLs towards both DPNP backend and DPCTL Sycl interface, otherwise
```

dpnp/__init__.py
```diff
@@ -30,7 +30,16 @@
...
+# DPNP import will be failing. This is because the libraries are not installed
+# under any of default paths where Python is looking for.
```
oleksandr-pavlyk, line 36:
```suggestion
# under any of default paths where Python is searching.
```

dpnp/__init__.py
```diff
@@ -30,7 +30,16 @@
 import dpctl
 dpctlpath = os.path.dirname(dpctl.__file__)
 
-os.environ["PATH"] += os.pathsep + mypath + os.pathsep + dpctlpath
+# For Windows OS it is required to explicitly define a path where to search
```
oleksandr-pavlyk, line 33:
```suggestion
# For Windows OS, with Python >3.7, it is required to explicitly define a path where to search
```

dpnp/__init__.py
```diff
@@ -30,7 +30,16 @@
...
+# for DLLs towards both DPNP backand and DPCTL Sycl interface, otherwise
+# DPNP import will be failing. This is because the libraries are not installed
+# under any of default paths where Python is looking for.
+from platform import system
+if system() == 'Windows':
+    if hasattr(os, "add_dll_directory"):
+        os.add_dll_directory(mypath)
+        os.add_dll_directory(dpctlpath)
+    os.environ["PATH"] += os.pathsep + mypath + os.pathsep + dpctlpath
```
oleksandr-pavlyk, line 42:
```suggestion
    os.environ["PATH"] = os.pathsep.join([os.getenv("PATH", ""), mypath, dpctlpath])
```

oleksandr-pavlyk (approved):
Look good overall, feedback provided

antonwolfy:
All the comments are applied.

oleksandr-pavlyk (commented):
Looks good to me now!

PR #1191, opened 2022-09-29
Fix github action for building docs

antonwolfy:
Since spinx 5.0.0, autosummary extension will trigger an error if any input rst file includes unknown module.
Thus, as a fix. the modules which aren't currently implemented in DPNP have been commented in the rst files.

Additionally build docs action will be partly running on each push request (excluding deploy step) now. It should help to find an issue similar to the fault observing with a newer spinx version on early stage (previously the action was running on PR merge only).

### DIFF
--- a/.github/workflows/build-sphinx.yml
+++ b/.github/workflows/build-sphinx.yml
@@ -71,7 +71,7 @@ jobs:
 
       - name: Install dpnp dependencies
         run: |
-          conda install dpctl mkl-devel-dpcpp tbb-devel dpcpp_linux-64 cmake=3.19 cython pytest \
+          conda install dpctl mkl-devel-dpcpp tbb-devel dpcpp_linux-64 cmake cython pytest \
                         -c dppy/label/dev -c intel -c conda-forge
 
       - name: Install cuPy dependencies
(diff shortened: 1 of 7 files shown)


.github/workflows/build-sphinx.yml
```diff
@@ -3,89 +3,109 @@ on:
...
       - name: Install sphinx dependencies
-        shell: bash -l {0}
-        run: |
-          conda install sphinx sphinx_rtd_theme
+        run: conda install sphinx sphinx_rtd_theme
+
       - name: Install dpnp dependencies
-        shell: bash -l {0}
         run: |
           conda install dpctl mkl-devel-dpcpp tbb-devel dpcpp_linux-64 cmake=3.19 cython pytest \
```
oleksandr-pavlyk, line 74: Is there a reason to pin cmake to 3.19 ? I think newer cmake may do just as well

oleksandr-pavlyk (approved):

.github/workflows/build-sphinx.yml
antonwolfy: I agree, we can remove the pinning.

PR #1201, opened 2022-10-17
Raise not-implemented exception on numpy fallback

antonwolfy:
Previously DPNP fell back on NumPy implementation when a requested functionality isn't available and can't be provided by DPNP.
The PR is indented to raise `Non-implemented` exception in that scenario by default, rather than falling back.

It will be possible to keep old behavior with falling back on NumPy calls by setting a new environment variable `DPNP_RAISE_EXCEPION_ON_NUMPY_FALLBACK = 0`.
In all other cases, the fallback is prohibited and lead to the exception.

### DIFF
--- a/dpnp/dpnp_utils/dpnp_algo_utils.pyx
+++ b/dpnp/dpnp_utils/dpnp_algo_utils.pyx
@@ -114,9 +114,13 @@ def call_origin(function, *args, **kwargs):
     Call fallback function for unsupported cases
     """
 
-    if config.__DPNP_RAISE_EXCEPION_ON_NUMPY_FALLBACK__:
+    allow_fallback = kwargs.pop("allow_fallback", False)
+
+    if not allow_fallback and config.__DPNP_RAISE_EXCEPION_ON_NUMPY_FALLBACK__ == 1:
         raise NotImplementedError(f"Requested funtion={function.__name__} with args={args} and kwargs={kwargs} "
-                                  "isn't currently supported and would fall back on NumPy implementation.")
+                                   "isn't currently supported and would fall back on NumPy implementation. "
+                                   "Define enviroment variable `DPNP_RAISE_EXCEPION_ON_NUMPY_FALLBACK` to `0` "
+                                   "if the fall back is required to be supported without rasing an exception.")
 
     dpnp_inplace = kwargs.pop("dpnp_inplace", False)
     sycl_queue = kwargs.pop("sycl_queue", None)
--- a/dpnp/dpnp_utils/dpnp_algo_utils.pyx
+++ b/dpnp/dpnp_utils/dpnp_algo_utils.pyx
@@ -1,7 +1,7 @@
 # cython: language_level=3
 # -*- coding: utf-8 -*-
 # *****************************************************************************
-# Copyright (c) 2016-2020, Intel Corporation
+# Copyright (c) 2016-2022, Intel Corporation
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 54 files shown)


dpnp/dpnp_utils/dpnp_algo_utils.pyx
```diff
@@ -114,6 +114,10 @@ def call_origin(function, *args, **kwargs):
     Call fallback function for unsupported cases
     """
 
+    if config.__DPNP_RAISE_EXCEPION_ON_NUMPY_FALLBACK__:
+        raise NotImplementedError(f"Requested funtion={function.__name__} with args={args} and kwargs={kwargs} "
+                                  "isn't currently supported, fallback on NumPy implementation.")
```
oleksandr-pavlyk, line 119:
```suggestion
                                  "isn't currently supported, falls back on NumPy implementation.")
```

oleksandr-pavlyk: or would fall back on NumPy implementation

antonwolfy, line 119:
```suggestion
                                  "isn't currently supported and would fall back on NumPy implementation.")
```

oleksandr-pavlyk:
Implementation in this PR falls back by default, allows opting out.  I should clarify, but I thought we agreed to raises by default, permits opt-in.

oleksandr-pavlyk:
> Implementation in this PR falls back by default, allows opting out. I should clarify, but I thought we agreed to raises by default, permits opt-in.

Never mind I misread the code, it works as expected

dpnp/random/dpnp_random_state.py
```diff
@@ -125,12 +125,15 @@ def normal(self, loc=0.0, scale=1.0, size=None, dtype=None, usm_type="device"):
...
                 max_double = numpy.finfo('double').max
-                if (loc >= max_double or loc <= min_double) and dpnp.isfinite(loc):
+
+                # TODO: switch to dpnp.isfinite() and dpnp.signbit() once functions are available,
+                # but for now use direct numpy calls without call_origin() wrapper, since data is a scalar
+                if (loc >= max_double or loc <= min_double) and numpy.isfinite(loc):
                     raise OverflowError(f"Range of loc={loc} exceeds valid bounds")
 
-                if (scale >= max_double) and dpnp.isfinite(scale):
+                if (scale >= max_double) and numpy.isfinite(scale):
```
oleksandr-pavlyk, line 134: Perhaps factor out uses of `numpy.isfinite` into your own routine which can be modified as needed. 

If `scale` always remains a Python or NumPy scalar, than not compute-follows-data violation occurs, but if `scale` can be an array, then your modular routine can be modified to apply appropriate validation function.

tests/conftest.py
```diff
@@ -73,3 +73,7 @@ def pytest_collection_modifyitems(config, items):
             # exact match of the test name with items from excluded_list
             if test_name == item_tbl_str:
                 item.add_marker(skip_mark)
+
+@pytest.fixture
+def allow_fall_back_on_numpy(monkeypatch):
+    monkeypatch.setattr(dpnp.config, '__DPNP_RAISE_EXCEPION_ON_NUMPY_FALLBACK__', 0)
```
oleksandr-pavlyk, line 79: Cool trick!

dpnp/random/dpnp_random_state.py
antonwolfy: Added internal functions _is_finite_scalar() and _is_finite_scalar() to dpnp_random_state.py with TODO comment inside to replace with dpnp functionality once availbale.

oleksandr-pavlyk (approved):

PR #1210, opened 2022-10-21
Import of deprecated dpnp.dparray class caused exception

antonwolfy:
A `dpnp.dparray` class is actually obsolete and needs to be declared as deprecated. There is `dpnp.dpnp_array` class which has to be used instead. So the PR implements a deprecation warning for `dpnp.dparray` class.
Another issue which is fixed by the PR is an exception raised during import of `dpnp.dparray`, like in below example:
```
> python -c "import dpnp; import dpnp.dparray"
  File "<string>", line 1, in <module>
  File "dpnp/dparray.pyx", line 40, in init dpnp.dparray
    from dpnp.dpnp_iface import *
TypeError: Cannot overwrite C type array
```

### DIFF
--- a/dpnp/dparray.pyx
+++ b/dpnp/dparray.pyx
@@ -1,7 +1,7 @@
 # cython: language_level=3
 # -*- coding: utf-8 -*-
 # *****************************************************************************
-# Copyright (c) 2016-2020, Intel Corporation
+# Copyright (c) 2016-2022, Intel Corporation
 # All rights reserved.
 #
 # Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


dpnp/dparray.pyx
```diff
@@ -124,6 +127,12 @@ cdef class dparray:
 
     """
 
+    with warnings.catch_warnings():
```
oleksandr-pavlyk, line 130: I think we want this warning to be user visible, not [suppress it](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings).

dpnp/dparray.pyx
```diff
@@ -33,23 +33,26 @@ using USB interface for an Intel GPU device.
 
 """
 
+import warnings
 
 from libcpp cimport bool as cpp_bool
 
 from dpnp.dpnp_iface_types import *
-from dpnp.dpnp_iface import *
```
oleksandr-pavlyk, line 40: Is importing of this module fixed by changes on lines 52-55 without removal of `from dpnp.dpnp_iface import *`.

dpnp/dparray.pyx
antonwolfy: No, it will not suppress, it will create a context which copies the current filter and will restore upon exit from the context, as I got from [here](https://docs.python.org/3/library/warnings.html#available-context-managers).
The idea was to modify the filter with enabling deprecation warnings within the context, but without impact on the global filter.

dpnp/dparray.pyx
antonwolfy: No, it didn't help, the same exception raised.

dpnp/dparray.pyx
oleksandr-pavlyk: > I think we want this warning to be user visible, not [suppress it](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings).

I am in the wrong here. The `filterwarnings` turns the filter on, and use of context manager ensures its use does not alter users's setting of warning filtering.

PR #1218, opened 2022-11-02
Use full() function from dpctl.tensor

npolina4:
Use full() function from dpctl.tensor module instead of DPNP backend implementation.
A cython code relating to full() call is cleaned up.


antonwolfy (commented):
Also it would be great to extend test_array_creation() test in `tests/test_sycl_queue.py` and to add new test_full() in `tests/test_arraycreation.py` covering basic functionality and corner cases.

The following tests needs to be excluded from ignore list, i.e. to be removed from `tests/skipped_tests_gpu.tbl` file:
```
tests/third_party/cupy/creation_tests/test_basic.py::TestBasic::test_full
tests/third_party/cupy/creation_tests/test_basic.py::TestBasic::test_full_default_dtype
tests/third_party/cupy/creation_tests/test_basic.py::TestBasic::test_full_default_dtype_cpu_input
```

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -697,14 +704,15 @@ def full(shape, fill_value, dtype=None, order='C'):
...
-    if not use_origin_backend():
-        if order not in ('C', 'c', None):
-            pass
-        else:
-            if dtype is None:
-                dtype = numpy.array(fill_value).dtype.type  # TODO simplify
-
-            return dpnp_full(shape, fill_value, dtype).get_pyobj()
+    if like is None:
+        if dtype is None:
```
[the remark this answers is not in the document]
npolina4: This case work with  fill_value=xp.array(1, dtype=dtype) and cannot be removed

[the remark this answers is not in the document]
antonwolfy: I believe, dpctl should support `fill_value` as `usm_ndarray` in case `dtype` is `None` and such logic needs to be moved into dpctl code, since NumPy supports the same case.
While in dpnp we need to add a check to `dpnp_container.py` to explicitly pass `fill_value` as `usm_ndarray`, like:
```
    """Converts `fill_value` to `usm_ndarray`"""
    value = fill_value.get_array() if isinstance(fill_value, dpnp_array) else fill_value

    """Creates `dpnp_array` where every element is equal to fill_value."""
    array_obj = dpt.full(shape,
                         value,
                         dtype=dtype,
                         order=order,
                         usm_type=usm_type,
                         sycl_queue=sycl_queue_normalized)
```

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -685,29 +695,43 @@ def full(shape, fill_value, dtype=None, order='C'):
...
+              sycl_queue=None):
     """
     Return a full array with the same shape and type as a given array.
 
     For full documentation refer to :obj:`numpy.full_like`.
 
     Limitations
     -----------
-    Parameter ``order`` is supported only with default value ``"C"``.
+    Parameter ``order`` is supported only with values ``"C"`` and ``"F"``.
```
antonwolfy, line 732:
```suggestion
    Parameters ``x1`` is supported only as :class:`dpnp.dpnp_array`.
    Parameter ``order`` is supported only with values ``"C"`` and ``"F"``.
```

tests/test_arraycreation.py
```diff
@@ -348,3 +348,77 @@ def test_vander(array, type, n, increase):
...
+    assert ia.flags.c_contiguous == a.flags.c_contiguous
+    assert ia.flags.f_contiguous == a.flags.f_contiguous
+    assert numpy.array_equal(dpnp.asnumpy(ia), a)
+
+
+def test_full_strides():
+    a = numpy.full((3, 3), numpy.arange(3, dtype="i4"))
+    ia = dpnp.full((3, 3), dpnp.arange(3, dtype="i4"))
+    assert ia.strides == tuple(el // a.itemsize for el in a.strides)
+    assert numpy.array_equal(dpnp.asnumpy(ia), a)
```
antonwolfy, line 412:
```suggestion
    numpy.testing.assert_array_equal(dpnp.asnumpy(ia), a)
```

antonwolfy (approved):

dpnp/dpnp_container.py
```diff
@@ -111,3 +112,28 @@ def empty(shape,
...
+         usm_type=None,
+         sycl_queue=None):
+    """Validate input parameters before passing them into `dpctl.tensor` module"""
+    dpu.validate_usm_type(usm_type, allow_none=True)
+    sycl_queue_normalized = dpnp.get_normalized_queue_device(fill_value, sycl_queue=sycl_queue, device=device)
+
+    if isinstance(fill_value, dpnp_array):
+        fill_value = fill_value.get_array()
+
+    """Creates `dpnp_array` where every element is equal to fill_value."""
```
oleksandr-pavlyk, line 132: This docstring is blatantly wrong when `fill_value` is an array. NumPy says "Return a new array of given shape and type, filled with `fill_value`."

oleksandr-pavlyk (approved):
LGTM. It would be nice to try numba_dpex with this change in dpnp to verify that `dpnp.full` stays properly supported in JIT-ted functions.

PR #1223, opened 2022-11-07
Use zeros() function from dpctl.tensor.

npolina4:
Use zeros() function from dpctl.tensor module instead of DPNP backend implementation.
A cython code relating to zeros() call is cleaned up.

### DIFF
--- a/dpnp/dpnp_iface_arraycreation.py
+++ b/dpnp/dpnp_iface_arraycreation.py
@@ -1341,7 +1341,14 @@ def vander(x1, N=None, increasing=False):
     return call_origin(numpy.vander, x1, N=N, increasing=increasing)
 
 
-def zeros(shape, dtype=None, order='C'):
+def zeros(shape,
+          *,
+          dtype=None,
+          order="C",
+          like=None,
+          device=None,
+          usm_type="device",
+          sycl_queue=None):
     """
     Return a new array of given shape and type, filled with zeros.
 
@@ -1349,7 +1356,8 @@ def zeros(shape, dtype=None, order='C'):
 
     Limitations
     -----------
-    Parameter ``order`` is supported only with default value ``"C"``.
+    Parameter ``like`` is supported only with default value ``None``.
+    Otherwise the function will be executed sequentially on CPU.
 
     See Also
     --------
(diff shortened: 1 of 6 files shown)


oleksandr-pavlyk:
Have we checked whether removal of `DPNP_FN_ZEROS_EXT` is ok with `numba_dpex` team?

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -1415,15 +1429,16 @@ def zeros_like(x1, dtype=None, order='C', subok=False, shape=None):
 
     x1_desc = dpnp.get_dpnp_descriptor(x1, copy_when_nondefault_queue=False)
```
antonwolfy, line 1430:
```suggestion
    if subok is False:
        x1_desc = dpnp.get_dpnp_descriptor(x1, copy_when_nondefault_queue=False)
        if x1_desc:
            _shape = shape if shape is not None else x1_desc.shape
            _dtype = dtype if dtype is not None else x1_desc.dtype
```

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -1370,29 +1378,35 @@ def zeros(shape, dtype=None, order='C'):
...
+               usm_type="device",
+               sycl_queue=None):
     """
     Return an array of zeros with the same shape and type as a given array.
 
     For full documentation refer to :obj:`numpy.zeros_like`.
 
     Limitations
     -----------
-    Parameter ``order`` is supported only with default value ``"C"``.
```
antonwolfy, line 1395: Order is still limited with "C" and "F" in dpctl, and default is "C" while "K" in NumPy. I think we need to update the limitation, rather than to remove it.

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -1370,29 +1378,35 @@ def zeros(shape, dtype=None, order='C'):
...
-            _dtype = dtype if dtype is not None else dpnp.float64
-            result = dpnp_zeros(shape, _dtype).get_pyobj()
-
-            return result
-
-    return call_origin(numpy.zeros, shape, dtype=dtype, order=order)
+    return call_origin(numpy.zeros, shape, dtype=dtype, order=order, like=like)
 
 
 # numpy.zeros_like(a, dtype=None, order='K', subok=True, shape=None)
```
antonwolfy, line 1393: remove obsolete comment

PR #1224, opened 2022-11-07
Leave is_host() call with old compiler only

antonwolfy:
Since SYCL DPC++ 2023 compiler the function is_host() is deprecated as the host device is no longer supported.
The PR excludes is_host() call from the code when the new compiler is used, i.e. when `__SYCL_COMPILER_VERSION < 20221102L`.


oleksandr-pavlyk (approved):

oleksandr-pavlyk:
Related change in `dpctl`: https://github.com/IntelPython/dpctl/pull/987/

PR #1226, opened 2022-11-09
Use icpx instead of dpcpp on Linux

antonwolfy:
Since SYCL DPC++ 2023 compiler use of `dpcpp` is deprecated and will be removed in a future release.
The PR replaces `dpcpp` with use of 'icpx -fsycl' in case of building for Linux OS.

### DIFF
--- a/.github/workflows/build-sphinx.yml
+++ b/.github/workflows/build-sphinx.yml
@@ -85,7 +85,7 @@ jobs:
       - name: Build library
         run: |
           python setup.py build_clib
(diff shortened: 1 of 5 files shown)


oleksandr-pavlyk:
Hopefully one day `dpnp` gets to transition to using CMake, in which case both Windows and Linux would be addressed by a single change in `CMakeLists.txt`. 

Look good for now!

oleksandr-pavlyk (approved):

PR #1255, opened 2022-12-13
Use zeros() function from dpctl.tensor.

npolina4:
Use zeros() function from dpctl.tensor module instead of DPNP backend implementation.
A cython code relating to zeros() call is cleaned up.

### DIFF
--- a/dpnp/dpnp_container.py
+++ b/dpnp/dpnp_container.py
@@ -46,6 +46,7 @@ __all__ = [
     "asarray",
     "empty",
     "full",
+    "zeros",
 ]
 
 
@@ -136,4 +137,26 @@ def full(shape,
                          order=order,
                          usm_type=usm_type,
                          sycl_queue=sycl_queue_normalized)
+    return dpnp_array(array_obj.shape, buffer=array_obj, order=order)
+
+
+def zeros(shape,
+          *,
+          dtype=None,
+          order="C",
+          device=None,
+          usm_type="device",
+          sycl_queue=None):
+    """Validate input parameters before passing them into `dpctl.tensor` module"""
+    dpu.validate_usm_type(usm_type, allow_none=False)
+    sycl_queue_normalized = dpnp.get_normalized_queue_device(sycl_queue=sycl_queue, device=device)
+    if order is None:
+        order = 'C'
+
+    """Creates `dpnp_array` with zero elements."""
+    array_obj = dpt.zeros(shape,
+                          dtype=dtype,
+                          order=order,
+                          usm_type=usm_type,
+                          sycl_queue=sycl_queue_normalized)
     return dpnp_array(array_obj.shape, buffer=array_obj, order=order)
--- a/dpnp/dpnp_container.py
+++ b/dpnp/dpnp_container.py
@@ -87,6 +87,8 @@ def asarray(x1,
         x1_obj = x1
 
     sycl_queue_normalized = dpnp.get_normalized_queue_device(x1_obj, device=device, sycl_queue=sycl_queue)
+    if order is None:
+        order = 'C'
 
     """Converts incoming 'x1' object to 'dpnp_array'."""
     array_obj = dpt.asarray(x1_obj,
@@ -126,6 +128,8 @@ def full(shape,
     """Validate input parameters before passing them into `dpctl.tensor` module"""
     dpu.validate_usm_type(usm_type, allow_none=True)
     sycl_queue_normalized = dpnp.get_normalized_queue_device(fill_value, sycl_queue=sycl_queue, device=device)
+    if order is None:
+        order = 'C'
 
     if isinstance(fill_value, dpnp_array):
         fill_value = fill_value.get_array()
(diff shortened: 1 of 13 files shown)


dpnp/dpnp_iface_arraycreation.py
```diff
@@ -1372,15 +1372,24 @@ def vander(x1, N=None, increasing=False):
...
 
 
-def zeros(shape, dtype=None, order='C'):
+def zeros(shape,
+          *,
+          dtype=numpy.float64,
+          order="C",
+          like=None,
+          device=None,
+          usm_type="device",
```
antonwolfy, line 1381:
```suggestion
          usm_type=None,
```
to align with dpnp.asarray() and dpnp.full().
Or is there a reason why dpctl.tensor has different default value for `usm_type` (even in zeros_like() it is `None`) ? Or probably a support of `None` is just missing in dpctl and needs to be added?

dpnp/dpnp_container.py
```diff
@@ -137,3 +138,25 @@ def full(shape,
...
                          sycl_queue=sycl_queue_normalized)
     return dpnp_array(array_obj.shape, buffer=array_obj, order=order)
+
+
+def zeros(shape,
+          *,
+          dtype=None,
+          order="C",
+          device=None,
+          usm_type="device",
```
antonwolfy, line 148:
```suggestion
          usm_type=None,
```

dpnp/dpnp_container.py
```diff
@@ -137,3 +138,25 @@ def full(shape,
...
+          sycl_queue=None):
+    """Validate input parameters before passing them into `dpctl.tensor` module"""
+    dpu.validate_usm_type(usm_type, allow_none=False)
+    sycl_queue_normalized = dpnp.get_normalized_queue_device(sycl_queue=sycl_queue, device=device)
+    if order is None:
+        order = 'C'
```
antonwolfy, line 154: Should we add the same for other constructors, like full() and asarray() ?

dpnp/dpnp_container.py
```diff
@@ -137,3 +138,25 @@ def full(shape,
...
+
+    """Creates `dpnp_array` with zero elements."""
```
antonwolfy, line 156:
```suggestion
    """Creates `dpnp_array` of zeros with the given shape, dtype, and order."""
```

tests/test_arraycreation.py
```diff
@@ -418,3 +418,32 @@ def test_full_strides():
...
+                         [(), 0, (0,), (2, 0, 3), (3, 2)],
+                         ids=['()', '0', '(0,)', '(2, 0, 3)', '(3, 2)'])
+@pytest.mark.parametrize("dtype",
+                         [numpy.complex128, numpy.complex64, numpy.float64, numpy.float32, numpy.float16, numpy.int64, numpy.int32],
+                         ids=['complex128', 'complex64', 'float64', 'float32', 'float16', 'int64', 'int32'])
+def test_zeros(shape, dtype):
+    expected = numpy.zeros(shape, dtype=dtype)
+    result = dpnp.zeros(shape, dtype=dtype)
+
+    assert expected.dtype == result.dtype
```
antonwolfy, line 433: I guess it is a part of numpy.testing.assert_array_equal() call and can be removed.

tests/test_arraycreation.py
```diff
@@ -418,3 +418,32 @@ def test_full_strides():
...
         dpnp.full(10, fill_value=fill_value)
+
+
+@pytest.mark.parametrize("shape",
+                         [(), 0, (0,), (2, 0, 3), (3, 2)],
+                         ids=['()', '0', '(0,)', '(2, 0, 3)', '(3, 2)'])
+@pytest.mark.parametrize("dtype",
+                         [numpy.complex128, numpy.complex64, numpy.float64, numpy.float32, numpy.float16, numpy.int64, numpy.int32],
+                         ids=['complex128', 'complex64', 'float64', 'float32', 'float16', 'int64', 'int32'])
+def test_zeros(shape, dtype):
```
antonwolfy, line 429: Probably it would be good to parametrize `order` argument also here and in test_zeros_like().

antonwolfy (approved):

dpnp/dpnp_iface_arraycreation.py
oleksandr-pavlyk: For `asarray`, the `usm_type=None` is take to understand as infer the `usm_type` of the result from input array (if present). This does not apply to `zeros`. Default value of `usm_type` should be `"device"`.

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -1442,19 +1462,22 @@ def zeros_like(x1, dtype=None, order='C', subok=False, shape=None):
...
+    if not isinstance(x1, dpnp.ndarray):
+        pass
+    elif order not in ('C', 'c', 'F', 'f', None):
+        pass
+    elif subok is not False:
+        pass
+    else:
+        _shape = shape if shape is not None else x1.shape
+        _dtype = dtype if dtype is not None else x1.dtype
+        _usm_type = usm_type if usm_type is not None else x1.usm_type
```
oleksandr-pavlyk, line 1475: Simpler: 
```suggestion
        _usm_type = x1.usm_type if usm_type is None else usm_type
```

PR #1271, opened 2023-01-18
Use eye() function from dpctl.tensor.

npolina4:
Use eye() function from dpctl.tensor module instead of DPNP backend implementation.
A cython code relating to eye() call is cleaned up.


dpnp/dpnp_iface_arraycreation.py
```diff
@@ -572,29 +572,40 @@ def empty_like(x1,
...
-            pass
-        elif order != 'C':
-            pass
-        elif len(kwargs) != 0:
-            pass
-        else:
-            return dpnp_eye(N, M=M, k=k, dtype=dtype).get_pyobj()
+    if order not in ('C', 'c', 'F', 'f', None):
+        pass
+    elif len(kwargs) != 0:
```
antonwolfy, line 598: I think we need to add `like` argument as it did in numpy

dpnp/dpnp_iface_arraycreation.py
```diff
@@ -572,29 +572,40 @@ def empty_like(x1,
...
     """
     Return a 2-D array with ones on the diagonal and zeros elsewhere.
     For full documentation refer to :obj:`numpy.eye`.
 
     Limitations
     -----------
-    Input array is supported as :obj:`dpnp.ndarray`.
-    Parameters ``order`` is supported only with default value.
+    Parameter ``order`` is supported only with values ``"C"`` and ``"F"``.
+    Otherwise the function will be executed sequentially on CPU.
```
antonwolfy, line 593: need to add a limitation for `kwargs` and `like` arguments

tests/test_arraycreation.py
```diff
@@ -101,20 +101,34 @@ def test_diag(v, k):
...
+                         ids=['None', '0', '1', '2', '3'])
+@pytest.mark.parametrize("dtype",
+                         [None, numpy.complex128, numpy.complex64, numpy.float64, numpy.float32,
+                          numpy.float16, numpy.int64, numpy.int32, numpy.bool],
+                         ids=['None', 'complex128', 'complex64', 'float64', 'float32',
+                         'float16', 'int64', 'int32', 'bool'])
+@pytest.mark.parametrize("order",
+                         [None, "C", "F"],
+                         ids=['None', 'C', 'F'])
+def test_eye(N, M, dtype, order):
```
antonwolfy, line 117: I think need to parametrize 'k` argument here instead of separate test below (it will allow us to secure correct behavior with different combination of input parameters rather than testing only different `k` value)

tests/test_sycl_queue.py
```diff
@@ -81,6 +81,9 @@ def vvsort(val, vec, size, xp):
         pytest.param("full",
                      (2,2),
                      {'fill_value': 5}),
+        pytest.param("eye",
+                     4,
```
antonwolfy, line 85: please add `M` here also

antonwolfy (approved):

Issue #1028, opened 2021-10-23
Access `dpctl.tensor.usm_array` attributes via `dpnp_array` 


PokhodenkoSA:
`dpnp_array` is a wrapper for `dpctl.tensor.usm_array` and does not provide access to attributes `usm_type` and `usm_data`.
Also `_array_obj` is not public. I think client code should not access internal attributes.

What is a recommended way to access attributes from `usm_array` via `dpnp_array` ?

PR #1274, opened 2023-01-20
Get rid of 'has_aspect_host' property in tests

antonwolfy:
A device property `has_aspect_host` was removed from `dpctl` in PR [#1028](https://github.com/IntelPython/dpctl/pull/1028).
This PR is to support a use case in `dpnp` tests when a device doesn't have `has_aspect_host` attribute.


tests/test_sycl_queue.py
```diff
@@ -17,7 +17,7 @@
     "cpu",
 ]
 
-available_devices = [d for d in dpctl.get_devices() if not d.has_aspect_host]
+available_devices = [d for d in dpctl.get_devices() if hasattr(d, 'has_aspect_host') and not d.has_aspect_host]
```
oleksandr-pavlyk, line 20:
```suggestion
available_devices = [d for d in dpctl.get_devices() if getattr(d, 'has_aspect_host', False)]
```

PR #1276, opened 2023-01-23
Add support of NumPy 1.24

antonwolfy:
This PR is adaptation of `dpnp` with numpy 1.24 which has been released.

The below list of introduced numpy changes which impact `dpnp` and need to be resolved:
1. deprecation aliases `np.object`, `np.bool`, `np.float`, `np.complex`, `np.str` and `np.int` have been expired and removed
2. keyword argument `normed` has been removed from `np.histogram`, `np.histogram2d` and `np.histogramdd` (need to use `density` instead)
3. `np.arange` and `np.indices` now raise `TypeError` instead of `ValueError` for boolean dtype when it's requested to generate an array of length > 2
4. no `warnings` module in numpy namespace


oleksandr-pavlyk (approved):
These changes are great! I hope they reduce the number of warnings spewed by running of the test suite.

PR #1283, opened 2023-01-29
Get rid of testing the array creation functions with unsupported types by a default device in positive scenarios

antonwolfy:
The PR assumes that array creation functions will be tested in positive scenarios with a list of only supported types (which is based on capability of a default device where the tests are executing on).
Previously, the tests include list of all possible types and they check if an exception is raised by dpctl when a type is out of support on target device.
Instead, the PR proposes to use a common parametrization approach to iterate over the list of supported types, like:
```
@pytest.mark.parametrize("dtype", get_all_dtypes())
```

Additionally, an existing dpnp function `dpnp.default_float_type()` is updated to consider the device capabilities in return type.

### DIFF
--- a/tests/helper.py
+++ b/tests/helper.py
@@ -1,0 +1,39 @@
+import dpctl
+import dpnp
+
+
+def get_all_dtypes(no_bool=False,
+                   no_float16=True,
+                   no_complex=False,
+                   no_none=False,
+                   device=None):
+    """
+    Build a list of types supported by DPNP based on input flags and device capabilities.
+    """
+
+    dev = dpctl.select_default_device() if device is None else device
+
+    # add boolean type
+    dtypes = [dpnp.bool] if not no_bool else []
+
+    # add integer types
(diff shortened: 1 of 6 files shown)


tests/helper.py
```diff
@@ -0,0 +1,39 @@
...
+def get_all_dtypes(no_bool=False,
+                   no_float16=True,
+                   no_complex=False,
+                   no_none=False,
+                   device=None):
+    """
+    Build a list of types supported by DPNP based on input flags and device capabilities.
+    """
+
+    dev = dpctl.SyclQueue().sycl_device if device is None else device
```
oleksandr-pavlyk, line 14:
```suggestion
    dev = dpctl.select_default_device() if device is None else device
```

oleksandr-pavlyk: The `SyclQueue().sycl_device` does a lot more work, also creating a context and structures for scheduler.

oleksandr-pavlyk (approved):
LGTM, but I only looked superficially.

Issue #1273, opened 2023-01-19
Division by scalar raise NotImplementedError


npolina4:
``` python
import dpnp as xp
a = xp.array([2., 4.])
a/2.
```
``` python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/localdisk/work/npolina/ws/dpnp/dpnp/dpnp_array.py", line 304, in __truediv__
    return dpnp.true_divide(self, other)
  File "/localdisk/work/npolina/ws/dpnp/dpnp/dpnp_iface_mathematical.py", line 1675, in true_divide
    return dpnp.divide(*args, **kwargs)
  File "/localdisk/work/npolina/ws/dpnp/dpnp/dpnp_iface_mathematical.py", line 585, in divide
    return call_origin(numpy.divide, x1, x2, dtype=dtype, out=out, where=where, **kwargs)
  File "dpnp/dpnp_utils/dpnp_algo_utils.pyx", line 132, in dpnp.dpnp_utils.dpnp_algo_utils.call_origin
    raise NotImplementedError(f"Requested funtion={function.__name__} with args={args} and kwargs={kwargs} "
NotImplementedError: Requested funtion=divide with args=(<dpnp.dpnp_array.dpnp_array object at 0x7f11c51ef9d0>, 2.0) and kwargs={'dtype': None, 'out': None, 'where': True} isn't currently supported and would fall back on NumPy implementation. Define enviroment variable `DPNP_RAISE_EXCEPION_ON_NUMPY_FALLBACK` to `0` if the fall back is required to be supported without rasing an exception.
```

npolina4:
Same problem with subtract and add.

antonwolfy:
@npolina4, the issue was fixed by above PRs and merged.

PR #1288, opened 2023-02-03
dpnp.add() doesn't work properly with a scalar

antonwolfy:
Since the change implemented in #1201 and until now, implicit `operation +` of array with a scalar or explicit one through `dpnp.add` call leads to raising of the not-implemented exception in dpnp. This is because dpnp was falling back on NumPy implementation in that use case, but has to call an internal kernel function from the backend.
The issue was reported in #1273.

The PR proposes to copy a scalar data from host memory into USM allocated memory, which means to create zero-dimension dpnp array from the scalar. This way the add of array with a scalar will be handled in the same way as addition of two arrays, which already works fine in dpnp.

Also, minor fixes in `dpnp.multiply` implementation were included in the PR:
- use `oneapi::mkl::vm::mul()` also for complex types
- fallback on numpy or raise an exception if both inputs are scalars
- control USM type of output array to be compliant with compute follows data paradigm

A common macro `MACRO_2ARG_3TYPES_OP` used in all elementwise arithmetical operation was refactor to get rid of excess memory copy in shared memory.

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -17,7 +17,6 @@ env:
       test_linalg.py
       test_mathematical.py
       test_random_state.py
-      test_special.py
       test_usm_type.py
   VER_JSON_NAME: 'version.json'
   VER_SCRIPT1: "import json; f = open('version.json', 'r'); j = json.load(f); f.close(); "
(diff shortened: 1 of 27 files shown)


oleksandr-pavlyk (approved):
Built, ran few examples. Looks good! Thank you @antonwolfy

oleksandr-pavlyk (approved):

PR #1292, opened 2023-02-11
dpnp.subtract() doesn't work properly with a scalar

antonwolfy:
Since the change implemented in #1201 and until now, implicit `operation -` of array with a scalar or explicit one through `dpnp.subtract` leads to raising of the not-implemented exception in dpnp. This is because dpnp was falling back on NumPy implementation in that use case, but has to call an internal kernel function from the backend.
The issue was reported in https://github.com/IntelPython/dpnp/issues/1273.

The PR proposes to copy a scalar data from host memory into USM allocated memory, which means to create zero-dimension dpnp array from the scalar. This way the subtraction of array with a scalar will be handled in the same way as subtraction of two arrays, which already works fine in dpnp.

Also, missing `__rsub__` function is added to handle subtraction of a `dpnp` array from a scalar.

### DIFF
--- a/dpnp/backend/include/dpnp_gen_2arg_3type_tbl.hpp
+++ b/dpnp/backend/include/dpnp_gen_2arg_3type_tbl.hpp
@@ -111,8 +111,8 @@
 
 MACRO_2ARG_3TYPES_OP(dpnp_add_c,
                      input1_elem + input2_elem,
-                     sycl::add_sat(x1, x2),
-                     MACRO_UNPACK_TYPES(int, long),
+                     x1 + x2,
+                     MACRO_UNPACK_TYPES(int, long, bool),
                      oneapi::mkl::vm::add,
                      MACRO_UNPACK_TYPES(float, double, std::complex<float>, std::complex<double>))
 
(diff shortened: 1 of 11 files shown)


dpnp/backend/include/dpnp_gen_2arg_3type_tbl.hpp
```diff
@@ -184,9 +184,9 @@ MACRO_2ARG_3TYPES_OP(dpnp_power_c,
 
 MACRO_2ARG_3TYPES_OP(dpnp_subtract_c,
                      input1_elem - input2_elem,
-                     nullptr,
-                     std::false_type,
+                     sycl::sub_sat(x1, x2),
```
oleksandr-pavlyk, line 187: Why use subract with saturation? I think we should rather do `[](auto x1, auto x2) { return x1 - x2; }`.

antonwolfy: right, no reason for saturation here, I'll resolve that

oleksandr-pavlyk (approved):

Issue #1293, opened 2023-02-13
Update all dpnp tests to run on IRIS Xe


vlad-perevezentsev:
Most tests fail when running them on gpu device without double type support (Iris XE).
Need to update all dpnp tests to consider the specifics of this device.

test_absolute.py output example from WSL:
```
=short test summary info =
FAILED tests/test_absolute.py::test_absolute_float[float64] - AssertionError: 
FAILED tests/test_absolute.py::test_absolute_float_3[float64] - AssertionError: 
=2 failed, 2 passed in 0.38s =

E   AssertionError: 
E   Arrays are not equal
E   
E   Mismatched elements: 1 / 9 (11.1%)
E   Max absolute difference: 3.81469727e-07
E   Max relative difference: 4.19197485e-08
E    x: array([[2. , 3. , 9.1],
E          [2. , 5. , 2. ],
E          [1. , 2. , 5. ]])
E    y: array([[2. , 3. , 9.1],
E          [2. , 5. , 2. ],
E          [1. , 2. , 5. ]], dtype=float32)
```

vlad-perevezentsev:
At the moment most of  dpnp tests have been fixed to run on Iris Xe.
The remaining tests will be fixed as soon as all functions from `dpctl` are reused.
You can find all failed tests in `tests/skipped_tests_gpu_no_fp64.tbl`

PR #1294, opened 2023-02-13
Update dpnp tests to run on Iris Xe

vlad-perevezentsev:
This PR solves the issue #1293 

This PR adds a new **skip_or_change_if_dtype_not_supported** function to helper.py and fixes some tests to run them on Iris Xe which does not support double precision floating point type.
The new function is needed to change an unsupported type to a supported type and skip text if necessary.
Also this PR uses **black 22.6.0** for tests.

Needed to create new PRs and fix the code of function directly for the rest of the tests. 


vlad-perevezentsev:
@oleksandr-pavlyk @antonwolfy what do you think about this solution?

oleksandr-pavlyk:
Test should check for the aspect of the device where the array has been allocated, not of the default-selected device.

tests/test_absolute.py
```diff
@@ -1,51 +1,47 @@
...
 import numpy
 
 
-@pytest.mark.parametrize("type",
-                         [numpy.int64],
-                         ids=['int64'])
-def test_abs_int(type):
-    a = numpy.array([1, 0, 2, -3, -1, 2, 21, -9])
+@pytest.mark.parametrize(
+    "dtype", get_all_dtypes(no_bool=True, no_none=True, no_complex=True)
```
antonwolfy, line 10: I guess it might be helpful to check with default dtype also.
```suggestion
    "dtype", get_all_dtypes(no_bool=True, no_complex=True)
```

tests/test_flat.py
```diff
@@ -5,9 +5,7 @@
 import numpy
 
 
-@pytest.mark.parametrize("type",
-                         [numpy.int64],
-                         ids=['int64'])
+@pytest.mark.parametrize("type", [numpy.int64], ids=["int64"])
```
antonwolfy, line 8: Looks `type` is never used in the test

antonwolfy:
@vlad-perevezentsev, do we need still the PR or it can be rejected?

vlad-perevezentsev:
I will close this PR cause all tests here have been updated with #1472 and other PRs.
At the moment we have no problems with running these tests on Iris Xe.

PR #1295, opened 2023-02-13
dpnp.divide() doesn't work properly with a scalar

antonwolfy:
Since the change implemented in #1201 and until now, implicit `operation /` of array with a scalar or explicit one through `dpnp.divide` and `dpnp.true_divide` leads to raising of the not-implemented exception in dpnp. This is because dpnp was falling back on NumPy implementation in that use case, but has to call an internal kernel function from the backend.
The issue was reported in https://github.com/IntelPython/dpnp/issues/1273.

The PR proposes to copy a scalar data from host memory into USM allocated memory, which means to create zero-dimension dpnp array from the scalar. This way the division of array with a scalar will be handled in the same way as division of two arrays, which already works fine in dpnp.

The default type of result array will depend on device capabilities, that is dpnp will return the result of type either `float32` or `complex64` if no `fp64` support provided be the device.

The compiler flag `fno-finite-math-only` is included now to explicitly disable a floating-point optimizations that assume arguments and results are not NaNs or +/-Inf, since it is not the case in `dpnp`.

### DIFF
--- a/dpnp/backend/include/dpnp_gen_2arg_3type_tbl.hpp
+++ b/dpnp/backend/include/dpnp_gen_2arg_3type_tbl.hpp
@@ -112,7 +112,7 @@
 MACRO_2ARG_3TYPES_OP(dpnp_add_c,
                      input1_elem + input2_elem,
                      x1 + x2,
-                     MACRO_UNPACK_TYPES(int, long, bool),
+                     MACRO_UNPACK_TYPES(bool, std::int32_t, std::int64_t),
                      oneapi::mkl::vm::add,
                      MACRO_UNPACK_TYPES(float, double, std::complex<float>, std::complex<double>))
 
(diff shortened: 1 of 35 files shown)


dpnp/backend/include/dpnp_gen_2arg_3type_tbl.hpp
```diff
@@ -111,8 +111,8 @@
 
 MACRO_2ARG_3TYPES_OP(dpnp_add_c,
                      input1_elem + input2_elem,
-                     sycl::add_sat(x1, x2),
-                     MACRO_UNPACK_TYPES(int, long),
+                     x1 + x2,
+                     MACRO_UNPACK_TYPES(int, long, bool),
```
oleksandr-pavlyk, line 115: On Windows `long` and `int` map to the same 32-bit integer type. Perhaps we need `int` and `long long`.

antonwolfy: Thanks! Changed to `std::int32_t` and `std::int64_t`.

oleksandr-pavlyk:
`x = dpnp.arange(30, dtype="i4")` and `x / 3` raises `ValueError: Device Intel(R) Graphics [0x9a49] does not provide native support for double-precision floating point type.` on Iris Xe.

antonwolfy:
> `x = dpnp.arange(30, dtype="i4")` and `x / 3` raises `ValueError: Device Intel(R) Graphics [0x9a49] does not provide native support for double-precision floating point type.` on Iris Xe.

Thanks for the comment. New commit with the fix is pushed.

oleksandr-pavlyk (approved):
LGTM!

Issue #1307, opened 2023-02-17
DPNP is missing aliases for numpy dtypes


samaid:
I do not want to import numpy to use dpnp. My expectation is that my numpy script will work out of the box with import statement change to 

`import dpnp as np`

 DPNP is lacking numpy dtype aliases which leads to errors like this:
` z = np.empty(c.shape, np.csingle)  # 2d array too 
AttributeError: module 'dpnp' has no attribute 'csingle'`

PR #1309, opened 2023-02-18
Add missing dtype aliases

antonwolfy:
The list of missing aliases on numpy types are ported to dpnp:
- cdouble
- complex_
- complexfloating
- cfloat
- csingle
- double
- float_
- floating
- inexact
- int_
- intc
- number
- object_
- signedinteger
- single
- singlecomplex

The flowing type-related functions are added to dpnp:
 - issubdtype
 - issubsctype

The following constants are borrowed from numpy:
 - e
 - euler_gamma
 - Inf
 - inf
 - Infinity
 - infty
 - NAN
 - NaN
 - NINF
 - NZERO
 - pi
 - PINF
 - PZERO

This PR is intended to resolve issue #1307.

### DIFF
--- a/dpnp/dpnp_algo/dpnp_algo.pyx
+++ b/dpnp/dpnp_algo/dpnp_algo.pyx
@@ -233,11 +233,11 @@ cpdef dpnp_queue_is_cpu():
 Internal functions
 """
 cdef DPNPFuncType dpnp_dtype_to_DPNPFuncType(dtype):
-    dt_c = numpy.dtype(dtype).char
-    kind = numpy.dtype(dtype).kind
+    dt_c = dpnp.dtype(dtype).char
+    kind = dpnp.dtype(dtype).kind
     if isinstance(kind, int):
         kind = chr(kind)
-    itemsize = numpy.dtype(dtype).itemsize
+    itemsize = dpnp.dtype(dtype).itemsize
 
     if dt_c == 'd':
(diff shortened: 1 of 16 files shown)


vlad-perevezentsev (approved):
LGTM!

oleksandr-pavlyk:
I would suggest to remove `dpnp.object_` since `dnp` does not support this data type.

oleksandr-pavlyk (changes requested):
I do not understand why we export and `dpnp.longcomplex` (a.k.a `np.complex256`) as we do not export `dpnp.longdouble` (a.k.a. `np.float128`) and we do not support this data type. 

We also do no support `np.void`, so I suggest we remove `dpnp.void` too.

antonwolfy:
> I do not understand why we export and `dpnp.longcomplex` (a.k.a `np.complex256`) as we do not export `dpnp.longdouble` (a.k.a. `np.float128`) and we do not support this data type.
> 
> We also do no support `np.void`, so I suggest we remove `dpnp.void` too.

Ok, agree.
What do you think to add constants like `dpnp.pi`, `dpnp.inf` and other ones from numpy? (they are also available in cupy as I can see)

oleksandr-pavlyk:
`pi`, `e`, `nan`, `inf`, `newaxis` are mandated by array API.

oleksandr-pavlyk (approved):

PR #1323, opened 2023-03-01
dpnp.power() doesn't work properly with a scalar

antonwolfy:
Since the change implemented in #1201 and until now, implicit `operation **` of array with a scalar or explicit one through `dpnp.power` leads to raising of the not-implemented exception in dpnp. This is because dpnp was falling back on NumPy implementation in that use case, but has to call an internal kernel function from the backend.

The PR proposes to copy a scalar data from host memory into USM allocated memory, which means to create zero-dimension dpnp array from the scalar. This way the power of array with a scalar or the power of scalar with an array will be handled in the same way as power with two arrays, which already works fine in dpnp.

### DIFF
--- a/dpnp/dpnp_iface_mathematical.py
+++ b/dpnp/dpnp_iface_mathematical.py
@@ -45,6 +45,7 @@ from dpnp.dpnp_utils import *
 
 import dpnp
 import numpy
+import dpctl.tensor as dpt
 
 
 __all__ = [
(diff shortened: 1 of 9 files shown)


dpnp/dpnp_iface_mathematical.py
```diff
@@ -1325,18 +1326,35 @@ def negative(x1, **kwargs):
...
+    Returns
+    -------
+    y : dpnp.ndarray
+        The bases in `x1` raised to the exponents in `x2`.
+    
     Limitations
     -----------
-    Parameters ``x1`` and ``x2`` are supported as either :obj:`dpnp.ndarray` or scalar.
-    Parameters ``dtype``, ``out`` and ``where`` are supported with their default values.
+    Parameters `x1` and `x2` are supported as either :class:`dpnp.ndarray` or scalar,
```
npolina4, line 1353: Parameters `x1` and `x2` are supported as either :class:`dpnp.ndarray`,  :class:`dpctl.tensor.usm_ndarray`, or scalar,

npolina4 (approved):

oleksandr-pavlyk (approved):
LGTM from skimming through the code.

PR #1324, opened 2023-03-02
Add support of complex types for dpnp.abs()

antonwolfy:
The PR adds a support of missing bool and complex types for `dpnp.abs` and `dpnp.absolute` functions.

In case of complex types, `oneapi::mkl::vm::abs` function to be reused.
In case of bool type, `dpnp` will return a copy of input array, like numpy does.
For integer types, `sycl::abs` with a vector of input data will be called.


oleksandr-pavlyk (approved):
LGTM. I only was able to skim through the code

PR #1340, opened 2023-03-16
Add support of dpnp.extract()

vlad-perevezentsev:
This PR adds support  of `dpnp.extract()` function in dpnp that reuses `dpctl.tensor.extract()` implementation. 

### DIFF
--- a/dpnp/dpnp_iface_indexing.py
+++ b/dpnp/dpnp_iface_indexing.py
@@ -238,17 +238,22 @@ def extract(condition, x):
     Return the elements of an array that satisfy some condition.
     For full documentation refer to :obj:`numpy.extract`.
 
+    Returns
+    -------
+    y : dpnp.ndarray
+        Rank 1 array of values from `x` where `condition` is True.
+
     Limitations
     -----------
     Parameters `condition` and `x` are supported either as
     :class:`dpnp.ndarray` or :class:`dpctl.tensor.usm_ndarray`.
-    Parameter `x` must be the same size and shape as `condition`.
+    Parameter `x` must be the same shape as `condition`.
     Otherwise the function will be executed sequentially on CPU.
     """
 
     check_input_type = lambda x: isinstance(x, (dpnp_array, dpt.usm_ndarray))
     if check_input_type(condition) and check_input_type(x):
-        if condition.size != x.size or condition.shape != x.shape:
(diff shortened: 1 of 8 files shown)


dpnp/dpnp_iface_indexing.py
```diff
@@ -232,6 +233,35 @@ def diagonal(x1, offset=0, axis1=0, axis2=1):
...
+    -----------
+    Parameters `condition` and `x` are supported either as
+    :class:`dpnp.ndarray` or :class:`dpctl.tensor.usm_ndarray`.
+    Parameter `x` must be the same size and shape as `condition`.
+    Otherwise the function will be executed sequentially on CPU.
+    """
+
+    check_input_type = lambda x: isinstance(x, (dpnp_array, dpt.usm_ndarray))
+    if check_input_type(condition) and check_input_type(x):
+        if condition.size != x.size or condition.shape != x.shape:
```
antonwolfy, line 251: Do we really need to check the sizes here? Isn't that enough to check shapes only?

tests/test_indexing.py
```diff
@@ -53,6 +55,18 @@ def test_diagonal(array, offset):
     assert_array_equal(expected, result)
 
 
+@pytest.mark.parametrize("arr_dtype", get_all_dtypes())
+def test_extract_1d(arr_dtype):
+    a = numpy.array([-2, -1, 0, 1, 2, 3], dtype=arr_dtype)
+    ia = dpnp.array(a)
+    cond = numpy.ones(6, dtype='bool')
```
antonwolfy, line 62: Please add a test when `condition` isn't of boolean type

dpnp/dpnp_iface_indexing.py
```diff
@@ -232,6 +233,35 @@ def diagonal(x1, offset=0, axis1=0, axis2=1):
     return call_origin(numpy.diagonal, x1, offset, axis1, axis2)
 
 
+def extract(condition, x):
+    """
+    Return the elements of an array that satisfy some condition.
+    For full documentation refer to :obj:`numpy.extract`.
+
```
antonwolfy, line 240:
```suggestion

    Returns
    -------
    y : dpnp.ndarray
        Rank 1 array of values from `x` where `condition` is True.

```

PR #1341, opened 2023-03-16
Support more use cases for 'out' parameter

antonwolfy:
The PR is about to improve handling of `out` parameter in supported `dpnp` functions.
It's required to consider use cases when
- `out` refers to the same array as any of input ones
- `out` with strides
- `out` where dtype is different from expected one

In both cases extra memory is allocated and it's need to copy the result back from temporary array into `out` array.

### DIFF
--- a/dpnp/dpnp_utils/dpnp_algo_utils.pyx
+++ b/dpnp/dpnp_utils/dpnp_algo_utils.pyx
@@ -35,6 +35,8 @@ import numpy
 
 import dpctl
 import dpctl.utils as dpu
+import dpctl.tensor._copy_utils as dpt_cu
+import dpctl.tensor._tensor_impl as dpt_ti
 
 import dpnp.config as config
 import dpnp.dpnp_container as dpnp_container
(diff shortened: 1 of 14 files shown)


vlad-perevezentsev (approved):
Good job!

dpnp/dpnp_utils/dpnp_algo_utils.pyx
```diff
@@ -740,6 +745,14 @@ cdef class dpnp_descriptor:
     def is_scalar(self):
         return self.dpnp_descriptor_is_scalar
 
+    @property
+    def is_tempored(self):
```
oleksandr-pavlyk, line 749:
```suggestion
    def is_temporary(self):
```

dpnp/dpnp_utils/dpnp_algo_utils.pyx
```diff
@@ -784,6 +797,13 @@ cdef class dpnp_descriptor:
             "expected either dpctl.tensor.usm_ndarray or dpnp.dpnp_array.dpnp_array, got {}"
             "".format(type(self.origin_pyobj)))
 
+    def get_result_desc(self):
+        if self.is_tempored:
```
oleksandr-pavlyk, line 801:
```suggestion
        if self.is_temporary:
```

vlad-perevezentsev (approved):
LGTM!

PR #1348, opened 2023-03-22
dpnp.fmod() doesn't work properly with a scalar

antonwolfy:
The PR is aimed to implement support of `dpnp.power` call where one of arguments is a scalar.

Currently, since #1201, it raises not-implemented exception in dpnp. This is because dpnp was falling back on NumPy implementation in that use case, but has to call an internal kernel function from the backend.

The PR proposes to copy a scalar data from host memory into USM allocated memory, which means to create zero-dimension dpnp array from the scalar. This way the power of array with a scalar or the power of scalar with an array will be handled in the same way as power with two arrays, which already works fine in dpnp.


npolina4 (approved):

antonwolfy:
#1883 covers that PR

PR #1359, opened 2023-04-03
Implement 'nbytes' property of dpnp array

### LINKED ISSUE #1354: Dpnp array does not support nbytes attribute
`numpy.ndarray` has an attribute `nbytes`, but dpnp does not. Not sure if it is expected to have, just bringing it up from here: https://github.com/IntelPython/dpbench/pull/185#discussion_r1152212723

antonwolfy:
The PR is about to close #1354.

It introduces `nbytes` as a property of dpnp array.

### DIFF
--- a/dpnp/dpnp_array.py
+++ b/dpnp/dpnp_array.py
@@ -29,6 +29,23 @@ import numpy
 
 import dpnp
 
+
+def _get_unwrapped_index_key(key):
+    """
+    Return a key where each nested instance of DPNP array is unwrapped into USM ndarray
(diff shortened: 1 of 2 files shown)


dpnp/dpnp_array.py
```diff
@@ -763,6 +763,8 @@ def item(self, id=None):
     @property
     def itemsize(self):
         """
+        Length of one array element in bytes.
```
oleksandr-pavlyk, line 766:
```suggestion
        Size of one array element in bytes.
```

oleksandr-pavlyk (approved):

PR #1362, opened 2023-04-03
Implement support of tuple key in __getitem__ and __setitem__

### LINKED ISSUE #1361: Improper handling of key as a tuple in dpnp_array.__getitem__()
The below code caused a fault in dpnp:

```
# import dpnp

# x = dpnp.ones((4, 5))
# mask = dpnp.full(x.shape[1], True)
# x[:, mask]
Cell In[4], line 1
----> 1 x[:, mask]

File /localdisk/work/antonvol/code/dpnp_dev/dpnp/dpnp/dpnp_array.py:182, in dpnp_array.__getitem__(self, key)
    179 if isinstance(key, dpnp_array):
    180     key = key.get_array()
--> 182 item = self._array_obj.__getitem__(key)
    183 if not isinstance(item, dpt.usm_ndarray):
    184     raise RuntimeError(
    185         "Expected dpctl.tensor.usm_ndarray, got {}"
    186         "".format(type(item)))

File dpctl/tensor/_usmarray.pyx:716, in dpctl.tensor._usmarray.usm_ndarray.__getitem__()

File dpctl/tensor/_slicing.pxi:171, in dpctl.tensor._usmarray._basic_slice_meta()

File dpctl/tensor/_slicing.pxi:63, in dpctl.tensor._usmarray._is_integral()

File /localdisk/work/antonvol/code/dpnp_dev/dpnp/dpnp/dpnp_array.py:217, in dpnp_array.__index__(self)
    216 def __index__(self):
--> 217     return self._array_obj.__index__()

File dpctl/tensor/_usmarray.pyx:888, in dpctl.tensor._usmarray.usm_ndarray.__index__()

IndexError: only integer arrays are valid indices

```

It looks dpnp wrongly handles indexing use cases when key is a tuple.

antonwolfy:
The PR is about to close #1361.

It introduced support of key as a tuple in `dpnp.__getitem__()` and `dpnp.__setitem__()` functions.

Previously dpnp was falling back on dpctl call in case of tuple as a key, but it is not always correct.
It requires to unwrap dpnp array into USM ndarray for each dpnp array in tuple key before invoking dpctl functions.


oleksandr-pavlyk (approved):

antonwolfy:
> By the way, using `test_equal` may not be the most efficient way of testing array equality:
> 
> ```
> In [13]: import dpnp, numpy as np
> 
> In [14]: x = dpnp.arange(6)
> 
> In [15]: np.asarray(x)
> Out[15]:
> array([array(0), array(1), array(2), array(3), array(4), array(5)],
>       dtype=object)
> 
> In [16]: x = dpnp.arange(6).reshape((2,3))
> 
> In [17]: np.asarray(x)
> Out[17]:
> array([[array(0), array(1), array(2)],
>        [array(3), array(4), array(5)]], dtype=object)
> ```
> 
> I can not quite put my finger on what property of the container makes NumPy behave like this.
> 
> I have tried to create a mock indexable array, but it behaves different:
> 
> ```
> 
> In [24]: class Indexable:
>     ...:     def __init__(self, n):
>     ...:         self.data_ = list(range(n))
>     ...:     def __getitem__(self, key):
>     ...:         return self.data[key]
>     ...:     def __setitem__(self, key, val):
>     ...:         self.data_[key] = val
>     ...:     @property
>     ...:     def shape(self):
>     ...:         return (len(self.data_),)
>     ...:     @property
>     ...:     def dtype(self):
>     ...:         return np.int_
>     ...:
> 
> In [25]: np.asarray(Indexable(6))
> Out[25]: array(<__main__.Indexable object at 0x7f5d742c7790>, dtype=object)
> ```

There will be `dpnp.dpnp_utils.convert_item` function called to convert to numpy array, before invoking `numpy.testing.assert_equal`. Thus `dtype` will be correct one.

PR #1392, opened 2023-05-04
Not use DPC++ 2023.2.0 for conda build

vlad-perevezentsev:
Temporarily do not use DPC++ 2023.2.0 for conda build until an issue with it is solved.

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -18,7 +18,7 @@ requirements:
       - wheel
     build:
       - {{ compiler('cxx') }}
-      - {{ compiler('dpcpp') }}  >=2023.0  # [not osx]
+      - {{ compiler('dpcpp') }}  =2023.1.0  # [not osx]
     run:
       - python
       - dpctl >=0.14
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -19,6 +19,7 @@ requirements:
     build:
       - {{ compiler('cxx') }}
       - {{ compiler('dpcpp') }}  =2023.1.0  # [not osx]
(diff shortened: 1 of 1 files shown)


conda-recipe/meta.yaml
```diff
@@ -18,7 +18,8 @@ requirements:
       - wheel
     build:
       - {{ compiler('cxx') }}
-      - {{ compiler('dpcpp') }}  >=2023.0  # [not osx]
+      - {{ compiler('dpcpp') }}  =2023.1.0  # [not osx]
```
xaleryb, line 21: What is the purpose of this hard pinning?

antonwolfy: Unblock internal CI loop until compiler issue is resolved. It's a kind of possible temporary solution, instead of muting all tests which are crashing now.

conda-recipe/meta.yaml
```diff
@@ -18,7 +18,8 @@ requirements:
...
+      - sysroot_linux-64 >=2.17  # [linux]
```
ekomarova, line 22: Is this problem still not fixed in the new compiler build? Is it possible to test this?

antonwolfy: If we are pinning compiler to 2023.1, we have to pin `sysroot_linux-64` also, otherwise `2.12` will be used for building.
The PR is about to temporary hide test crashes with messages like:
```
Abort was called at 192 line in file:
/opt/src/l0_gpu_driver/shared/source/kernel/kernel_arg_descriptor.h
```
when the latest 2023.2 compiler is used, until the issue is resolve, and so unblock the internal CI.
The idea is to build with 2023.1 but run with 2023.2 runtime.

ekomarova (approved):

PR #1431, opened 2023-06-13
Implementation of dpnp.mean()

vlad-perevezentsev:
This PR implements `mean` function for `dpnp` using `dpnp.sum` and removes the old implantation of `mean` function

- [X] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to issue with a reproducer?
- [X] Have you tested your changes locally for CPU and GPU devices?
- [X] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?

PR #1434, opened 2023-06-14
Add inplace support of divide

antonwolfy:
The PR implements a workaround solution until in-place support of `divide` is provided by dpctl.
And it adds an exception when subtracting two boolean arrays (later in might be move to dpctl).

Additionally the changes includes unmuting of dpnp tests where in-place operations were used before.


vlad-perevezentsev (approved):
It works for #1431 
Thanks @antonwolfy

PR #1450, opened 2023-06-21
Add clang format to pre-commit config

antonwolfy:
The PR proposes to extend pre-commit configuration with `clang` tool.

There is a hard dependency to run the check with `clang-format-12`, since it is the latest version available for Ubuntu 20.04.
Need to note that different versions of `clang-format` might be incompatible with each other and so format C++ code differently.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,71 +1,40 @@
 BasedOnStyle: LLVM
-IndentWidth: 4
-UseTab: Never
-Language: Cpp
-Standard: Cpp11
 
 AccessModifierOffset: -4
 
-AlignConsecutiveDeclarations: false
-AlignConsecutiveAssignments: false
 AlignConsecutiveMacros: true
-AlignTrailingComments: true
+AlignConsecutiveAssignments: true
 AlignEscapedNewlines: Right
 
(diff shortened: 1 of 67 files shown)


.clang-format
```diff
@@ -1,71 +1,43 @@
...
-Standard: Cpp11
 
 AccessModifierOffset: -4
 
-AlignConsecutiveDeclarations: false
-AlignConsecutiveAssignments: false
 AlignConsecutiveMacros: true
-AlignTrailingComments: true
+AlignConsecutiveAssignments: true
+AlignConsecutiveBitFields: true
```
vlad-perevezentsev, line 7:
```suggestion
AlignConsecutiveAssignments: false
AlignConsecutiveBitFields: false
```

it's a matter of opinion but I like the option without alignments

antonwolfy: I would propose to keep `AlignConsecutiveBitFields: true`, since it make the code more readable and structed, but I will fix `AlignConsecutiveAssignments`.

vlad-perevezentsev (approved):
LGTM
Thank @antonwolfy

PR #1463, opened 2023-07-03
dpnp.power doesn`t work properly on Iris Xe

vlad-perevezentsev:
When `dpnp.power` is called with integer and boolean types it uses the `std::pow` function which casts the passed arguments to double so `RuntimeError` error is raised 
```
import dpnp

dpnp.array([4]) ** 2
RuntimeError: Required aspect fp64 is not supported on the device
```

This PR adds a new implementation of template function `ipow` which is used for integer and boolean types of input arguments.


antonwolfy (dismissed):
LGTM! thank you @vlad-perevezentsev !

antonwolfy:
@vlad-perevezentsev , can we close the PR, since all the required changes are going to be implemented by #1463 1476?

PR #1472, opened 2023-07-07
Update some of the dpnp tests to run on Iris Xe

vlad-perevezentsev:
This PR solves the issue #1293 

This PR fixes some tests to run them on Iris Xe which does not support double precision floating point type.
Updated tests: `test_amin_amax.py`, `test_arraycreation.py`, `test_dparray`, `test_fft`,  `test_histograms` , `test_indexing` , `test_sort` and `test_special`


tests/test_arraycreation.py
```diff
@@ -117,6 +121,7 @@ def test_eye(N, M, k, dtype, order):
 @pytest.mark.usefixtures("allow_fall_back_on_numpy")
 @pytest.mark.parametrize("dtype", get_all_dtypes(no_float16=False))
 def test_frombuffer(dtype):
+    skip_dtype_not_supported(dtype)
```
AlexanderKalistratov, line 124: This is probably a bit late, but since you already has `get_all_dtypes` function you could consider and option to mark certain dtypes with `skip` flag inside it using `pytest.param`.

https://docs.pytest.org/en/7.3.x/how-to/parametrize.html

```python
# content of test_expectation.py
import pytest


@pytest.mark.parametrize(
    "test_input,expected",
    [("3+5", 8), ("2+4", 6), pytest.param("6*9", 42, marks=pytest.mark.xfail)],
)
def test_eval(test_input, expected):
    assert eval(test_input) == expected
```

If it is not already done, you probably also want API like this:

```python
@pytest.mark.parametrize("dtype", get_all_dtypes(['float', 'double', 'int32']))
```

I.e. make `get_all_dtypes` to accept list of types and mark unsupported with `skip` flag.

tests/helper.py
```diff
@@ -1,6 +1,7 @@
 from sys import platform
 
 import dpctl
+import pytest
```
antonwolfy, line 4: looks like unnecessary import

tests/test_amin_amax.py
```diff
@@ -79,14 +64,14 @@ def _get_min_max_input(type, shape):
 @pytest.mark.parametrize(
     "shape", [(4,), (2, 3), (4, 5, 6)], ids=["(4,)", "(2,3)", "(4,5,6)"]
 )
-def test_amax(type, shape):
+def test_amax_diff_shape(type, shape):
     a = _get_min_max_input(type, shape)
 
     ia = dpnp.array(a)
```
antonwolfy, line 70: use `get_all_dtypes()` for parametrizing `type`, otherwise dpnp will test twice with "float32" on Iris Xe

tests/test_amin_amax.py
```diff
@@ -102,15 +87,15 @@ def test_amax(type, shape):
 @pytest.mark.parametrize(
     "shape", [(4,), (2, 3), (4, 5, 6)], ids=["(4,)", "(2,3)", "(4,5,6)"]
 )
-def test_amin(type, shape):
+def test_amin_diff_shape(type, shape):
```
antonwolfy, line 90: the same comment about `type`

tests/test_fft.py
```diff
@@ -3,6 +3,12 @@
 
 import dpnp
 
+from .helper import has_support_aspect64
```
antonwolfy, line 6: looks as never used

tests/test_sort.py
```diff
@@ -67,21 +64,15 @@ def test_partition(array, dtype, kth):
...
-    [[1, 2, 3, 4], [-5, -1, 0, 3, 17, 100]],
+    [[1, 2, 3, 4], [-5, -1, 0, 3, 17, 100], [1, 0, 3, 0]],
     ids=[
         "[1, 2, 3, 4]",
-        "[-5, -1, 0, 3, 17, 100]"
-        # '[1, 0, 3, 0]',
-        # '[3, 2, 1, 6]',
-        # '[4, 2, 3, 3, 4, 1]',
-        # '[1, -3, 3, 0, 5, 2, 0, 1, 1, 0, 0, 1]',
-        # '[8, 2, 3, 0, 5, 2, 0, 1, 1, 3, 3, 1, 5, 2, 0, 1]'
```
antonwolfy, line 84: what is about other commented lines? is there any reason to completely remove from the test scope?

tests/test_fft.py
vlad-perevezentsev: This is used on the next line
`
pytestmark = pytest.mark.skipif(
    not has_support_aspect64(), reason="Aborted on Iris Xe"
)
`

antonwolfy: but where is `pytestmark` used?

vlad-perevezentsev: to skip all tests in a module based on some condition

antonwolfy: then you need to add a reference on opened issue to keep track of the problem

vlad-perevezentsev: done

antonwolfy (approved):
Great changes! A lot of the tests are passing on Iris Xe now. Thank you @vlad-perevezentsev

PR #1477, opened 2023-07-13
Removed extra copy for transpose arrays in dot()

antonwolfy:
The PR removes unnecessary copying of input arrays into C-contagious format to call `dpnp.dot()` function.
The change is indented to speed up execution of code like below:
```
import dpnp

a = dpnp.ones((128, 4), dtype=dpnp.float32)[::2]
b = dpnp.ones((1048576, 128), dtype=dpnp.float32)
_ = dpnp.dot(a.T, b.T)
```
This is because `gemm()` function from MKL blas library supports arrays contiguous on inner dimension, i.e. where a stride for at least one dimension should be equal to 1.


vlad-perevezentsev (approved):
LGTM!
thanks @antonwolfy

Issue #1398, opened 2023-05-13
PCA benchmark: dpnp on CPU is 3 times slower than NumPy


antonwolfy:
Results from dpbench of _PCA_ workload (running on a laptop with 11th Gen processor and Iris Xe graphics):

| workload  | numpy | dpnp CPU | dpnp GPU | size |
| :--- | :---: | :---: | :---: | :---: |
|       pca      | 1.2 s  | 3.6 s | 1.2 s | 1048576 |

demonstrates dpnp execution time on CPU which is 3 times greater than one of NumPy.

antonwolfy:
The numbers with _dpnp=0.12.0_:

| workload  | numpy | dpnp CPU | dpnp GPU | size |
| :--- | :---: | :---: | :---: | :---: |
|       pca      | 0.53 s  | 2.01 s | 0.93 s | 1048576 |

The result is faster, but still not in target.

PR #1488, opened 2023-07-19
Add more tests for dpnp.sum and sum_over_axis_0 extension

AlexanderKalistratov:
- [x] Have you provided a meaningful PR description?
- [x] Have you added a test, reproducer or referred to issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [x] Have you checked performance impact of proposed changes?
- [x] If this PR is a work in progress, are you filing the PR as a draft?

Some cleanups, more tests for sum_over_axis_0 extension and dpnp.sum

Issue #1487, opened 2023-07-19
dpnp.sum with "keepsdims" returns incorrect shape in case one or more axis is equal to 0


AlexanderKalistratov:
```python
import dpnp
import numpy

a = dpnp.ones((0, 10))
b = a.sum(keepdims=True)
print("dpnp shape", b.shape)

a_np = numpy.ones((0, 10))
b_np = a_np.sum(keepdims=True)
print("numpy shape", b_np.shape)
```

result:
```
dpnp shape ()
numpy shape (1, 1)
```

antonwolfy:
Resolved by fix in dpctl

PR #1489, opened 2023-07-19
Use specilized kernel for f-arrays and sum by axis=1. Add keepdims support

AlexanderKalistratov:

Tests in https://github.com/IntelPython/dpnp/pull/1488 covers this functionality. Except `keepdims` due to https://github.com/IntelPython/dpnp/issues/1487

Speeds up `cov` with `rowvar=False` parameter.  As it is used in `pca` workload.

```python
import dpnp
import numpy
import time

a = dpnp.ones((1048576, 128))

dpnp.cov(a, rowvar=False, dtype=a.dtype)
start = time.time()
dpnp.cov(a, rowvar=False, dtype=a.dtype)
print(time.time() - start)
```

Before:
```
0.6248629093170166
```

After:
```
0.229935884475708
```

### DIFF
--- a/dpnp/dpnp_iface_mathematical.py
+++ b/dpnp/dpnp_iface_mathematical.py
@@ -1828,12 +1828,18 @@ def sum(
     elif where is not True:
         pass
     else:
-        if axis == (0,) and len(x.shape) == 2 and not keepdims:
+        if len(x.shape) == 2 and (
+            (axis == (0,) and x.flags.c_contiguous)
+            or (axis == (1,) and x.flags.f_contiguous)
+        ):
             from dpctl.tensor._reduction import _default_reduction_dtype
 
             from dpnp.backend.extensions.sycl_ext import _sycl_ext_impl
 
-            input = dpnp.get_usm_ndarray(x)
+            input = x
+            if axis == (1,):
+                input = input.T
+            input = dpnp.get_usm_ndarray(input)
 
             queue = input.sycl_queue
             out_dtype = (
@@ -1850,7 +1856,14 @@ def sum(
 
             if sum:
                 sum(input, output, []).wait()
-                return dpnp_array._create_from_usm_ndarray(output)
+                result = dpnp_array._create_from_usm_ndarray(output)
+
+                if keepdims:
+                    result = result.reshape((1,) + output.shape)
+                    if axis == (1,):
+                        result = result.T
+
+                return result
 
         y = dpt.sum(
             dpnp.get_usm_ndarray(x), axis=axis, dtype=dtype, keepdims=keepdims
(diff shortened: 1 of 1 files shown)


antonwolfy:
I've checked the _PCA_ workload on the laptop:

|   | numpy | dpnp CPU | dpnp GPU OCL | dpnp GPU L0 | size |
| :--- | :---: | :---: | :---: | :---: | :---: |
|       baseline      | 0.61 s  | 0.75 s | 0.71 s | 0.68 s | 1048576 |
|       PR #1489    | 0.6 s  | 0.57 s | 0.3 s | 0.29 s | 1048576 |

Both CPU and GPU are now in targets for the workload, awesome.
I believe it closes #1398.

dpnp/dpnp_iface_mathematical.py
```diff
@@ -1850,7 +1856,14 @@ def sum(
 
             if sum:
                 sum(input, output, []).wait()
-                return dpnp_array._create_from_usm_ndarray(output)
+                result = dpnp_array._create_from_usm_ndarray(output)
+
+                if keepdims:
+                    result = result.reshape((1,) + output.shape)
+                    if axis == (1,):
+                        result = result.T
```
antonwolfy, line 1864:
```suggestion
                    res_sh = (1,) + output.shape if axis == (0,) else output.shape +(1,)
                    return result.reshape(res_sh)
```

antonwolfy, line 1864:
```suggestion
                    res_sh = (1,) + output.shape if axis == (0,) else output.shape +(1,)
                    result = result.reshape(res_sh)
```

antonwolfy: Yesp quite mess. Probably it might be better formatted, like explicitly:
```python
if axis == (0,):
    res_sh = (1,) + output.shape
else:
    res_sh = output.shape + (1,)
```
I just proposed to get rid of excess call of _dpnp.ndarray.T_ method.

AlexanderKalistratov: `dpnp.ndarray.T` method doesn't do anything anything bad. It returns new view with changed strides. Since second dimension is 1 the view keeps to be contiguous.
I'm preferring `dpnp.ndarray.T` method because it is less verbose than `if-else`.

But if you don't like `dpnp.ndarray.T` method I can modify it to `if-else`

antonwolfy: Yes, `dpnp.ndarray.T` methods is quite light, but still produces a number of consequence calls, like `dpnp.ndarray.get_array()`, `dpctl.tensor.usm_ndarray.T` and `dpnp.ndarray._create_from_usm_ndarray` with unpacking and packing dpnp array back from dpctl `usm_ndarray`.

While from another way, it just needs to put (1,) at right place within `if-else`. Thus I'd say that `if-else` looks better from my view.

AlexanderKalistratov: Done

antonwolfy (approved):

PR #1433, opened 2023-06-14
Workaround to Klocwork

antonwolfy:
The PR implements a workaround to Klocwork and enables cmake verbose mode during conda build.

- [x] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?

PR #1493, opened 2023-07-21
Removed w/a to Klocwork

antonwolfy:
The PR removes workaround to Klocwork implemented in #1433.
Due to transition to Coverity tool, the w/a is no longer necessary.

### DIFF
--- a/conda-recipe/build.sh
+++ b/conda-recipe/build.sh
@@ -1,11 +1,5 @@
 #!/bin/bash
 
-# Workaround to Klocwork overwriting LD_LIBRARY_PATH that was modified
-# by DPC++ compiler conda packages. Will need to be added to DPC++ compiler
-# activation scripts.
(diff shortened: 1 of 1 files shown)


oleksandr-pavlyk (approved):

oleksandr-pavlyk:
Note that `LD_LIBRARY_PATH` currently still needs to be modified: See https://github.com/IntelPython/dpctl/pull/1282

It might make sense to add this change in a separate PR to dpnp'd `build.sh`.

antonwolfy:
> Note that `LD_LIBRARY_PATH` currently still needs to be modified: See [IntelPython/dpctl#1282](https://github.com/IntelPython/dpctl/pull/1282)
> 
> It might make sense to add this change in a separate PR to dpnp'd `build.sh`.

Thanks you for the note. I agree that adding the change in a separate PR with python 3.11 support looks more clear.

vlad-perevezentsev (approved):

PR #1494, opened 2023-07-24
Update test_statistic.py to run on Iris Xe

vlad-perevezentsev:
This PR solves updates tests and backend of some `statistic functions` to run on Iris Xe which does not support double precision floating point type.
It solves part of #1293

Also it moves `get_ret_type_and_func` function to `dpnp_algo_utils` file for common use in cython files 

### DIFF
--- a/dpnp/backend/kernels/dpnp_krnl_statistics.cpp
+++ b/dpnp/backend/kernels/dpnp_krnl_statistics.cpp
@@ -1330,9 +1330,21 @@ void func_map_init_statistics(func_map_t &fmap)
         eft_DBL, (void *)dpnp_median_default_c<double, double>};
 
     fmap[DPNPFuncName::DPNP_FN_MEDIAN_EXT][eft_INT][eft_INT] = {
-        eft_DBL, (void *)dpnp_median_ext_c<int32_t, double>};
+        get_default_floating_type<>(),
+        (void *)dpnp_median_ext_c<
+            int32_t, func_type_map_t::find_type<get_default_floating_type<>()>>,
+        get_default_floating_type<std::false_type>(),
+        (void *)dpnp_median_ext_c<
+            int32_t, func_type_map_t::find_type<
+                         get_default_floating_type<std::false_type>()>>};
     fmap[DPNPFuncName::DPNP_FN_MEDIAN_EXT][eft_LNG][eft_LNG] = {
-        eft_DBL, (void *)dpnp_median_ext_c<int64_t, double>};
+        get_default_floating_type<>(),
+        (void *)dpnp_median_ext_c<
+            int64_t, func_type_map_t::find_type<get_default_floating_type<>()>>,
+        get_default_floating_type<std::false_type>(),
+        (void *)dpnp_median_ext_c<
+            int64_t, func_type_map_t::find_type<
+                         get_default_floating_type<std::false_type>()>>};
     fmap[DPNPFuncName::DPNP_FN_MEDIAN_EXT][eft_FLT][eft_FLT] = {
         eft_FLT, (void *)dpnp_median_ext_c<float, float>};
     fmap[DPNPFuncName::DPNP_FN_MEDIAN_EXT][eft_DBL][eft_DBL] = {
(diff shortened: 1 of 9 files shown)


dpnp/dpnp_utils/dpnp_utils_statistics.py
```diff
@@ -69,6 +74,10 @@ def _get_2dmin_array(x, dtype):
             x = x.T
 
         if x.dtype != dtype:
+            fp64 = x.sycl_device.has_aspect_fp64
+            # TODO: remove when dpctl.result_type() is fixed
+            if not fp64:
+                dtype = dtype_map[dtype.type]
```
antonwolfy, line 80: I'd propose to move the w/a to `if` block below where `dpt.result_type` is called, otherwise you are hiding the issue when incoming `dtype=dpnp.float64` is passed to `dpnp.cov()`.

dpnp/dpnp_utils/dpnp_utils_statistics.py
```diff
@@ -59,6 +59,11 @@ def _get_2dmin_array(x, dtype):
         It casts to another dtype, if the input array differs from requested one.
 
         """
+        # dtype map for devices without double precision type support
+        dtype_map = {
+            dpnp.float64: dpnp.float32,
+            dpnp.complex128: dpnp.complex64,
+        }
```
antonwolfy, line 66: I'd be better to move the map under the place where `TODO: remove when dpctl.result_type() is fixed` is commented in order to keep in mind that this has to be deleted once the w/a is removing

dpnp/backend/kernels/dpnp_krnl_statistics.cpp
```diff
@@ -1330,9 +1330,21 @@ void func_map_init_statistics(func_map_t &fmap)
         eft_DBL, (void *)dpnp_median_default_c<double, double>};
 
     fmap[DPNPFuncName::DPNP_FN_MEDIAN_EXT][eft_INT][eft_INT] = {
-        eft_DBL, (void *)dpnp_median_ext_c<int32_t, double>};
+        get_default_floating_type<>(),
```
antonwolfy, line 1333:
```suggestion
        get_default_floating_type(),
```

tests/test_statistics.py
```diff
@@ -1,37 +1,37 @@
...
 
-    numpy.testing.assert_allclose(dpnp_res, np_res)
+    assert_allclose(dpnp_res, np_res)
 
 
 @pytest.mark.usefixtures("allow_fall_back_on_numpy")
 @pytest.mark.parametrize("axis", [0, 1, -1, 2, -2, (1, 2), (0, -2)])
 def test_max(axis):
-    a = numpy.arange(768, dtype=numpy.float64).reshape((4, 4, 6, 8))
+    dtype = dpnp.default_float_type()
```
antonwolfy, line 27: can we extend the test to all supported dtypes, like `get_all_dtypes(no_none=True, no_bool=True, no_complex=True)`

tests/test_statistics.py
```diff
@@ -71,16 +71,17 @@ def test_max(axis):
     ],
 )
 def test_nanvar(array):
-    a = numpy.array(array)
+    dtype = dpnp.default_float_type()
```
antonwolfy, line 74: the same comment here

dpnp/dpnp_utils/dpnp_utils_statistics.py
```diff
@@ -81,6 +80,13 @@ def _get_2dmin_array(x, dtype):
         if y is not None:
             dtypes.append(y.dtype)
         dtype = dpt.result_type(*dtypes)
+        # TODO: remove when dpctl.result_type() is fixed
```
antonwolfy, line 83:
```suggestion
        dtype = dpnp.result_type(*dtypes)
        # TODO: remove when dpctl.result_type() is returned dtype based on fp64
```

antonwolfy (approved):
LGTM! thank you @vlad-perevezentsev

dpnp/dpnp_utils/dpnp_utils_statistics.py
vlad-perevezentsev: It will be fixed in the next PR.
thanks

PR #1498, opened 2023-07-25
Update tests to run on Iris Xe

npolina4:
Update random, logic, manipulation, and mixins tests to run on Iris Xe.

### DIFF
--- a/tests/test_mixins.py
+++ b/tests/test_mixins.py
@@ -40,5 +40,6 @@ class TestMatMul(unittest.TestCase):
 
         return out
 
+
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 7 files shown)


tests/test_mixins.py
```diff
@@ -4,72 +4,74 @@
...
-            (size, size)
-        )
-        expected = numpy.matmul(array_1, array_2)
-        # print(expected)
-
-        # passed
-        numpy.testing.assert_array_equal(expected, result)
-        # still failed
-        # self.assertEqual(expected, result)
+        for dtype in get_float_dtypes():
```
antonwolfy, line 15: Wouldn't it be better to use pytest parametrize instead of nested loop? Or is there any reason why it can't be implemented here?

antonwolfy (approved):
LGTM! thank you @npolina4

PR #1500, opened 2023-07-25
Use get_ret_type_and_func in call_fptr functions 

vlad-perevezentsev:
This PR changes signature of `get_ret_type_and_func` function. The updated function takes `sycl_device` argument instead of `x1_obj` and checks null pointer to kernel with no `fp64` support. 

Also the PR suggests using updated `get_ret_type_and_func` function in `call_fptr_1in_1out_strides`, `call_fptr_2in_1out` and `call_fptr_2in_1out_strides` functions for further test updates to run on Iris Xe. 

### DIFF
--- a/dpnp/dpnp_utils/dpnp_algo_utils.pyx
+++ b/dpnp/dpnp_utils/dpnp_algo_utils.pyx
@@ -663,15 +663,16 @@ cdef tuple get_common_usm_allocation(dpnp_descriptor x1, dpnp_descriptor x2):
     return (common_sycl_queue.sycl_device, common_usm_type, common_sycl_queue)
 
 
-cdef (DPNPFuncType, void *) get_ret_type_and_func(device, DPNPFuncData kernel_data):
+cdef (DPNPFuncType, void *) get_ret_type_and_func(DPNPFuncData kernel_data,
+                                                  cpp_bool has_aspect_fp64):
     """
     This function is responsible for determining the appropriate return type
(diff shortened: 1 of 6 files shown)


dpnp/dpnp_utils/dpnp_algo_utils.pyx
```diff
@@ -663,13 +663,19 @@ cdef tuple get_common_usm_allocation(dpnp_descriptor x1, dpnp_descriptor x2):
     return (common_sycl_queue.sycl_device, common_usm_type, common_sycl_queue)
 
 
-cdef (DPNPFuncType, void *) get_ret_type_and_func(x1_obj, DPNPFuncData kernel_data):
-    if dpnp.issubdtype(x1_obj.dtype, dpnp.integer) and not x1_obj.sycl_device.has_aspect_fp64:
+cdef (DPNPFuncType, void *) get_ret_type_and_func(device, DPNPFuncData kernel_data):
```
antonwolfy, line 666: Wouldn't it be better to explicitly pass only `has_aspect_fp64`, since no more information from device is required?
```suggestion
cdef (DPNPFuncType, void *) get_ret_type_and_func(DPNPFuncData kernel_data, cpp_bool has_aspect_fp64):
```

vlad-perevezentsev: You are right. It will be better 
thanks

PR #1513, opened 2023-08-07
Resolve `Arguments in wrong order` issue reported by Coverity tool

antonwolfy:
The TR is intended to resolve an issue from Coverity report with "high" severity.

The tool claims that the positions of arguments in the call to `mkl_blas::gemm` do not match the ordering of the parameters: `ldb` is passed to `lda`, `lda` is passed to `ldb`.
Actually, the issue is false positive, since local variables `lda` and `ldb` are assigning to correct values. But the PR proposes to resolve the reported problem to avoid future misleading.

### DIFF
--- a/dpnp/backend/kernels/dpnp_krnl_common.cpp
+++ b/dpnp/backend/kernels/dpnp_krnl_common.cpp
@@ -852,15 +852,15 @@ DPCTLSyclEventRef dpnp_matmul_c(DPCTLSyclQueueRef q_ref,
     {
         // using std::max for these ldx variables is required by math library
         const std::int64_t lda =
-            std::max<size_t>(1UL, size_k); // First dimensions of array_1
+            std::max<size_t>(1UL, size_n); // First dimensions of array_1
         const std::int64_t ldb =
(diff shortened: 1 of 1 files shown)


dpnp/backend/kernels/dpnp_krnl_common.cpp
```diff
@@ -852,15 +852,15 @@ DPCTLSyclEventRef dpnp_matmul_c(DPCTLSyclQueueRef q_ref,
     {
         // using std::max for these ldx variables is required by math library
         const std::int64_t lda =
-            std::max<size_t>(1UL, size_k); // First dimensions of array_1
+            std::max<size_t>(1UL, size_n); // First dimensions of array_1
         const std::int64_t ldb =
-            std::max<size_t>(1UL, size_n); // First dimensions of array_2
+            std::max<size_t>(1UL, size_k); // First dimensions of array_2
```
oleksandr-pavlyk, line 857: Comment have to be changed. Perhaps variable could also be renamed for clarity. Instead of `lda` I would suggest `ld_array_2`, and instead of current `ldb` use `ld_array_1`.

PR #1524, opened 2023-08-17
Update dpnp fft implementations to run on Iris Xe

vlad-perevezentsev:
This PR suggests expanding type mapping for `fft` kernels and calling them when running on Iris Xe without support for double-precision types.
Additionally updates the docstrings and tests for all `fft` functions


dpnp/backend/kernels/dpnp_krnl_fft.cpp
```diff
@@ -476,8 +476,12 @@ DPCTLSyclEventRef dpnp_fft_fft_c(DPCTLSyclQueueRef q_ref,
         else if constexpr (std::is_same<_DataType_input, int32_t>::value ||
                            std::is_same<_DataType_input, int64_t>::value)
         {
-            double *array1_copy = reinterpret_cast<double *>(
-                dpnp_memory_alloc_c(q_ref, input_size * sizeof(double)));
+            using CastType = std::conditional_t<
```
antonwolfy, line 479: Need to add a static assertion that sycl::detail::is_complex<_DataType_output> is true.

dpnp/backend/kernels/dpnp_krnl_fft.cpp
```diff
@@ -476,8 +476,12 @@ DPCTLSyclEventRef dpnp_fft_fft_c(DPCTLSyclQueueRef q_ref,
...
+                std::is_same<_DataType_output, std::complex<double>>::value,
+                double, float>;
```
antonwolfy, line 481: it can be simplified, like:
```suggestion
            using CastType = _DataType_output::value_type;
```

dpnp/fft/dpnp_iface_fft.py
```diff
@@ -86,23 +86,24 @@ def get_validated_norm(norm):
...
-    Parameter ``axis`` is supported with its default value.
-    Parameter ``x1`` supports ``dpnp.int32``, ``dpnp.int64``, ``dpnp.float32``, ``dpnp.float64``,
-    ``dpnp.complex64`` and ``dpnp.complex128`` datatypes only.
-
-    For full documentation refer to :obj:`numpy.fft.fft`.
+    Parameter `x` is supported either as :class:`dpnp.ndarray`.
+    Parameter `axis` is supported with its default value.
+    Only `dpnp.float64`, `dpnp.float32`, `dpnp.int64`, `dpnp.int32`,
+    `dpnp.complex128`, `dpnp.complex64` data types are supported.
+    Otherwise the function will be executed sequentially on CPU.
```
antonwolfy, line 101: I guess there will an exception rather than fallback on numpy in case of boolean type of input array.
I think we need to clear state that.

antonwolfy: Probably it would be better to add a type check before calling backend function and to raise an exception for boolean type rather than to have RuntimeError exception.

antonwolfy (approved):
Thank you @vlad-perevezentsev LGTM!

PR #1526, opened 2023-08-17
Align default order value with numpy in asarray-like functions

antonwolfy:
The PR proposes to align default `order` value with numpy implementation in `dpnp.array`, `dpnp.asarray` and `dpnp.asanyarray` functions.

An array creation function `dpnp.ascontiguousarray` is extended with missing arguments and aligned with numpy.
A missing function `dpnp.asfortranarray` is added to dpnp.
The device keyword arguments are implemented for both functions.

The parameter `like` is described as currently unsupported and will cause `ValueError` exception if passed in each updated function. The same is about `subok` and `ndim` parameters for `dpnp.array` function.
No fallback on numpy will be invoked anymore for the functions.

The test scope is extended with new scenarios and previously muted tests are enabled.


vlad-perevezentsev (approved):
Nice work on this PR!
Thank you @antonwolfy

oleksandr-pavlyk (approved):
LGTM! Thank you @antonwolfy

PR #1536, opened 2023-08-24
Implement dpnp.allclose() for a device without fp64 aspect

antonwolfy:
The PR proposes to provide support of `dpnp.allclose` function for a device without fp64 aspect.
Additionally, use cases with `nan` and `inf` values in input arrays are handling properly now.


dpnp/backend/kernels/dpnp_krnl_logic.cpp
```diff
@@ -127,8 +127,79 @@ DPCTLSyclEventRef (*dpnp_all_ext_c)(DPCTLSyclQueueRef,
...
+            {
+                if (std::isinf(array1[i]) || std::isinf(array2[i])) {
+                    partial &= (array1[i] == array2[i]);
+                    continue;
+                }
+            }
+
+            // casting integeral to floating type to avoid bad behavior
+            // on abs(MIN_INT), which leads to undefined result
+            using _Arr2Type = std::conditional_t<std::is_integral_v<_DataType2>,
```
vlad-perevezentsev, line 180: It would be better if `_Arr2Type` is declared outside the loop to avoid unnecessary computations

antonwolfy: but it's done at compile stage, what kind of computation do you mean?

dpnp/dpnp_iface.py
```diff
@@ -453,6 +454,28 @@ def get_usm_ndarray_or_scalar(a):
...
 
+def is_supported_array_or_scalar(a):
+    """
+    Return ``True`` if `a` is a scalar or an array of either
+    :class:`dpnp.ndarray` or :class:`dpctl.tensor.usm_ndarray` type,
+    ``False`` otherwise.
+
+    Parameters
+    ----------
+    a : array
```
vlad-perevezentsev, line 465: `a` may be a scalar

dpnp/backend/kernels/dpnp_krnl_logic.cpp
vlad-perevezentsev: Yes, you are right.
My remark was more about improving the readability of code but your changes also look good.

vlad-perevezentsev (approved):
great job!
thank you @antonwolfy

PR #1616, opened 2023-11-03
Add new blas extension and update dpnp.matmul func

vlad-perevezentsev:
This PR updates `dpnp.matmul` implementation with `gemm` function call from `oneapi::mkl::blas` added to `dpnp.backend.extensions.blas`.
In addition, `gemm_batch` function is also added which is the batched versions of `gemm`, performing multiple `gemm` operations in a single call.
The implementation is written as a pybind11 extension above required BLAS functions.


npolina4 (changes requested):
The result does not match with numpy
```python
import numpy, dpnp

def init():
    N = 5
    A = numpy.empty((N, N), dtype=numpy.float32)
    for i in range(N):
        A[i, : i + 1] = numpy.fromfunction(lambda j: (-j % N) / N + 1, (i + 1,))
        A[i, i + 1 :] = 0.0
        A[i, i + 1 :] = 0.0
    return A

def kernel(A):
    for i in range(A.shape[0]):
        for j in range(i):
            A[i, j] -= A[i, :j] @ A[:j, j]

a = init()
b = dpnp.asarray(a)

kernel(a)
kernel(b)
print(a - b.asnumpy())
```
```
[[0.  0.  0.  0.  0. ]
 [0.  0.  0.  0.  0. ]
 [0.  0.  0.  0.  0. ]
 [0.  0.  0.  0.  0. ]
 [0.  0.  0.  1.6 0. ]]
```

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -274,58 +290,328 @@ def matmul(x1, x2, out=None, **kwargs):
     >>> c = np.ones([9, 5, 4, 3])
     >>> np.matmul(a, c).shape
```
vlad-perevezentsev, line 291: This behavior is demonstrable in the third example (`>>> a = np.arange(2 * 2 * 4).reshape((2, 2, 4))`) so it can be removed

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -274,58 +290,328 @@ def matmul(x1, x2, out=None, **kwargs):
...
     (9, 5, 7, 3)
+
     >>> a = np.array([[1, 0], [0, 1]])
     >>> b = np.array([[4, 1], [2, 2]])
     >>> np.matmul(a, b)
     array([[4, 1],
            [2, 2]])
```
vlad-perevezentsev, line 299: Please add descriptions for each example
```suggestion
For 2-D arrays it is the matrix product:

```

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -274,58 +290,328 @@ def matmul(x1, x2, out=None, **kwargs):
...
+        stridec,
+        True,  # transa
+        True,  # transb
+        [copy_ev_x1, copy_ev_x2],
+    )
+
+    return ht_blas_ev, ht_copy_ev_x1, ht_copy_ev_x2, res
+
+
+def _gemm_res_dtype(*arrays, casting):
```
vlad-perevezentsev, line 592: Please add a description for this function

tests/third_party/cupy/math_tests/test_matmul.py
```diff
@@ -25,39 +25,38 @@
...
+                ((6, 5, 3, 0), (0,)),
+                ((2,), (6, 5, 2, 4)),
+                ((0,), (6, 5, 0, 4)),
                 ((1, 3, 3), (10, 1, 3, 1)),
             ],
         }
     )
 )
-@pytest.mark.usefixtures("allow_fall_back_on_numpy")
 @testing.gpu
```
vlad-perevezentsev, line 60: This decorator can be removed for all tests

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +285,364 @@ def matmul(x1, x2, out=None, **kwargs):
...
+            raise ValueError(
+                "matmul: Input operand does not have enough dimensions"
+            )
+
+        exec_q = dpctl.utils.get_execution_queue((x1.sycl_queue, x2.sycl_queue))
+        if exec_q is None:
+            raise ValueError(
+                "Execution placement can not be unambiguously inferred "
+                "from input arguments."
+            )
```
vlad-perevezentsev, line 353:
```suggestion
        res_usm_type, exec_q = get_usm_allocations([x1, x2])
```

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +285,364 @@ def matmul(x1, x2, out=None, **kwargs):
...
-                    """
-                    Floating point types are handled via original math library better than SYCL math library
-                    """
-                    cost_size = 262144  # 2D array shape(512, 512)
-
-                if (array1_size > cost_size) and (array2_size > cost_size):
-                    return dpnp_matmul(x1_desc, x2_desc, out)
+            if x1.size == 0 or x2.size == 0:
+                result = dpnp.zeros(
+                    res_shape, dtype=gemm_dtype, sycl_queue=exec_q
```
vlad-perevezentsev, line 460:
```suggestion
                    res_shape, dtype=gemm_dtype, sycl_queue=exec_q, usm_type=res_usm_type
```

Additionally please add a test for this case in `test_usm_type.py`

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +285,364 @@ def matmul(x1, x2, out=None, **kwargs):
...
+
+        if not x2.flags["C_CONTIGUOUS"] and not x2.flags["F_CONTIGUOUS"]:
+            v = dpnp.empty_like(x2, order="C")
+            ht_copy_ev_x2, _ = ti._copy_usm_ndarray_into_usm_ndarray(
+                src=dpnp.get_usm_ndarray(x2),
+                dst=v.get_array(),
+                sycl_queue=x2.sycl_queue,
+            )
+            x2 = v
+            ht_copy_ev_x2.wait()
```
vlad-perevezentsev, line 374: [Tag: copy]
You do not have to do copying at the beginning of  `matmul` function
It would be better to do it before calculating the result
```suggestion
```

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +285,364 @@ def matmul(x1, x2, out=None, **kwargs):
...
+        # Determine the appropriate data types
+        if dtype is not None:
+            res_dtype = dtype
+            gemm_dtype, _ = _gemm_res_dtype(x1, x2, casting=casting)
+        else:
+            gemm_dtype, res_dtype = _gemm_res_dtype(x1, x2, casting=casting)
+        if x1.dtype != gemm_dtype:
+            x1 = dpnp.astype(x1, gemm_dtype, casting=casting)
+        if x2.dtype != gemm_dtype:
+            x2 = dpnp.astype(x2, gemm_dtype, casting=casting)
```
vlad-perevezentsev, line 403: [Tag: copy]
It looks like extra copying
You can set `dtype` during copying before calculating the result
```suggestion
```

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +285,364 @@ def matmul(x1, x2, out=None, **kwargs):
...
+                        if not x2_is_2D:
+                            x2 = dpnp.repeat(x2, x1_shape[i], axis=i)
+                    else:
+                        raise ValueError(
+                            "operands could not be broadcast together with remapped shapes."
+                        )
+            x1_shape = x1.shape
+            x2_shape = x2.shape
+            res_shape = tuple(tmp_shape) + (x1_shape[-2], x2_shape[-1])
+
```
vlad-perevezentsev, line 446: [Tag: copy]
This will avoid multiple copies and improve performance
Also created a list of events to achieve parallel copying for `x2` array with copying `x1` array 
```suggestion
        is_x1_c_f_contig = x1.flags.c_contiguous or x1.flags.f_contiguous
        is_x2_c_f_contig = x2.flags.c_contiguous or x2.flags.f_contiguous

        list_of_events = []

        # input arrays should be C_CONTIGUOUS or F_CONTIGUOUS
        if not is_x1_c_f_contig or x1.dtype != gemm_dtype:
            x1_copy = dpnp.empty_like(x1, dtype=gemm_dtype, order="C")
            ht_copy_ev_x1, _ = ti._copy_usm_ndarray_into_usm_ndarray(
                src=dpnp.get_usm_ndarray(x1),
                dst=x1_copy.get_array(),
                sycl_queue=x1.sycl_queue,
            )
            x1 = x1_copy
            list_of_events.append(ht_copy_ev_x1)

        if not is_x2_c_f_contig or x2.dtype != gemm_dtype:
            x2_copy = dpnp.empty_like(x2, dtype=gemm_dtype, order="C")
            ht_copy_ev_x2, _ = ti._copy_usm_ndarray_into_usm_ndarray(
                src=dpnp.get_usm_ndarray(x2),
                dst=x2_copy.get_array(),
                sycl_queue=x2.sycl_queue,
            )
            x2 = x2_copy
            list_of_events.append(ht_copy_ev_x2)

        if list_of_events:
            for ev in list_of_events:
                ev.wait()
```

tests/test_mathematical.py
```diff
@@ -2147,3 +2148,237 @@ def test_inplace_remainder(dtype):
...
+        ((4, 3), (6, 7, 3, 5)),
+        ((6, 1, 4, 3), (6, 7, 3, 5)),
+        ((1, 7, 4, 3), (6, 7, 3, 5)),
+        ((7, 4, 3), (6, 7, 3, 5)),
+        ((1, 5, 3, 2), (6, 5, 2, 4)),
+        ((5, 3, 2), (6, 5, 2, 4)),
+        ((1, 3, 3), (10, 1, 3, 1)),
+    ],
+)
+def test_matmul(order_pair, shape_pair):
```
vlad-perevezentsev, line 2202: Please create `TestMatmul` class and incorporate `test_matmul` , `test_matmul_dtype`,  `test_matmul_order` ,`test_matmul_strided`  and `test_matmul_out` functions into it

npolina4 (approved):

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +285,364 @@ def matmul(x1, x2, out=None, **kwargs):
...
+        else:
+            gemm_dtype, res_dtype = _gemm_res_dtype(x1, x2, casting=casting)
+        if x1.dtype != gemm_dtype:
+            x1 = dpnp.astype(x1, gemm_dtype, casting=casting)
+        if x2.dtype != gemm_dtype:
+            x2 = dpnp.astype(x2, gemm_dtype, casting=casting)
+
+        # find the result shape
+        x1_is_2D = False
+        x1_is_2D = False
```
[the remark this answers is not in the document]
vtavana: Actually, I can remove them. I guess I had a different logic previously and had to define them here. But currently, they only needed in the `else` section of `if` condition.

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +285,364 @@ def matmul(x1, x2, out=None, **kwargs):
...
+        if out is None and order not in ["k", "K"]:
+            v = dpnp.empty_like(result, order=order)
+            ht_copy_ev, _ = ti._copy_usm_ndarray_into_usm_ndarray(
+                src=dpnp.get_usm_ndarray(result),
+                dst=v.get_array(),
+                sycl_queue=result.sycl_queue,
+            )
+            ht_copy_ev.wait()
+            result = v
+        return dpnp.get_result_array(result, out, casting=casting)
```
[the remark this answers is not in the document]
vlad-perevezentsev: I did not consider that we support  `order` parameter in the suggestion above.
My idea is to get rid of redundant array copying in your implementation. 
If order is set by default we just return `result`

```suggestion
        if out is not None:
            return dpnp.get_result_array(result, out, casting=casting)

        # If `order` was passed not as default
        # we must copy `result` to match the passed `order`.
        if order not in ["k", "K"]:
            return dpnp.copy(result, order=order)

        return result
```

dpnp/backend/extensions/blas/gemm_batch.cpp
```diff
@@ -0,0 +1,220 @@
...
+               const std::vector<sycl::event> &depends = {})
+{
+    if (!dpctl::utils::queues_are_compatible(
+            exec_q,
+            {matrixA.get_queue(), matrixB.get_queue(), resultC.get_queue()}))
+    {
+        throw std::runtime_error(
+            "USM allocations are not compatible with the execution queue.");
+    }
+
```
vlad-perevezentsev, line 152: Please add an overlap check

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +282,365 @@ def matmul(x1, x2, out=None, **kwargs):
...
+            return dpnp.copy(result, order=order)
+
+        return result
+
+
+def dpnp_matmul_batch(
+    exec_q, x1, x2, res, x1_is_2D, x2_is_2D, ht_copy_ev_x1, ht_copy_ev_x2
+):
+    copy_ev_x1 = dpctl.SyclEvent()
+    if not x1.flags["C_CONTIGUOUS"]:
```
vlad-perevezentsev, line 513: Is it possible to support `_gemm_batch` to work with F-contiguous arrays?

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +282,365 @@ def matmul(x1, x2, out=None, **kwargs):
...
+        v = dpnp.empty_like(x1, order="C")
```
vlad-perevezentsev, line 514:
```suggestion
        x1_copy = dpnp.empty_like(x1, order="C")
```
Same for x2

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +330,325 @@ def matmul(x1, x2, out=None, **kwargs):
...
+        batch_size = iter[0][0]
+        stridea = iter[1][0]
+        strideb = iter[3][0]
+
+    ht_blas_ev, _ = bi._gemm_batch(
+        exec_q,
+        dpnp.get_usm_ndarray(x1),
+        dpnp.get_usm_ndarray(x2),
+        dpnp.get_usm_ndarray(res),
+        m,
```
antonwolfy, line 636: Do we need to have `m, n, k` and `lda, ldb, ldc` as input arguments? Looks we can take it from input arrays already passed to `_gemm_batch`.

tests/test_mathematical.py
```diff
@@ -2242,3 +2243,310 @@ def test_inplace_remainder(dtype):
...
+    def test_invalid_dtype(self, dtype):
+        dpnp_dtype = get_all_dtypes(no_none=True)[-1]
+        a1 = dpnp.arange(5 * 4, dtype=dpnp_dtype).reshape(5, 4)
+        a2 = dpnp.arange(7 * 4, dtype=dpnp_dtype).reshape(4, 7)
+        dp_out = dpnp.empty((5, 7), dtype=dtype)
+
+        with pytest.raises(TypeError):
+            dpnp.matmul(a1, a2, out=dp_out)
+
+    def test_exe_q(self):
```
antonwolfy, line 2527: it is easier to use different SYCL queues than different devices (since we can have a use case where only single device is available)

dpnp/backend/extensions/blas/gemm.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+    const py::ssize_t *a_shape = matrixA.get_shape_raw();
+    const py::ssize_t *b_shape = matrixB.get_shape_raw();
+    const py::ssize_t *res_shape = resultC.get_shape_raw();
+
+    if (a_shape[1] != b_shape[0]) {
+        throw std::runtime_error("The number of columns in A must be equal to "
+                                 "the number of rows in B.");
+    }
+
```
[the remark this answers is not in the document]
vtavana: We create a new empty array for result in Python side. Is it technically possible to encounter a memory overlap?

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -60,6 +61,56 @@
...
+            dpnp.complex64 if gemm_dtype == dpnp.float32 else dpnp.complex128
+        )
+
+    if dpnp.can_cast(res_dtype, gemm_dtype, casting):
+        if res_dtype in [
+            dpnp.float64,
+            dpnp.complex128,
+        ]:  # in case device does not support fp64
+            return gemm_dtype, gemm_dtype
+        elif res_dtype in [
```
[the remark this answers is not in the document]
vtavana: I get the following on Iris Xe gpu which shows device capabilities is not taken into account.
>>> a=dpnp.array([1,2,3], dtype="i4")
>>> b=dpnp.array([1,2,3], dtype="f4")
>>> dpnp.result_type(a,b)
dtype('float64')

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +330,325 @@ def matmul(x1, x2, out=None, **kwargs):
...
-                ):
-                    """
-                    Floating point types are handled via original math library better than SYCL math library
-                    """
-                    cost_size = 262144  # 2D array shape(512, 512)
-
-                if (array1_size > cost_size) and (array2_size > cost_size):
-                    return dpnp_matmul(x1_desc, x2_desc, out)
+            if x1.size == 0 or x2.size == 0:
+                result = dpnp.zeros(
```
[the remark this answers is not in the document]
vtavana: what about using `result.fill(0)`?

dpnp/dpnp_iface_linearalgebra.py
```diff
@@ -269,63 +330,325 @@ def matmul(x1, x2, out=None, **kwargs):
...
+                ht_copy_ev_x1.wait()
+                ht_copy_ev_x2.wait()
+
+        if squeeze_flag:
+            result = dpnp.squeeze(result)
+
+        if gemm_dtype != res_dtype:
+            result = dpnp.astype(result, res_dtype)
+
+        if out is not None:
```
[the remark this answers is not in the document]
vtavana: Since we only perform the calculations for `matmul` in floating point data type using OneMKL. If the input is integer type, the intermediate result would be in floating point and if the output from out keyword is integer we will get an error unless we have change the data type of intermediate result to integer.

dpnp/dpnp_iface_linearalgebra.py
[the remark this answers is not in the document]
antonwolfy: This is because `dpnp.result_type` wasn't properly updated, should be something like:
```python
def result_type(*arrays_and_dtypes):
    usm_arrays_and_dtypes = [
        dpnp.get_usm_ndarray(X) if isinstance(X, (dpnp_array, dpt.usm_ndarray)) else X
        for X in arrays_and_dtypes
    ]
    return dpt.result_type(*usm_arrays_and_dtypes)
```

dpnp/dpnp_iface_linearalgebra.py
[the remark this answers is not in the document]
antonwolfy: but then we should do that only with `copy=False` and for non floating types - to avoid extra copying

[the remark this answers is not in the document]
vtavana: Yes, I have added `copy=False` in recent updates and the `if gemm_dtype != res_dtype` will only be true for for non-floating types.

dpnp/dpnp_utils/dpnp_utils_linearalgebra.py
```diff
@@ -0,0 +1,358 @@
...
+        # was inehrently 2D
+        new_size = max(x1_ndim, x2_ndim)
+        for _ in range(new_size - 2):
+            result = result[dpnp.newaxis, :]
+
+    if gemm_dtype != res_dtype:
+        result = dpnp.astype(result, res_dtype, copy=False)
+    if out is None:
+        # If `order` was not passed as default
+        # we must copy `result` to match the passed `order`.
```
antonwolfy, line 352: I'm a bit confused about the comment. We are passing `copy=False` to `dpnp.array(...)`, doesn't it mean that no copy will be done?

tests/test_manipulation.py
```diff
@@ -72,10 +73,13 @@ def test_repeat(arr):
 
 
 def test_result_type():
```
antonwolfy, line 75: currently failing in the internal CI, since no latest dpctl available there, probably we need to update the test by a separate PR

dpnp/dpnp_utils/dpnp_utils_linearalgebra.py
vtavana: The comment is out of date. I updated it accordingly.

tests/test_manipulation.py
vtavana: For now, I skip the test and put it down to remove skip later.

dpnp/dpnp_utils/dpnp_utils_linearalgebra.py
```diff
@@ -0,0 +1,358 @@
+# *****************************************************************************
+# Copyright (c) 2024, Intel Corporation
```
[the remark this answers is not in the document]
vtavana: I will create a separate PR and update the year for all the files to make all files consistent.

antonwolfy (approved):
Thank you @vtavana !

PR #1637, opened 2023-12-05
Redesign `dpnp.diff` thorough existing calls

antonwolfy:
The PR proposes to implement `dpnp.diff` function through existing dpnp python API. While the previous implementation of `dpnp.diff` was done through cython and has limited support of input arguments.

The PR assumes to get rid of all the limitations and to align with numpy implementation of `diff` function.


vtavana (approved):
Thank you, @antonwolfy!

PR #1644, opened 2023-12-12
Implement a helper alias template for complex types

antonwolfy:
The PR proposes to implement own alias template `is_complex` rather than relying on internal one from SYCL library, since it is from internal SYCL namespace, also it might have a limitation (for instance, latest DPC++ package for 2024.1 breaks dpnp compilation on Windows) and not indented initially to be used in host code.


vtavana (approved):
Thank you, @antonwolfy!

PR #1645, opened 2023-12-15
Implement of dpnp.clip()

npolina4:
dpnp.clip is implemented using  dpctl.tensor.clip.

### DIFF
--- a/dpnp/dpnp_iface_mathematical.py
+++ b/dpnp/dpnp_iface_mathematical.py
@@ -388,7 +388,7 @@ def ceil(
     )
 
 
-def clip(a, a_min, a_max, *, out=None, order="K"):
+def clip(a, a_min, a_max, *, out=None, order="K", **kwargs):
     """
     Clip (limit) the values in an array.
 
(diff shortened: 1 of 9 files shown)


dpnp/dpnp_iface_mathematical.py
```diff
@@ -387,6 +388,71 @@ def ceil(
...
+    usm_min = (
+        dpnp.get_usm_ndarray(a_min) if isinstance(a_min, dpnp_array) else a_min
+    )
+    usm_max = (
+        dpnp.get_usm_ndarray(a_max) if isinstance(a_max, dpnp_array) else a_max
+    )
+
+    if out is not None:
+        usm_out = dpnp.get_usm_ndarray(out)
+        dpt.clip(usm_arr, usm_min, usm_max, out=usm_out, order=order)
```
antonwolfy, line 448: dpctl doesn't support `order=None`, but numpy does and we need to handle that also

antonwolfy, line 448: we should return `dpnp_array` even when input `out` is usm_ndarray

tests/test_mathematical.py
```diff
@@ -2147,3 +2147,31 @@ def test_inplace_remainder(dtype):
...
+
+    result = dpnp.clip(dp_array, 2, 6)
+    expected = numpy.clip(np_array, 2, 6)
+    assert_allclose(expected, result)
+
+
+@pytest.mark.parametrize(
+    "dtype", get_all_dtypes(no_bool=True, no_none=True, no_complex=True)
+)
+def test_clip_arrays(dtype):
```
antonwolfy, line 2167: please add tests to check `sycl_queue` and `usm_type`

PR #1670, opened 2024-01-26
Clipping with both min and max values as None

antonwolfy:
Dpctl implemented `clip` function according to Python array API specification. And dpnp fully relies on `dpctl.tensor.clip` implementation.
While NumPy requires that only one of `min` or `max` be allowed to be `None` at the same time.

The PR proposes to align `dpnp.clip` with both `min` and `max` as `None` with NumPy implementation.


vtavana (approved):
Thank you, @antonwolfy!

vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy

PR #1712, opened 2024-02-15
Implemented nin and nout property for elementwise functions

npolina4:
Reused `dpctl.tensor.BinaryElementwiseFunc` and `dpctl.tensor.UnaryElementwiseFunc` classes for `DPNPBinaryFunc` and `DPNPUnaryFunc` for using nin and nout property.
Added special classes for `dpnp.angle`, `dpnp.real`, `dpnp.round`, and `dpnp.sign` because these `dpctl.tensor` functions has different behavior compared to `numpy` functions.
Updated all elementwise functions with using new classes.
Updated docs for elementwise functions. Added descriptions for input args.
Added tests.


antonwolfy:
@npolina4 , could you please add a detailed description to the PR?

tests/third_party/cupy/core_tests/test_ndarray_unary_op.py
```diff
@@ -141,6 +141,7 @@ def test_invert_zerodim(self):
     *testing.product({"xp": [numpy, cupy], "shape": [(3, 2), (), (3, 0, 2)]})
 )
 class TestBoolNeg(unittest.TestCase):
+    @pytest.mark.skip("dpnp does not raise TypeError")
```
antonwolfy, line 144: Does new framework allow to easily handle a behavior different from dpctl for some use case?
Can we get rid of the limitation here and in substract call below for boolean type?

dpnp/dpnp_algo/dpnp_elementwise_common.py
```diff
@@ -24,89 +24,453 @@
...
+                out=out,
+                where=where,
+                order=order,
+                dtype=dtype,
+                subok=subok,
+                **kwargs,
+            )
+        else:
+            raise NotImplementedError(
+                f"Requested function={self.name_} with args={x} and kwargs={kwargs} "
```
antonwolfy, line 181: Wouldn't it better to clear state what exact argument is out of supported range?

dpnp/dpnp_algo/dpnp_elementwise_common.py
```diff
@@ -24,89 +24,453 @@
...
+        )
+
+    def __call__(self, x, decimals=0, out=None):
+        if decimals != 0:
+            pass
+        elif dpnp.isscalar(x):
+            pass
+        else:
+            return super().__call__(x, out=out)
+        return call_origin(self.origin_fn, x, decimals=decimals, out=out)
```
antonwolfy, line 420: will lead to an exception when `self.origin_fn is None`

tests/third_party/cupy/core_tests/test_ndarray_unary_op.py
npolina4: Fix for this tests in dpctl.tensor functions

tests/test_mathematical.py
```diff
@@ -744,13 +737,6 @@ def test_positive(data, dtype):
         assert_allclose(result, expected)
 
 
-def test_positive_boolean():
```
antonwolfy, line 747: What is the reason you would like to remove the test?

NumPy doesn't support `positive` for an array of boolean type and previously dpnp also raised `TypeError` here.
While dpctl supports `positive` for boolean array without any warning or similar.
Python array API states [that](https://data-apis.org/array-api/2022.12/API_specification/generated/array_api.positive.html#positive):
```
x (array) – input array. Should have a numeric data type.
```
where `numeric data type` doesn't consider boolean type.
The same is stated for `negative` function there. And dpctl is aligned and raises an exception for boolean data type.
Thus I wonder if there is an issue in dpctl and the exception has to be raised for boolean data type passed in `positive` call.

Btw we definitely shouldn't decrease test coverage and remove test scenario except only the case where the same is verified by cupy test. But I failed to find any cupy test which might be considered as a replacement of that one.

tests/third_party/cupy/core_tests/test_ndarray_unary_op.py
```diff
@@ -147,5 +147,6 @@ def test_bool_neg(self):
             raise unittest.SkipTest("NumPy<1.13.0")
         shape = self.shape
         x = testing.shaped_random(shape, xp, dtype=numpy.bool_)
-        with pytest.raises(TypeError):
+        print(xp, x.dtype)
```
antonwolfy, line 150: please remove debugging print
```suggestion
```

tests/test_mathematical.py
```diff
@@ -878,13 +864,6 @@ def test_sign(data, dtype):
         assert_dtype_allclose(result, expected)
 
 
-def test_sign_boolean():
-    dpnp_a = dpnp.array([True, False])
-
-    with pytest.raises(TypeError):
-        dpnp.sign(dpnp_a)
```
antonwolfy, line 885: the same comment here...

- Do we have similar cupy test which cover that?
- Dpctl doesn't raise an exception, but probably has to. Should we address that as dpctl issue?

tests/test_mathematical.py
antonwolfy: 1. Will not it be confusing when we doesn't support boolean data type for `negative`, but support for `positive`? (as I can see in numpy PR#8967, it was one of the reasons why `numpy.positive` raises the exception from very beginning of implementation of `positive` call to numpy)
2. The exception which numpy raises in case of boolean data type passed in `positive` is inherit from `TypeError`:
```python
issubclass(numpy.core._exceptions._UFuncNoLoopError, TypeError)
# Out: True
```

vtavana:
Does the absence of the function signature in the documentation align with our intentions with the new approach?

image

.pre-commit-config.yaml
```diff
@@ -94,4 +94,4 @@ repos:
             "--disable=redefined-builtin",
             "--disable=unused-wildcard-import"
             ]
-        files: '^dpnp/dpnp_iface(|_arraycreation|_counting|_indexing|_libmath|_linearalgebra|_manipulation|_nanfunctions|_searching|_sorting|_statistics|_types).py$'
+        files: '^dpnp/dpnp_iface(|_arraycreation|_bitwise|_counting|_indexing|_libmath|_linearalgebra|_logic|_manipulation|_mathematical|_nanfunctions|_searching|_sorting|_statistics|_trigonometric|_types").py$'
```
antonwolfy, line 97: can be simplified now, like:
```suggestion
        files: '^dpnp/dpnp_iface.*'
```

dpnp/dpnp_algo/dpnp_elementwise_common.py
```diff
@@ -24,89 +24,546 @@
...
+    BinaryElementwiseFunc,
+    UnaryElementwiseFunc,
+)
 
 import dpnp
-import dpnp.backend.extensions.vm._vm_impl as vmi
 from dpnp.dpnp_array import dpnp_array
 from dpnp.dpnp_utils import call_origin
 
 __all__ = [
```
antonwolfy, line 37: The code coverage is only 59% per that file. The missing lines:
> 147-148, 153-154, 160, 165-166, 172-173, 180-183, 318-319, 324-325, 331, 336-337, 344, 351-354, 452, 458, 529, 540, 550, 587-631

Can we increase the reporting coverage rate>

doc/conf.py
```diff
@@ -188,7 +192,17 @@
 
 # -- Extension configuration -------------------------------------------------
 
+
 # -- Options for todo extension ----------------------------------------------
+def _can_document_member(member, *args, **kwargs):
+    if isinstance(member, (DPNPBinaryFunc, DPNPUnaryFunc)):
```
antonwolfy, line 198: Should it be replaced with `issubclass` to cover subclasses also, like `DPNPAngle`?

dpnp/dpnp_algo/dpnp_elementwise_common.py
```diff
@@ -24,89 +24,546 @@
...
+                    f"(got '{order}')"
+                )
+            x_usm = dpnp.get_usm_ndarray(x)
+            out_usm = None if out is None else dpnp.get_usm_ndarray(out)
+            res_usm = super().__call__(x_usm, out=out_usm, order=order)
+            if out is not None and isinstance(out, dpnp_array):
+                return out
+            return dpnp_array._create_from_usm_ndarray(res_usm)
+
+        return call_origin(
```
antonwolfy, line 194: Is there any simple way to get rid of `call_origin` completely for all elementwise functions?

doc/conf.py
npolina4: It works for all subclasses.

dpnp/dpnp_algo/dpnp_elementwise_common.py
npolina4: Elementwise functions may raise NotImplementedError for all unsupported parameters

antonwolfy: Yes, I think we should be that way.

tests/third_party/cupy/core_tests/test_ndarray_math.py
```diff
@@ -52,13 +51,14 @@ def test_round_out(self, xp):
...
-@pytest.mark.usefixtures("allow_fall_back_on_numpy")
 class TestRoundHalfway(unittest.TestCase):
     shape = (20,)
 
     @testing.for_float_dtypes()
-    @testing.numpy_cupy_array_equal()
+    @testing.numpy_cupy_allclose(atol=1e-5)
     def test_round_halfway_float(self, xp, dtype):
+        if self.decimals is -3:
+            pytest.skip("decimals=-3 have has a rounding error in this case")
```
antonwolfy, line 61: why do we need to mute that?

dpnp/dpnp_algo/dpnp_elementwise_common.py
```diff
@@ -304,11 +303,15 @@ def __call__(
                 f"Requested function={self.name_} with subok={subok} "
                 "isn't currently supported."
             )
-        elif dpnp.isscalar(x1) and dpnp.isscalar(x2):
-            # input has to be an array
-            raise NotImplementedError(
-                f"Requested function={self.name_} with args={x1, x2} "
-                "isn't currently supported."
+        elif (
+            not dpnp.is_supported_array_or_scalar(x1)
```
antonwolfy, line 307: Shouldn't we use `check_supported_arrays_type` here instead?

dpnp/dpnp_algo/dpnp_elementwise_common.py
```diff
@@ -24,3096 +24,444 @@
...
+            docs,
+        )
 
-    # dpctl.tensor only works with usm_ndarray or scalar
-    x1_usm_or_scalar = dpnp.get_usm_ndarray_or_scalar(x1)
-    x2_usm_or_scalar = dpnp.get_usm_ndarray_or_scalar(x2)
-    out_usm = None if out is None else dpnp.get_usm_ndarray(out)
+    def __call__(self, x, decimals=0, out=None, dtype=None):
+        if decimals != 0:
+            if numpy.issubdtype(x.dtype, numpy.integer) and dtype is None:
```
antonwolfy, line 411:
```suggestion
            if dpnp.issubdtype(x.dtype, dpnp.integer) and dtype is None:
```

antonwolfy (approved):
Thank you @npolina4 , I have only minor comment.

dpnp/dpnp_iface_mathematical.py
```diff
@@ -1819,1017 +1940,797 @@ def maximum(
...
+array([-2., -2., -0.,  0.,  2.,  2.,  2.])
+"""
 
-    >>> np.signbit(np.array([1, -2.3, 2.1]))
-    array([False,  True, False])
 
-    """
+rint = DPNPUnaryFunc(
+    "rint",
+    ti._round_result_type,
```
antonwolfy, line 2499: Is it expected that `_mkl_round_to_call` is never used now?

PR #1718, opened 2024-02-17
Adding pylint pre-commit hook

antonwolfy:
The PR extends pre-commit hooks with `pylint` configuration.
The current files filter for the verification is set to `dpnp/dpnp_iface*.py` (excluding files from #1712) to reduce number of issues. The other files will be added to the scope in later PRs

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -43,6 +43,9 @@ per-file-ignores =
     dpnp/linalg/__init__.py: F401
     dpnp/random/__init__.py: F401
     dpnp/dpnp_iface.py: D205
+    dpnp/dpnp_iface_manipulation.py: D205
+    dpnp/dpnp_iface_nanfunctions.py: D205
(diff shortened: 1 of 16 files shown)


.flake8
```diff
@@ -43,6 +43,9 @@ per-file-ignores =
     dpnp/linalg/__init__.py: F401
     dpnp/random/__init__.py: F401
     dpnp/dpnp_iface.py: D205
```
vlad-perevezentsev, line 45: I think it would be better to ignore this rule in all `dpnp` files. 
Cause the line break does not affect the function description in the generated documentation.

antonwolfy: Totally agree, I'll disable the warning globally

vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy for adding a useful pre-commit hook

PR #1735, opened 2024-03-01
Replace use of queue mgr

oleksandr-pavlyk:
`DPCTLQueueMgr_*` is being removed, see https://github.com/IntelPython/dpctl/pull/1576.

This PR removes use of `DPCTLQueueMgr_*` in `queue_sycl.cpp` and `queue_sycl.hpp`. 

It effectively reverts to using `DPNP_LOCAL_QUEUE=1`.

Examples, documentation were also updated to avoid uses of `device_selector` as it is being removed.

### DIFF
--- a/dpnp/backend/src/queue_sycl.hpp
+++ b/dpnp/backend/src/queue_sycl.hpp
@@ -40,16 +40,6 @@
 
 #include <ctime>
 
-#if !defined(DPNP_LOCAL_QUEUE)
-#if defined __has_include
-#if __has_include(<dpctl_sycl_interface.h>)
-#include <dpctl_sycl_interface.h>
-#else
-#include <dpctl_sycl_queue_manager.h>
-#endif
-#endif
-#endif
-
 #include "dpnp_pstl.hpp" // this header must be included after <mkl.hpp>
 
 #include "verbose.hpp"
@@ -69,10 +59,8 @@ namespace mkl_rng = oneapi::mkl::rng;
  */
 class backend_sycl
 {
-#if defined(DPNP_LOCAL_QUEUE)
     static sycl::queue *queue; /**< contains SYCL queue pointer initialized in
                                   @ref backend_sycl_queue_init */
-#endif
     static mkl_rng::mt19937
         *rng_engine; /**< RNG MT19937 engine ptr. initialized in @ref
                         backend_sycl_rng_engine_init */
@@ -83,10 +71,7 @@ class backend_sycl
     static void destroy()
     {
         backend_sycl::destroy_rng_engine();
-#if defined(DPNP_LOCAL_QUEUE)
         delete queue;
-        queue = nullptr;
-#endif
     }
 
     static void destroy_rng_engine()
@@ -101,10 +86,9 @@ class backend_sycl
 public:
     backend_sycl()
     {
-#if defined(DPNP_LOCAL_QUEUE)
         queue = nullptr;
         rng_engine = nullptr;
-#endif
+	rng_mcg59_engine = nullptr;
     }
 
     virtual ~backend_sycl()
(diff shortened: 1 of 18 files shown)


oleksandr-pavlyk:
This change should not have made any difference, but please see if tests say otherwise.

dpnp/backend/src/queue_sycl.hpp
```diff
@@ -40,25 +41,108 @@
...
+                "Could not create MCG59 engine with given seed"
+            );
+        }
+
+        rng_mt19937_engine_ptr.swap(rng_eng_mt19937);
+        rng_mcg59_engine_ptr.swap(rng_eng_mcg59);
+    }
+
+    bool backend_sycl_is_cpu() const {
+        if (queue_ptr) {
```
antonwolfy, line 98: Is it possible to have `queue_ptr == nullptr`? If memory can't be allocated for `queue_ptr`, then there will be runtime exception thrown from the constructor.
Otherwise, probably we also need to add similar check `if (queue_ptr)` to `set_rng_engines_seed` method.

dpnp/backend/src/queue_sycl.hpp
```diff
@@ -67,26 +151,23 @@ namespace mkl_rng = oneapi::mkl::rng;
  * initialization order is undefined. This class postpone initialization of the
  * SYCL queue and mt19937 random number generation engine.
  */
+#if 0
```
antonwolfy, line 154: Do we need to clean up and to remove `class backend_sycl`?

dpnp/backend/src/queue_sycl.hpp
oleksandr-pavlyk: The `shared_ptr<queue>` is default initialized by default constructor and so has `nullptr` content until after it is reset with `std::make_shared`. This command would call `sycl::queue` constructor which may throw an exception is queue construction was not successful. So exception will be thrown, but I'd keep the check here.

dpnp/backend/src/queue_sycl.hpp
oleksandr-pavlyk: Yes, let me push that. I thought it was done

dpnp/backend/src/queue_sycl.hpp
oleksandr-pavlyk: Actually, since constructor of the class is private, it can only be created via `get` static method, which would call private constructor. Any exceptions raised during this constructor call would result in the instance of this class not being created, and hence all `bool(queue_ptr)` tests in regular methods are superfluous. I am going to remove them.

antonwolfy (approved):
@oleksandr-pavlyk, thanks a lot for implementing a proper local queue handing in `dpnp`.

dpnp/backend/src/queue_sycl.hpp
```diff
@@ -69,120 +60,95 @@ namespace mkl_rng = oneapi::mkl::rng;
...
-     * Return the @ref rng_engine to the user
-     */
-    static mkl_rng::mt19937 &get_rng_engine()
+    bool backend_sycl_is_cpu() const
     {
-        if (!rng_engine) {
-            backend_sycl_rng_engine_init();
-        }
-        return *rng_engine;
+        const sycl::queue &q = *queue_ptr;
```
diptorupd, line 111: What is the intent of this line?

diptorupd: Do you really intend to store a reference to the `sycl::queue`?

dpnp/backend/src/queue_sycl.hpp
```diff
@@ -69,120 +60,95 @@ namespace mkl_rng = oneapi::mkl::rng;
...
+        : queue_ptr{}, rng_mt19937_engine_ptr{}, rng_mcg59_engine_ptr{}
     {
-        if (!rng_engine) {
-            backend_sycl_rng_engine_init();
+        const sycl::property_list &prop =
+            (is_verbose_mode())
+                ? sycl::property_list{sycl::property::queue::enable_profiling()}
+                : sycl::property_list{};
+        queue_ptr =
+            std::make_shared<sycl::queue>(sycl::default_selector_v, prop);
```
diptorupd, line 124: Why do you need to call `make_shared`? `sycl::queue` is already a shared pointer.

dpnp/backend/src/queue_sycl.hpp
oleksandr-pavlyk: It is just so that I can call `q.get_device()`. Of course `queue_ptr->get_device()` could be used directly.

oleksandr-pavlyk: Resolved

dpnp/backend/src/queue_sycl.hpp
oleksandr-pavlyk: Changed `backend_sycl` class to store class instances instead of storing shared pointers to them.

oleksandr-pavlyk:
@antonwolfy Are we good to merge?

antonwolfy (approved):
I don't have any comment. Thank you @oleksandr-pavlyk

PR #1742, opened 2024-03-13
Remove find_common_type which is never used in DPNP and deprecated since NumPy 1.25.0

antonwolfy:
The PR removes unused function `find_common_type` from Cython backend layout.
The function also was present in NumPy and deprecated since `1.25.0`, so also it removes documentation reference on it.


vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy

PR #1754, opened 2024-03-20
Update thrid party testing f/w

antonwolfy:
The testing framework of third party tests were not updated for a long time. Now it might be difficult to align the tests without alignments in the base functionality, since many internal methods were rewritten or changed.

The PR attempts to update the base testing framework to speed up future development.


tests/third_party/cupy/testing/_random.py
```diff
@@ -94,12 +94,12 @@ def fix_random():
     """Decorator that fixes random numbers in a test.
 
     This decorator can be applied to either a test case class or a test method.
-    It should not be applied within ``_condition.retry`` or
-    ``_condition.repeat``.
+    It should not be applied within ``condition.retry`` or
```
vlad-perevezentsev, line 97: What is the reason for this renaming?
We have `_condition,py` instead of `condition.py` file

tests/third_party/cupy/random_tests/test_sample.py
```diff
@@ -40,7 +39,8 @@ def test_zero_sizes(self):
 
 
 # @testing.fix_random()
```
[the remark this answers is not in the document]
antonwolfy: Let leave that to a time when we will update all random tests with upcoming changes we are planned around here.

tests/third_party/cupy/testing/_random.py
antonwolfy: It is aligned with cupy. I would like to avoid additional lines of diff in base f/w between our snapshot and what is in cupy code.

vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy

PR #1762, opened 2024-03-28
Temporary mute tests for `dpnp.solve`

antonwolfy:
The PR proposes to temporary mute tests in public CI which causes hanging until the issue is resolved.

- [x] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to issue with a reproducer?
- [ ] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?

PR #1778, opened 2024-04-08
Fix `dpnp.linalg.solve()` hang on CPU

vlad-perevezentsev:
This PR adds a workaround for solving `dpnp.linalg.solve()` hang issue for batch implementation on CPU as adding `wait()` for each host task before calling MKL function (`gesv`) in the loop. 
This is a known issue and was addressed to MKL team.

Also this PR unlocks skipped tests added in #1762 


tests/third_party/cupy/linalg_tests/test_solve.py
```diff
@@ -47,7 +47,6 @@ def check_x(self, a_shape, b_shape, xp, dtype):
         testing.assert_array_equal(b_copy, b)
         return result
 
-    @pytest.mark.skipif(is_cpu_device(), reason="SAT-6842")
```
antonwolfy, line 50: `tests/test_sycl_queue.py::test_solve` is muted with the same reason
Should the skip mark be removed also there?
Otherwise probably we need to update the reason there...

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -1882,6 +1888,15 @@ def dpnp_solve(a, b):
...
+            # on CPU causes deadlock due to serialization of all host tasks
+            # in the queue.
+            # We need to wait for each host tasks before calling _gesv to avoid deadlock.
+            if is_cpu_device:
+                b_ht_copy_ev[i].wait()
+                ht_lapack_ev[i].wait()
+
         for i in range(batch_size):
             ht_lapack_ev[i].wait()
             b_ht_copy_ev[i].wait()
```
antonwolfy, line 1902: I believe it would be more safe to have:
```suggestion
            if not is_cpu_device:
                ht_lapack_ev[i].wait()
                b_ht_copy_ev[i].wait()
```

tests/third_party/cupy/linalg_tests/test_solve.py
vlad-perevezentsev: Yes, I am going to do it in a separate PR since not all functions with this skip mark work (tests solve works)

antonwolfy (approved):

PR #1780, opened 2024-04-09
Update `dpnp.linalg.eig/eigvals`

vlad-perevezentsev:
This PR proposes to update `dpnp.linalg.eig()` and `dpnp.linalg.eigvals()` functions using `numpy` call. 
Using `numpy.linalg.eig/eigvals` is a temporary solution until MKL implements `oneapi::mkl::lapack::geev` for USM interface.

This PR also removes `dpnp_algo_linalg.pyx` file because it no longer contains Cython functions


vlad-perevezentsev:
@antonwolfy also I would also remove the backend implementation of `dpnp_eigvals_c` and `dpnp_eig_c` , because it is only implemented for symmetric arrays which is wrong.
What do you think about this?

antonwolfy:
> @antonwolfy also I would also remove the backend implementation of `dpnp_eigvals_c` and `dpnp_eig_c` cause it is only implemented for symmetric arrays which is wrong. What do you think about this?

We definitely need to plan a clean up, but I would prefer to do that by a separate PR (like to remove all unused linalg function from backend and so on). But needs to be careful to avoid an unexpected impact on numba-dpex.

dpnp/linalg/dpnp_iface_linalg.py
```diff
@@ -249,20 +246,101 @@ def det(a):
...
+    array([1.+1.j, 1.-1.j])
+    array([[0.70710678+0.j        , 0.70710678-0.j        ],
+           [0.        -0.70710678j, 0.        +0.70710678j]])
+
+    Complex-valued matrix with real e-values (but complex-valued e-vectors);
+    note that ``a.conj().T == a``, i.e., `a` is Hermitian.
+
+    >>> a = np.array([[1, 1j], [-1j, 1]])
+    >>> w, v = LA.eig(a)
+    >>> w; v
```
antonwolfy, line 315:
```suggestion
    >>> w, v
```

dpnp/linalg/dpnp_iface_linalg.py
```diff
@@ -249,20 +246,101 @@ def det(a):
...
+                              (conjugate symmetric) or a real symmetric matrix.
+    :obj:`dpnp.linalg.eigvalsh` : Compute the eigenvalues of a complex Hermitian or
+                                  real symmetric matrix.
+
+    Examples
+    --------
+    >>> import dpnp as np
+    >>> from dpnp import linalg as LA
+
+    (Almost) trivial example with real e-values and e-vectors.
```
antonwolfy, line 292:
```suggestion
    (Almost) trivial example with real eigenvalues and eigenvectors.
```

dpnp/linalg/dpnp_iface_linalg.py
```diff
@@ -249,20 +246,101 @@ def det(a):
...
+
+    >>> w, v = LA.eig(np.diag((1, 2, 3)))
+    >>> w; v
+    array([1., 2., 3.])
+    array([[1., 0., 0.],
+           [0., 1., 0.],
+           [0., 0., 1.]])
+
+    Real matrix possessing complex e-values and e-vectors; note that the
+    e-values are complex conjugates of each other.
```
antonwolfy, line 302:
```suggestion
    Real matrix possessing complex eigenvalues and eigenvectors;
    note that the eigenvalues are complex conjugates of each other.
```

dpnp/linalg/dpnp_iface_linalg.py
```diff
@@ -249,20 +246,101 @@ def det(a):
...
+
+    >>> w, v = LA.eig(np.array([[1, -1], [1, 1]]))
+    >>> w; v
+    array([1.+1.j, 1.-1.j])
+    array([[0.70710678+0.j        , 0.70710678-0.j        ],
+           [0.        -0.70710678j, 0.        +0.70710678j]])
+
+    Complex-valued matrix with real e-values (but complex-valued e-vectors);
+    note that ``a.conj().T == a``, i.e., `a` is Hermitian.
```
antonwolfy, line 311:
```suggestion
    Complex-valued matrix with real eigenvalues (but complex-valued
    eigenvectors); note that ``a.conj().T == a``, i.e., `a` is Hermitian.
```

dpnp/linalg/dpnp_iface_linalg.py
```diff
@@ -249,20 +246,101 @@ def det(a):
...
-            return dpnp_eig(x1_desc)
+    dpnp.check_supported_arrays_type(a)
+    check_stacked_2d(a)
+    check_stacked_square(a)
 
-    return call_origin(numpy.linalg.eig, x1)
+    a_sycl_queue = a.sycl_queue
+    a_usm_type = a.usm_type
+
+    w_np, v_np = numpy.linalg.eig(dpnp.asnumpy(a))
```
antonwolfy, line 339: please add a comment why we are calling numpy here

dpnp/linalg/dpnp_iface_linalg.py
```diff
@@ -334,32 +412,70 @@ def eigh(a, UPLO="L"):
...
-    x1_desc = dpnp.get_dpnp_descriptor(input, copy_when_nondefault_queue=False)
-    if x1_desc:
-        if x1_desc.size > 0:
-            return dpnp_eigvals(x1_desc).get_pyobj()
+    dpnp.check_supported_arrays_type(a)
+    check_stacked_2d(a)
+    check_stacked_square(a)
 
-    return call_origin(numpy.linalg.eigvals, input)
+    w_np = numpy.linalg.eigvals(dpnp.asnumpy(a))
```
antonwolfy, line 477: please add a comment why we are calling numpy here

PR #1792, opened 2024-04-19
Add `dpnp.linalg.lstsq()` implementation

vlad-perevezentsev:
This PR adds a new dpnp.linalg.lstsq) function to calculate the least-squares solution using `dpnp.linalg.svd()`.
Also it adds a new helper function `check_2d` to check if the input arrays are two-dimensional.


dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -1224,6 +1285,60 @@ def dpnp_inv(a):
...
+        s1[sing_vals] = 0
+        rank -= sing_vals.sum(dtype="int32")
+
+    # Solve the least-squares solution
+    # x = vh.T.conj() @ diag(s1) @ u.T.conj() @ b
+    z = (dpnp.dot(b.T, u.conj()) * s1).T
+    x = dpnp.dot(vh.T.conj(), z)
+    # Calculate squared Euclidean 2-norm for each column in b - a*x
+    if m <= n or rank != n:
+        resids = dpnp.empty(
```
antonwolfy, line 1332: I guess we can use `empty_like` here instead

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -503,6 +504,35 @@ def _multi_svd_norm(x, row_axis, col_axis, op):
     return result
 
 
+def _nrm2_last_axis(x):
```
antonwolfy, line 507: is there any benefit to have a separate call instead of using `linalg.norm`?

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -2201,22 +2341,42 @@ def dpnp_svd(
             s = dpnp.abs(s)
             return dpnp.sort(s)[..., ::-1]
 
-    uv_type = _common_type(a)
+    uv_type = (
```
antonwolfy, line 2344: why do we need to calculate dtype based on `related_arrays`? what is the use case?

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -2277,22 +2439,20 @@ def dpnp_svd(
         u_shape,
         dtype=uv_type,
         order="F",
-        usm_type=a_usm_type,
-        sycl_queue=a_sycl_queue,
+        usm_type=usm_type,
```
antonwolfy, line 2442: guessing everywhere here we can use `empty_like`, because we have the same `order`, `usm_type` and `sycl_queue` as `a_h`

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -2031,16 +2146,29 @@ def dpnp_slogdet(a):
...
+    # Set USM type and SYCL queue to be used based on `a`
+    # and optionally provided `related_arrays`.
+    # If `related_arrays` is not provided, default to USM type and SYCL queue of `a`.
+    # Otherwise, determine USM type and SYCL queue using
+    # compute-follows-data execution model for `a` and `related arrays`.
+    if related_arrays is None:
+        usm_type = a.usm_type
+        exec_q = a.sycl_queue
+    else:
+        usm_type, exec_q = get_usm_allocations([a] + related_arrays)
```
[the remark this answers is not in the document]
vlad-perevezentsev: This line is uncovered in `dpnp_svd_batch` cause `dpnp_lstsq` only accepts 2D `a` array.
I added this line to align `dpnp_svd_batch` implementation with `dpnp_svd`

tests/third_party/cupy/linalg_tests/test_solve.py
```diff
@@ -226,6 +226,98 @@ def test_pinv_size_0(self):
...
+    @testing.for_float_dtypes(no_float16=True)
+    @testing.numpy_cupy_allclose(atol=1e-3)
+    def test_warn_rcond(self, xp, dtype):
+        a = testing.shaped_random((3, 3), xp, dtype)
+        b = testing.shaped_random((3,), xp, dtype)
+        with testing.assert_warns(FutureWarning):
+            return xp.linalg.lstsq(a, b)
+
+
 class TestTensorInv(unittest.TestCase):
```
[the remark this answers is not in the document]
vlad-perevezentsev: Thank you for noticing that. 
I will fix it for `TestInv`, `TestInvInvalid` and `TestTensorInv`

tests/third_party/cupy/linalg_tests/test_solve.py
```diff
@@ -226,6 +226,98 @@ def test_pinv_size_0(self):
         self.check_x((2, 0, 3), rcond=1e-15)
```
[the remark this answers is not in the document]
vlad-perevezentsev: It was removed when `fix_random()` did not work.
We can add it now.
Thank you

dpnp/linalg/dpnp_utils_linalg.py
antonwolfy: Doesn't that mean we will use `_gesvd` with wider dtype than required, which may have negative performance drawback?

dpnp/linalg/dpnp_utils_linalg.py
vlad-perevezentsev: This improves performance rather than `return dpnp.square(dpnp.linalg.norm(x, axis=-1))`


```
def old_impl(x):
    real_dtype = _real_type(x.dtype)
    if dpnp.issubdtype(x.dtype, dpnp.complexfloating):
        return dpnp.sum(dpnp.abs(x) ** 2, axis=-1, dtype=real_dtype)
    else:
        return dpnp.sum(dpnp.square(x), axis=-1, dtype=real_dtype)

def new_impl(x):
    return dpnp.square(dpnp.linalg.norm(x, axis=-1))

```
```
In [1]: import numpy
In [2]: import dpnp
In [3]: from qwe import new_impl, old_impl

In [4]: na = numpy.random.randint(-10**4, 10**4, size=(1024,512))

In [5]: a_dp = dpnp.array(na,dtype='f4',device='cpu')

In [6]: %timeit run_old = old_impl(a_dp)
1.04 ms ± 25.5 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [7]: %timeit run_new = new_impl(a_dp)
2.15 ms ± 454 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [8]: a_dp = dpnp.array(na,dtype='f4',device='gpu')

In [9]: %timeit run_old = old_impl(a_dp)
394 µs ± 8.44 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [10]: %timeit run_new = new_impl(a_dp)
839 µs ± 6.49 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [11]: a_dp = dpnp.array(na,dtype='int64',device='cpu')

In [12]: %timeit run_old = old_impl(a_dp)
1.52 ms ± 302 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [13]: %timeit run_new = new_impl(a_dp)
4.5 ms ± 311 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [14]: a_dp = dpnp.array(na,dtype='int64',device='gpu')

In [15]: %timeit run_old = old_impl(a_dp)
1.05 ms ± 444 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [16]: %timeit run_new = new_impl(a_dp)
1.08 ms ± 52.4 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

```

antonwolfy (approved):
Thank you @vlad-perevezentsev

PR #1799, opened 2024-04-23
Fixed an issue in `dpnp.random.vonmises` function identified by running tests scope on PVC

antonwolfy:
Function `dpnp.isnan` doesn't support scalar input and raises `TypeError` exception. But the call with scalar input is using by implementation of `dpnp.random.vonmises` function.

The PR proposes to always call `numpy.isnan` there instead, because it's secured that `kappa` is a scalar by a prior check in the code.


vlad-perevezentsev (approved):
Thank you @antonwolfy

PR #1810, opened 2024-05-07
update wheel tag with pep600

ekomarova:
This PR is designed to fix the problem described here: [SAT-6873](https://jira.devtools.intel.com/browse/SAT-6873). I did some experiments and found out that using pep600 wheel tag prevents the wheel from being installed on non-supported platforms

### DIFF
--- a/conda-recipe/build.sh
+++ b/conda-recipe/build.sh
@@ -23,7 +23,7 @@ SKBUILD_ARGS=("${SKBUILD_ARGS[@]}" "-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON")
 if [ "$CONDA_PY" == "36" ]; then
     WHEELS_BUILD_ARGS=("-p" "manylinux1_x86_64")
 else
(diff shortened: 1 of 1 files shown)


conda-recipe/build.sh
```diff
@@ -23,7 +23,7 @@ SKBUILD_ARGS=("${SKBUILD_ARGS[@]}" "-DCMAKE_VERBOSE_MAKEFILE:BOOL=ON")
 if [ "$CONDA_PY" == "36" ]; then
     WHEELS_BUILD_ARGS=("-p" "manylinux1_x86_64")
```
antonwolfy, line 24: I believe that needs to be removed, since env variable `CONDA_PY` is never used now, including any Jenkin job I and  because python 3.6 is out of support.

ekomarova: Removed

antonwolfy (approved):
Thank you @ekomarova for resolving the issue with `dpnp` wheels package dependencies.

PR #1821, opened 2024-05-13
Use explicit comparison in `if` condition of github action

antonwolfy:
The PR proposes to use explicit comparison against a bool flag in `if` conditions of steps for `Build docs` GH action.
It will resolve the unexpected behavior in run of the action while merging a PR to master branch, since `if env.<env name>` does not result to `False` when `env.<env name> == 'false'`.


vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy

PR #1823, opened 2024-05-13
Implement `dpnp.histogram_bin_edges`

antonwolfy:
This PR implements `dpnp.histogram_bin_edges` function. All the required tests are added to verify the new functionality.
The function leverages on existing internal functions previously used in implementation of `dpnp.histogram`.


dpnp/dpnp_iface_histograms.py
```diff
@@ -337,3 +340,88 @@ def histogram(a, bins=10, range=None, density=None, weights=None):
...
+    array([1, 2])
+
+    This function allows one set of bins to be computed, and reused across
+    multiple histograms:
+
+    >>> shared_bins = np.histogram_bin_edges(arr, bins=5)
+    >>> shared_bins
+    array([0., 1., 2., 3., 4., 5.])
+
+    >>> gid = np.array([0, 1, 1, 0, 1, 1, 0, 1, 1], sycl_queue=arr.sycl_queue)
```
vlad-perevezentsev, line 406: I am not sure you should pass `sycl_queue` here.
According to the examples above `gid` should be allocated on the same queue as `arr`

antonwolfy: It's because the default queue is used, in case when you consequently run the code from the example.
But in general, `arr` might be allocating on another queue and when we must secure `gid` is allocating on the same queue to comply with compute follows data.
By passing explicitly `sycl_queue=arr.sycl_queue` I wanted to point on that one more time.

dpnp/dpnp_iface_histograms.py
```diff
@@ -337,3 +340,88 @@ def histogram(a, bins=10, range=None, density=None, weights=None):
         return n / db / n.sum(), bin_edges
 
     return n, bin_edges
+
+
+def histogram_bin_edges(a, bins=10, range=None, weights=None):
+    r"""
```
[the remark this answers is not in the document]
antonwolfy: Removed

dpnp/dpnp_iface_histograms.py
vlad-perevezentsev: Maybe it would be better to leave a comment for this line?
What do you think?

antonwolfy: Why do you think it would be better?

vlad-perevezentsev: Сause in another examples, for example, in dpnp.searsorted([297 line](https://github.com/IntelPython/dpnp/blob/master/dpnp/dpnp_iface_searching.py#L297)), we do not pass sycl_queue for `v`

antonwolfy: Ok, removed passing `sycl_queue` keyword from the example.
Btw, I still don't feel that having it in this example was confusing. It's just one of the possible way to allocate an array. But since it doesn't relate to reviewing functionality I agreed to align with examples of other functionality.

We can discuss if we would like to pass `sycl_queue` in the examples separately.

vlad-perevezentsev (approved):
Thank you @antonwolfy for implementing this. Looks good

PR #1825, opened 2024-05-14
Fix dpnp.flatten

npolina4:
dpnp.flatten does not work for not default queue because sycl_device and sycl_queue cannot be provide together. 
This is fixed by this PR


dpnp/dpnp_array.py
```diff
@@ -856,24 +856,24 @@ def flatten(self, order="C"):
...
 
-        if self.size > 0:
-            dpt._copy_utils._copy_from_usm_ndarray_to_usm_ndarray(
-                new_arr._array_obj, self._array_obj
-            )
-            new_arr._array_obj = dpt.reshape(new_arr._array_obj, (self.size,))
+        """
 
-        return new_arr
+        res = dpnp.reshape(self, -1, order=order)
```
antonwolfy, line 870: what if to pass `copy=True` instead of `res.copy()`?

dpnp/dpnp_array.py
```diff
@@ -856,24 +856,24 @@ def flatten(self, order="C"):
         --------
         :obj:`dpnp.ravel`, :obj:`dpnp.flat`
```
antonwolfy, line 857: need to add a description per each function

dpnp/dpnp_array.py
```diff
@@ -856,24 +856,24 @@ def flatten(self, order="C"):
         --------
```
antonwolfy, line 856: need to update return type with `dpnp.ndarray`

dpnp/dpnp_array.py
```diff
@@ -841,39 +841,37 @@ def flatten(self, order="C"):
...
-            'A' means to flatten in column-major order if a is Fortran contiguous in memory, row-major order otherwise.
-            'K' means to flatten a in the order the elements occur in memory. The default is 'C'.
+        order : {"C", "F"}, optional
+            Read the elements using this index order, and place the elements
+            into the reshaped array using this index order. ``"C"`` means to
+            read / write the elements using C-like index order, with the last
+            axis index changing fastest, back to the first axis index changing
+            slowest. ``"F"`` means to read / write the elements using
+            Fortran-like index order, with the first index changing fastest,
+            and the last index changing slowest.
```
antonwolfy, line 851: I believe it would be much more readable to have it as a list:

```suggestion
            Read the elements using this index order, and place the elements
            into the reshaped array using this index order.

                - "C" means to read / write the elements using C-like index
                  order, with the last axis index changing fastest, back to the
                  first axis index changing slowest.
                - "F" means to read / write the elements using Fortran-like
                  index order, with the first index changing fastest, and the
                  last index changing slowest.

            The default is ``"C"``.
```

dpnp/dpnp_array.py
```diff
@@ -841,39 +841,37 @@ def flatten(self, order="C"):
...
 
-        if self.size > 0:
-            dpt._copy_utils._copy_from_usm_ndarray_to_usm_ndarray(
-                new_arr._array_obj, self._array_obj
-            )
-            new_arr._array_obj = dpt.reshape(new_arr._array_obj, (self.size,))
+        """
 
-        return new_arr
+        return dpnp.reshape(self, -1, order=order, copy=True)
```
antonwolfy, line 874:
```suggestion
        return self.reshape(-1, order=order, copy=True)
```

dpnp/dpnp_array.py
```diff
@@ -841,39 +841,37 @@ def flatten(self, order="C"):
 
         Parameters
```
antonwolfy, line 842:
```suggestion
        For full documentation refer to :obj:`numpy.ndarray.flatten`.

        Parameters
```

antonwolfy (approved):
LGTM! Thank you @npolina4

PR #1831, opened 2024-05-17
Add default exclusion of slow tests in pytest

vlad-perevezentsev:
This PR suggests to exclude running slow tests (marked as slow) when pytest is called by default and to register custom markers such as `slow` and `multi_gpu`. 

These changes will help to avoid spontaneous hangs in internal CI since some tests are decorated by `@testing.slow` which is not implemented yet. 
It is assumed that such tests (usually using huge data) should perform computation on multiple devices to reduce the computation load. Internal CI contains one device and running such tests causes spontaneous hangs.

To run all tests including those marked as slow (`@testing.slow` ) use `pytest -m "not slow or slow"`
```
$ pytest tests/third_party/cupy/math_tests/test_sumprod.py::TestSumprod

collected 29 items / 1 deselected / 28 selected 
tests/third_party/cupy/math_tests/test_sumprod.py ...........................s    

========================= 27 passed, 1 skipped, 1 deselected in 1.96s=========================

```

```
$ pytest -m "not slow or slow" tests/third_party/cupy/math_tests/test_sumprod.py::TestSumprod

collected 29 items  
tests/third_party/cupy/math_tests/test_sumprod.py ............................s 

=========================28 passed, 1 skipped in 3.77s=========================

```

Also removes test skipping for `dpnp.power()` due to using the latest OneAPI compiler for coverage GitHub Action #1777 

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -67,7 +67,7 @@ jobs:
     strategy:
       matrix:
         python: ['3.9', '3.10', '3.11']
-        os: [ubuntu-20.04, windows-latest]
+        os: [ubuntu-20.04, windows-2019]
 
     permissions:
       # Needed to cancel any previous runs that are not completed for a given workflow
@@ -77,7 +77,7 @@ jobs:
 
     defaults:
       run:
-        shell: ${{ matrix.os == 'windows-latest' && 'cmd /C CALL {0}' || 'bash -l {0}' }}
+        shell: ${{ matrix.os == 'windows-2019' && 'cmd /C CALL {0}' || 'bash -l {0}' }}
 
     continue-on-error: true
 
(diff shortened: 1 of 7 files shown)


vlad-perevezentsev:
about my last commit
@antonwolfy do we need to keep using `-p no:warnings` for `addopts`?
Does it make sense to add `filterwarnings` section and handle warnings in it?

antonwolfy (approved):
Thank you @vlad-perevezentsev, the change is very helpful and makes the tests more stable.
The `filterwarnings` marker can be added separately.

vlad-perevezentsev:
> Thank you @vlad-perevezentsev, the change is very helpful and makes the tests more stable. The `filterwarnings` marker can be added separately.

I will keep `filterwarnings` registration in markers for this PR cause running tests does not work without it.
I will remove it in the next PR along with removing `-p no:warnings` and add warning handling

PR #1832, opened 2024-05-17
Fix dpnp.linalg.eigh() hang with complex dtypes on CPU

vlad-perevezentsev:
This PR suggests adding a temporary w/a as waiting also for each host task before calling `oneapi::mkl::lapack::heevd` to avoid a deadlock for batch implementation of  `dpnp.linalg.eigh` for complex data types on CPU with the prerelease version of MKL.


antonwolfy (approved):
Thank you @vlad-perevezentsev

PR #1838, opened 2024-05-21
Get rid of falling back on numpy in `dpnp.put`

antonwolfy:
The PR proposes to get rid of `call_origin` (i.e. falling back on numpy) in `dpnp.put` function.
Additionally, the safe casting of indices to `intp` dtype is added to the implementation.

The tests are updated to remove temporary workaround of dpctl#1382, because it has been resolved.
Also they are separated to `class TestPut` and extended with more test scenarios.


vlad-perevezentsev:
@antonwolfy 
1.It would be good if you could change `test_put...` in `test_indexing.py` by implementing them via `class TestPut` 
2. Should we support the case where `a` and `ind` is a bool data type?
3.Would it make sense to add tests for `test_sycl_queue` and `test_usm_type`?

antonwolfy:
> 1.It would be good if you could change `test_put...` in `test_indexing.py` by implementing them via `class TestPut`
> 2. Should we support the case where `a` and `ind` is a bool data type?
> 3.Would it make sense to add tests for `test_sycl_queue` and `test_usm_type`?

1. The existing tests are move under dedicated `class TestPut`. Additionally the test scenario is extended to cover more use cases.
2. The safe casting to `intp` dtype is added to the implementation. The tests are extended to cover that.
3. `dpnp.put` doesn't create any result array, but updating input array in-place. What kind of tests do you propose to add to `test_sycl_queue` and `test_usm_type`?

vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy

PR #1689, opened 2024-02-05
Add `gitleaks` and `shellcheck` to pre-commit configuration

antonwolfy:
The PR extends pre-commit hooks with `gitleaks` (a tool for detecting and preventing hardcoded secrets) and with `shellcheck` (a tool that gives warnings and suggestions for bash/sh shell scripts).
All detected issues were resolved.

- [x] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to issue with a reproducer?
- [ ] Have you tested your changes locally for CPU and GPU devices?
- [ ] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?

PR #1851, opened 2024-05-23
Remove skip in test_digitize

vlad-perevezentsev:
This PR suggests removing a previously skipped test in `TestDigitize` in  case when `shape =()` and `increasing=False` since [#1689](https://github.com/IntelPython/dpctl/issues/1689) is resolved and `dpctl` package with this fix is available in the internal channel. 


antonwolfy (approved):
Thank you @vlad-perevezentsev

PR #1855, opened 2024-05-24
Remove Python 3.8 from list of Programming Languages

antonwolfy:
The PR removes `Python 3.8` as a `Programming Language` from the meta data, because `dpnp` package for python 3.8 will no longer be provided.


antonwolfy:
> I think it would be a good idea to add a check for python version requirement in setup. @antonwolfy what do you think?

Yes, we can think about that. But it's definitely out of the scope for that PR.

vlad-perevezentsev (approved):
Thank you @antonwolfy

PR #1858, opened 2024-05-27
Use mamba in GitHub actions

antonwolfy:
The PR performs a transition to use of `mamba` in all affected GitHub actions.

The proposed change includes two workaround to known issues:
1. The step `conda-incubator/setup-miniconda` allows enabling `Mambaforge` and `mamba` by the dedicated options.
While based on `mamba` specification we have to limit channels by only `conda-forge`, excluding `defaults` channel. It is achieved inside the step by running `conda config --append channels conda-forge` command during the setup phase.
But due to known issue gh-12356, conda injects the `defaults` channel anyway, which we have to w/a explicitly by a separate call of `conda config --remove channels defaults` till the issue is resolved in either `conda` or `conda-incubator`.
2.  `mamba` has a low speed limit check which might sometimes fails due to connectivity issue, specially for a bunch of packages installed from `intel` channel.
To mitigate that we have to disable the check by setting `MAMBA_NO_LOW_SPEED_LIMIT=1` environment variable.


vtavana (approved):
Thank you, @antonwolfy!

PR #1877, opened 2024-06-12
Implement gesv_batch via gesv call

vlad-perevezentsev:
This PR suggests adding `gesv_batch` implementation as a separate function and provide iteration of the batch array within it to avoid allocation of each 2D array before calling `gesv` to improve performance and avoid serialization of all host tasks in the queue in MKL call;

Performance of `dpnp.solve` on Iris Xe
image


dpnp/backend/extensions/lapack/gesv.cpp
```diff
@@ -273,6 +273,162 @@ std::pair<sycl::event, sycl::event>
...
+            sycl::event gesv_ev =
+                gesv_fn(exec_q, n, nrhs, coeff_matrix_batch, lda,
+                        dependent_vals_batch, ldb, host_task_events, depends);
+
+            gesv_task_events.push_back(gesv_ev);
+        }
+    }
+
+    sycl::event combine_ev = exec_q.submit(
+        [&](sycl::handler &cgh) { cgh.depends_on(gesv_task_events); });
```
antonwolfy, line 424: What is the reason to combine them? Would not it be better to return a vector of events, like:
```cpp
return std::make_pair(args_ev, gesv_task_events);
```
I guess extra kernel submission might impact the performance.

dpnp/backend/extensions/lapack/gesv.hpp
```diff
@@ -39,11 +39,17 @@ namespace ext
...
 extern std::pair<sycl::event, sycl::event>
-    gesv(sycl::queue exec_q,
+    gesv(sycl::queue &exec_q,
          dpctl::tensor::usm_ndarray coeff_matrix,
          dpctl::tensor::usm_ndarray dependent_vals,
          const std::vector<sycl::event> &depends);
 
+extern std::pair<sycl::event, sycl::event>
+    gesv_batch(sycl::queue &exec_q,
+               dpctl::tensor::usm_ndarray coeff_matrix,
```
antonwolfy, line 49: Can it be const and passed by a reference?

dpnp/backend/extensions/lapack/gesv.cpp
```diff
@@ -273,6 +273,162 @@ std::pair<sycl::event, sycl::event>
...
+                "The batch_size of "
+                " coeff_matrix and dependent_vals must be"
+                " the same, but got " +
+                std::to_string(coeff_matrix_shape[2]) + " and " +
+                std::to_string(dependent_vals_shape[2]) + ".");
+        }
+    }
+
+    // check compatibility of execution queue and allocation queue
+    if (!dpctl::utils::queues_are_compatible(exec_q,
```
antonwolfy, line 340: Can we have common checks decoupled to single internal function to reduce code duplication?

antonwolfy, line 340: Do we need to add the checks like?
```cpp
dpctl::tensor::validation::CheckWritable::throw_if_not_writable(dst);
// ...
dpctl::tensor::validation::AmpleMemory::throw_if_not_ample(dst, src_nelems);
```

dpnp/backend/extensions/lapack/gesv.cpp
```diff
@@ -273,6 +273,162 @@ std::pair<sycl::event, sycl::event>
...
+                                             {coeff_matrix, dependent_vals}))
+    {
+        throw py::value_error(
+            "Execution queue is not compatible with allocation queues.");
+    }
+
+    auto const &overlap = dpctl::tensor::overlap::MemoryOverlap();
```
antonwolfy, line 347: Should we extend the check like did in dpctl for ufunc?
```cpp
    // check memory overlap
    auto const &overlap = dpctl::tensor::overlap::MemoryOverlap();
    auto const &same_logical_tensors =
        dpctl::tensor::overlap::SameLogicalTensors();
    if (overlap(src, dst) && !same_logical_tensors(src, dst)) {
        throw py::value_error("Arrays index overlapping segments of memory");
    }
```

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -267,101 +267,76 @@ def _batched_solve(a, b, exec_q, res_usm_type, res_type):
...
-        out_v = out_v.reshape(orig_shape_b)
-    return out_v
+    # Gesv call overwtires `b` in Fortran order, reorder the axes
+    # to match C order by moving the last axis to the front and
+    # reshape it back to the original shape of `b`.
+    v = dpnp.moveaxis(b_f, -1, 0).reshape(b_shape)
+
+    # dpnp.moveaxis can make the array non-contiguous if it is not 2D
+    if b.ndim > 2:
+        v = dpnp.ascontiguousarray(v)
```
antonwolfy, line 337: Is that only to be aligned with numpy? Do we need to add an explicit comment here? Otherwise the motivation isn't fully clear.

dpnp/backend/extensions/lapack/gesv.cpp
vlad-perevezentsev: We should not use this check cause `coeff_matrix` and `dependent_vals` may have different shapes

dpnp/backend/extensions/lapack/gesv.cpp
vlad-perevezentsev: Yes, I will add it
Thanks

dpnp/backend/extensions/lapack/gesv.hpp
vlad-perevezentsev: We can not pass `exec_q` as a constant argument because `mkl::lapack::gesv` expects queue as a regular argument
```
error: no matching function for call to 'gesv'
  145 |             gesv_event = mkl_lapack::gesv(
      |                          ^~~~~~~~~~~~~~~~
miniforge3/envs/test/include/oneapi/mkl/lapack/lapack.hpp:68:24: note: candidate function not viable: 1st argument ('const sycl::queue') would lose const qualifier
```

dpnp/linalg/dpnp_utils_linalg.py
vlad-perevezentsev: I updated this comment in the last commit

dpnp/backend/extensions/lapack/gesv.cpp
vlad-perevezentsev: The implementation has been changed.
Now `gesv` and `gesv_batch` return executing event and host event

dpnp/backend/extensions/lapack/gesv.cpp
vlad-perevezentsev: Done!

dpnp/backend/extensions/lapack/gesv.cpp
```diff
@@ -42,49 +43,150 @@ namespace mkl_lapack = oneapi::mkl::lapack;
 namespace py = pybind11;
 namespace type_utils = dpctl::tensor::type_utils;
 
-typedef sycl::event (*gesv_impl_fn_ptr_t)(sycl::queue,
+void common_gesv_checks(sycl::queue &exec_q,
```
antonwolfy, line 46: Can we move that to `gesv_common.hpp` or `gesv_common_utils.hpp` and make it inline rather than using extern specification.

dpnp/backend/extensions/lapack/gesv.cpp
```diff
@@ -42,49 +43,150 @@ namespace mkl_lapack = oneapi::mkl::lapack;
...
+                              std::to_string(max_dependent_vals_ndim) +
+                              "-dimensional array is expected.");
+    }
+
+    // The coeff_matrix and dependent_vals arrays must be F-contiguous arrays
+    // for gesv
+    // with the shapes (n,n) and (n,nrhs) or (n,) respectively;
+    // for gesv_batch
+    // with the shapes (n,n,batch_size) and (n,nrhs,batch_size) or
+    // (n,batch_size) respectively
```
antonwolfy, line 81: I believe it would be more readable with the spaces
```suggestion
    // with the shapes (n, n) and (n, nrhs) or (n, ) respectively;
    // for gesv_batch
    // with the shapes (n, n, batch_size) and (n, nrhs,batch_size) or
    // (n, batch_size) respectively
```

dpnp/backend/extensions/lapack/gesv.cpp
```diff
@@ -42,49 +43,150 @@ namespace mkl_lapack = oneapi::mkl::lapack;
...
+    if (scratchpad_size <= 0) {
+        throw std::runtime_error(
+            "Invalid scratchpad size: must be greater than zero."
+            "Calculated scratchpad size: " +
+            std::to_string(scratchpad_size));
+    }
+
     T *scratchpad = nullptr;
+    // Allocate memory for the scratchpad
+    scratchpad = sycl::malloc_device<T>(scratchpad_size, exec_q);
```
antonwolfy, line 174: It should be under `try/catch` block, because it can produce SYCL exception.

dpnp/backend/extensions/lapack/gesv_batch.cpp
```diff
@@ -0,0 +1,326 @@
...
+                                   char *in_b,
+                                   const std::vector<sycl::event> &depends)
+{
+    type_utils::validate_type_for_device<T>(exec_q);
+
+    T *a = reinterpret_cast<T *>(in_a);
+    T *b = reinterpret_cast<T *>(in_b);
+
+    std::int64_t a_size = n * n;
+    std::int64_t b_size = n * nrhs;
```
antonwolfy, line 73: Can it be const?
```suggestion
    const std::int64_t a_size = n * n;
    const std::int64_t b_size = n * nrhs;
```

dpnp/backend/extensions/lapack/gesv_batch.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+    const std::int64_t lda = std::max<size_t>(1UL, n);
+    const std::int64_t ldb = std::max<size_t>(1UL, n);
+
+    // Get the number of independent linear streams
+    std::int64_t n_linear_streams =
```
antonwolfy, line 79: And that one
```suggestion
    const std::int64_t n_linear_streams =
```

dpnp/backend/extensions/lapack/gesv_batch.cpp
```diff
@@ -0,0 +1,326 @@
...
+    // Calculate the total size needed for the pivot indices array for all
+    // linear streams with proper alignment
+    size_t alloc_ipiv_size =
+        helper::round_up_mult(n_linear_streams * n, padding);
+
+    // Allocate memory for the total pivot indices array
+    std::int64_t *ipiv =
+        sycl::malloc_device<std::int64_t>(alloc_ipiv_size, exec_q);
+    if (!ipiv)
+        throw std::runtime_error("Device allocation for ipiv failed");
```
antonwolfy, line 101: Should we created `alloc_ipiv` like we did for scratchpad?

dpnp/backend/extensions/lapack/gesv_batch.cpp
```diff
@@ -0,0 +1,326 @@
...
+                        << "Unexpected MKL exception caught during gesv() "
+                           "call:\nreason: "
+                        << e.what() << "\ninfo: " << e.info();
+                }
+            }
+            else {
+                error_msg << "Unexpected MKL exception caught during gesv() "
+                             "call:\nreason: "
+                          << e.what() << "\ninfo: " << e.info();
+            }
```
antonwolfy, line 187: I guess that is the same between `gesv` and `gesv_batch` and so can be separated to a helper call.

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -260,99 +260,78 @@ def _batched_solve(a, b, exec_q, res_usm_type, res_type):
...
-    if a.ndim > 3:
-        # get 3d input arrays by reshape
-        if a.ndim == b.ndim:
-            b = dpnp.reshape(b, (-1, b_shape[-2], b_shape[-1]))
-        else:
-            b = dpnp.reshape(b, (-1, b_shape[-1]))
+    # gesv_batch expects `a` to be a 3D array and
+    # `b` to be either a 2D or 3D array.
+    if a.ndim == b.ndim:
+        b = b.reshape(-1, b_shape[-2], b_shape[-1])
```
antonwolfy, line 269: I assume we can get an error here when b is `usm_ndarray` array.
Do you have a test on that?

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -260,99 +260,78 @@ def _batched_solve(a, b, exec_q, res_usm_type, res_type):
...
+    else:
+        b = b.reshape(-1, b_shape[-1])
 
-        a = dpnp.reshape(a, (-1, a_shape[-2], a_shape[-1]))
+    a = a.reshape(-1, a_shape[-2], a_shape[-1])
```
antonwolfy, line 273: The same question here if it is secured that `a` is always dpnp array? Is is covered by any test?

dpnp/linalg/dpnp_utils_linalg.py
vlad-perevezentsev: Yes.  we have a test to check these cases `test_usm_ndarray_linalg_batch` and it falls.
I changed it. Thank you

dpnp/backend/extensions/lapack/gesv_batch.cpp
vlad-perevezentsev: Sounds good. I have added `alloc_ipiv` to `common_helpers.hpp`. Thanks

dpnp/backend/extensions/lapack/gesv.cpp
vlad-perevezentsev: Done!

dpnp/backend/extensions/lapack/gesv_batch.cpp
vlad-perevezentsev: I have implemented `handle_lapack_exc` to `gesv_common_utils.hpp`.
Thanks

dpnp/backend/extensions/lapack/common_helpers.hpp
```diff
@@ -24,9 +24,11 @@
 //*****************************************************************************
 
 #pragma once
+#include <oneapi/mkl.hpp>
```
antonwolfy, line 27: It looks the include never used in that header file. Do we need to add it here?
Probably we need to include `sycl/sycl.hpp` instead? It looks missing.

dpnp/backend/extensions/lapack/common_helpers.hpp
```diff
@@ -63,4 +65,119 @@ inline bool check_zeros_shape(int ndim, const py::ssize_t *shape)
...
+
+    std::int64_t *ipiv = nullptr;
+
+    // Allocate memory for the total pivot indices array
+    try {
+        ipiv = sycl::malloc_device<std::int64_t>(alloc_ipiv_size, exec_q);
+        if (!ipiv)
+            throw std::runtime_error("Device allocation for ipiv failed");
+    } catch (sycl::exception const &e) {
+        throw std::runtime_error(
```
antonwolfy, line 114: Do we need to release a memory here?

dpnp/backend/extensions/lapack/common_helpers.hpp
```diff
@@ -63,4 +65,119 @@ inline bool check_zeros_shape(int ndim, const py::ssize_t *shape)
...
+            std::string(
+                "Unexpected SYCL exception caught during ipiv allocation: ") +
+            e.what());
+    }
+
+    return ipiv;
```
antonwolfy, line 120: Is it the same like:
```suggestion
    return alloc_ipiv(alloc_ipiv_size, exec_q);
```
The same is applicable to `alloc_scratchpad_batch` also.

vlad-perevezentsev: That is a good point!
I will do the same for scratchpad. Thank you

PR #1878, opened 2024-06-12
Rework implementation of `dpnp.fabs` function

antonwolfy:
The PR proposes to rework implementation of `dpnp.fabs` function which is quite limited now.

It assumes to have a separate `_ufunc_impl` pybind11 extension to aggravate all universal functions which are not provided by dpctl tensor implementation, but requires to be aligned with NumPy.

The new extension is implemented following similar approach as for `_vm_impl` extension and proposes to reuse existing `UnaryElementwiseFunc` functionality from dpctl tensor.


vtavana (approved):
Thank you, @antonwolfy!

PR #1883, opened 2024-06-14
Rework implementation of `dpnp.fmod` function

antonwolfy:
The PR proposes to rework implementation of `dpnp.fmod` function which is quite limited now.

There is new VM function is added to use `fmod` from OneMKL library where it's possible.
Also the PR adds new binary universal function `_fmod` to _ufunc_impl pybind11 extension which leverages on dpctl f/w.

The all affected tests are updated to cover different use cases. The scope of impacted third party tests is aligned with the latest content.


vlad-perevezentsev (approved):
Looks good!
Thank you @antonwolfy

PR #1567, opened 2023-09-21
Refactor MKL and oneDPL find package and add MKL_VERSION_2024 variable

vlad-perevezentsev:
This PR includes refactoring of `MKL` and `oneDPL` find_package logic and using `MKL_VERSION_2024 ` variable to target linking in `CMakeLists` files. 
It updates the logic to provide build compatibility with versions `2024.0.0` and `2023.2.0` for external and internal CI 

- [X] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to issue with a reproducer?
- [ ] Have you tested your changes locally for CPU and GPU devices?
- [ ] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?

PR #1889, opened 2024-06-18
Remove MKL_VERSION_2024 variable from cmake files

antonwolfy:
OneMKL 2024.0 is a minimum required version in `meta.yaml` as both build and run dependency.

This PR is an attempt to clean up the code and to remove a temporary solution implemented in #1567 for proper MKL discovery during the build process. As for now we should not consider a building with `OneMKL < 2024.0 version`.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy !

PR #1896, opened 2024-06-25
Remove the w/a which breaks the build on Windows with python `3.12`

antonwolfy:
Since python 3.12 `distutils` module was removed.
There was a workaround to force `distutils` enabling in case of `dpnp` build for Windows. But it is not required anymore, because the transition to `cmake` had been completed on Windows.

Thus the PR proposes to remove the workaround and to unblock the build with python `3.12` in internal CI.

Also that PR proposes to set a variable `OVERRIDE_INTEL_IPO` in public CI to override using interprocedural optimization in public CI to avoid insufficient resources failure during compilation on Windows.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #1897, opened 2024-06-26
Adopt dpnp interface to asynchronous dpctl execution (Part #1)

antonwolfy:
There were the changes in dpctl interface to make it asynchronous. It caused a number of dpnp tests failing.

On the first stage it is proposed to update dpnp and to adopt with that dpctl changes, but to avoid implementing any change in dpnp interface. There will be no asynchronous handling introduce in dpnp by that stage.

Even with the first stage there are a lot of the changes assumed, because dpnp has to reconsider usage of dpctl interface.
Thus the adaptation will be implemented through several independent PRs separated by dpnp functional blocks.

That PR propose to update all dpnp functions from array creation, elementwise, manipulation, counting and histogram functional blocks.

Even that PR is enough to stabilize the public CI and to have GitHub actions back to green status. It should be helpful and to unblock other PRs which are under review now.

The work on remaining dpnp functions will be continued by further PRs.


dpnp/dpnp_iface_arraycreation.py
```diff
@@ -2183,7 +2188,7 @@ def geomspace(
 
     """
 
-    return dpnp_geomspace(
+    res = dpnp_geomspace(
```
npolina4, line 2191: Input args have dpnp.ndarray type, but output arg has usm_ndarray type. Looks unclear.
Can we move this code inside dpnp_geomspace? Or use usm_ndarray as input type for dpnp_geomspace function?
```
dpnp.synchronize_array_data(res)
res = dpnp_array._create_from_usm_ndarray(res)
```

antonwolfy: I will update to return dpnp array.

vtavana (approved):
Thank you, @antonwolfy!

PR #1917, opened 2024-07-11
Build conda enviroment for coverage action without `intel` channel

antonwolfy:
TBD

### DIFF
--- a/.github/workflows/generate_coverage.yaml
+++ b/.github/workflows/generate_coverage.yaml
@@ -50,6 +50,7 @@ jobs:
         run: |
           sudo apt-get install intel-oneapi-compiler-dpcpp-cpp
           sudo apt-get install intel-oneapi-mkl-devel
+          sudo apt-get install intel-oneapi-libdpstd-devel
           sudo apt-get install intel-oneapi-tbb-devel
 
       - name: Install Lcov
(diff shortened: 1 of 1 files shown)


tests/third_party/cupy/fft_tests/test_fft.py
```diff
@@ -26,7 +26,7 @@ class TestFft(unittest.TestCase):
         rtol=1e-4,
         atol=1e-7,
         accept_error=ValueError,
-        type_check=has_support_aspect64(),
+        type_check=False,
```
vlad-perevezentsev, line 29: Would it be useful to add python version condition for this flag? 
For example
`False if sys.version_info.minor >= 11 else has_support_aspect64()`

antonwolfy: No, the behavior doesn't depend on python version. It depends on kind of MKL backend used by numpy.
In general, as an improvement, we can investigate how to fetch MKL backed used by numpy in the test. As an idea it might be done like:
```python
numpy.show_config('dicts')['Build Dependencies']["blas"]["name"]
```
But I'd propose to keep currently this out of scope of that PR.

vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy

PR #1918, opened 2024-07-11
Build conda enviroment for build docs action without `intel` channel

antonwolfy:
The conda channel `intel` is temporary inaccessible.

Due to that the PR attempts to unblock GH action with building dpnp documentation. It proposes to use only `dppy/label/dev` and `conda-forge` channels. While `onedpl-devel` and `tbb-devel` conda packages are assuming to be installed now from latest Intel OneAPI by installing `intel-oneapi-libdpstd-devel` and `intel-oneapi-tbb-devel` respectively.

The new behavior will be controlled by action environment variable `INSTALL_ONE_API`. It is implemented in that way to provide a fast solution to switch back to `intel` channel approach.

### DIFF
--- a/.github/workflows/build-sphinx.yml
+++ b/.github/workflows/build-sphinx.yml
@@ -160,7 +160,9 @@ jobs:
         working-directory: doc
 
       - name: Build docs
-        run: make html
+        run: |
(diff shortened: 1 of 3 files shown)


vtavana (approved):
Thank you, @antonwolfy!

PR #1932, opened 2024-07-16
Align with `dpt.where` which support scalar as input

antonwolfy:
The PR aligns with the recent change in dpctl ([dpctl-1719](https://github.com/IntelPython/dpctl/pull/1719)) where support for Python scalars added to `dpt.where` function.
This way dpnp can fully rely on dpctl implementation when it receives the scalars.


vtavana (approved):
Thank you, @antonwolfy!

PR #1936, opened 2024-07-18
Implement `syevd_batch` and `heevd_batch` 

vlad-perevezentsev:
This PR is a duplicate and continues the work of #1867  . Due to the change in `syevd`/`heevd` function interface #1891 this PR is a quick way to avoid conflicts during rebasing.

This PR suggests adding `syevd_batch` and `heevd_batch` implementation as a separate function and provide iteration of the batch array within it to avoid allocation of each 2D array before calling `syevd` and `heevd` to improve performance and avoid serialization of all host tasks in the queue in MKL call;


dpnp/backend/extensions/lapack/evd_batch_common.hpp
```diff
@@ -0,0 +1,174 @@
...
+        static_cast<oneapi::mkl::uplo>(upper_lower);
+
+    sycl::event evd_batch_ev =
+        evd_batch_fn(exec_q, jobz_val, uplo_val, batch_size, n, eig_vecs_data,
+                     eig_vals_data, depends);
+
+    sycl::event ht_ev = dpctl::utils::keep_args_alive(
+        exec_q, {eig_vecs, eig_vals}, {evd_batch_ev});
+
+    return std::make_pair(ht_ev, evd_batch_ev);
```
antonwolfy, line 160: In reality `evd_batch_ev` is a host task. Would it be safe to return it by the 2nd argument here, because it can be passed later as dependent event to a function from other extension or any function which accepts `depends` keyword? Would it require any additional action, like `py::gil_scoped_release release;` before passing it as dependent event to avoid a dead lock, or it's absolutely safe?

dpnp/backend/extensions/lapack/heevd_batch.cpp
```diff
@@ -0,0 +1,207 @@
...
+    if (scratchpad_size > 0) {
+        // Calculate the total scratchpad memory size needed for all linear
+        // streams with proper alignment
+        size_t alloc_scratch_size =
+            helper::round_up_mult(n_linear_streams * scratchpad_size, padding);
+
+        // Allocate memory for the total scratchpad
+        scratchpad = sycl::malloc_device<T>(alloc_scratch_size, exec_q);
+        if (!scratchpad)
+            throw std::runtime_error("Device allocation for scratchpad failed");
```
antonwolfy, line 76: Do we need to have similar check in `heevd_impl` if memory allocation for scratchpad returns `nullptr`?

dpnp/backend/extensions/lapack/heevd_batch.cpp
```diff
@@ -0,0 +1,207 @@
...
+                                    char *in_a,
+                                    char *out_w,
+                                    const std::vector<sycl::event> &depends)
+{
+    type_utils::validate_type_for_device<T>(exec_q);
+    type_utils::validate_type_for_device<RealT>(exec_q);
+
+    T *a = reinterpret_cast<T *>(in_a);
+    RealT *w = reinterpret_cast<RealT *>(out_w);
+
```
antonwolfy, line 49: Do we need to have any special handling in case of `batch_size == 0`?
Is it possible that `batch_size == 0` passed into `heevd_batch_impl`?

dpnp/backend/extensions/lapack/heevd_batch.cpp
```diff
@@ -0,0 +1,207 @@
...
+                             // triangular part of A, or the lower triangular
+                             // otherwise
+                n,           // The order of the matrix A (0 <= n)
+                a_batch,     // Pointer to the square A (n x n)
+                             // If 'jobz == job::vec', then on exit it will
+                             // contain the eigenvectors of A
+                lda, // The leading dimension of A, must be at least max(1, n)
+                w_batch, // Pointer to array of size at least n, it will contain
+                         // the eigenvalues of A in ascending order
+                current_scratch_heevd, // Pointer to scratchpad memory to be
```
antonwolfy, line 119: As I can see it's possible to have `current_scratch_heevd == nullptr` while passing here. Is it a valid use case from MKL interface?

dpnp/backend/extensions/lapack/heevd_batch.hpp
```diff
@@ -0,0 +1,38 @@
...
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//*****************************************************************************
+
+#pragma once
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include "evd_batch_common.hpp"
```
antonwolfy, line 31: That include isn't used here and would be better to move it to `heevd_batch.cpp` where it is used.

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -96,85 +96,74 @@ def _batched_eigh(a, UPLO, eigen_mode, w_type, v_type):
...
-            a_sycl_queue,
-            jobz,
-            uplo,
-            eig_vecs[i].get_array(),
-            w[i].get_array(),
-            depends=[copy_ev],
-        )
-        _manager.add_event_pair(ht_ev, lapack_ev)
+    # get 2d dpnp array with eigenvalues by reshape
+    w = w.reshape(-1, w_orig_shape[-1])
```
antonwolfy, line 144: What is the reason `w` isn't allocated with expected shape at the beginning at line#140? Why do we need to have an extra step with reshape?

dpnp/backend/extensions/lapack/heevd_batch.cpp
vlad-perevezentsev: I think this check is unnecessary because in python we have `a.size == 0` check and `batch_size=0` can not be passed to `heevd_batch_impl or syevd_batch_impl`

dpnp/backend/extensions/lapack/heevd_batch.hpp
vlad-perevezentsev: I changed it. Thank you

dpnp/linalg/dpnp_utils_linalg.py
vlad-perevezentsev: This has been changed. Thank you

dpnp/backend/extensions/lapack/evd_batch_common.hpp
```diff
@@ -0,0 +1,174 @@
...
+#pragma once
+
+#include <oneapi/mkl.hpp>
+#include <pybind11/pybind11.h>
+
+// dpctl tensor headers
+#include "utils/memory_overlap.hpp"
+#include "utils/output_validation.hpp"
+#include "utils/type_dispatch.hpp"
+#include "utils/type_utils.hpp"
```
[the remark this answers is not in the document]
vlad-perevezentsev: I changed it. Thank you

dpnp/backend/extensions/lapack/heevd_batch.cpp
vlad-perevezentsev: I have added additional checks for this. Thank you

dpnp/backend/extensions/lapack/heevd_batch.cpp
vlad-perevezentsev: I have checked `heevd_batch_imp`l and `syevd_batch_impl` with `batch_size=0` and they work correctly so I do not think an additional check is needed for this case.

antonwolfy: We no need to allocate memory for scratchpad in case of `batch_size == 0`

dpnp/backend/extensions/lapack/evd_batch_common.hpp
```diff
@@ -62,25 +60,18 @@ std::pair<sycl::event, sycl::event>
...
     const py::ssize_t *eig_vecs_shape = eig_vecs.get_shape_raw();
     const py::ssize_t *eig_vals_shape = eig_vals.get_shape_raw();
 
-    if (eig_vecs_shape[0] != eig_vecs_shape[1]) {
-        throw py::value_error("The batch shape of 'eig_vecs' must be square.");
-    }
-    else if (eig_vecs_shape[2] != eig_vals_shape[0] ||
-             eig_vecs_shape[0] != eig_vals_shape[1])
+    const int expected_eig_vecs_nd = 3;
+    const int expected_eig_vals_nd = 2;
```
antonwolfy, line 68:
```suggestion
    constexpr int expected_eig_vecs_nd = 3;
    constexpr int expected_eig_vals_nd = 2;
```

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -96,85 +96,72 @@ def _batched_eigh(a, UPLO, eigen_mode, w_type, v_type):
...
-    a_order = "C" if a.flags.c_contiguous else "F"
+
+    a_orig_shape = a.shape
+    a_orig_order = "C" if a.flags.c_contiguous else "F"
+    # get 3d input array by reshape
+    a = dpnp.reshape(a, (-1, a_orig_shape[-2], a_orig_shape[-1]))
+    a_new_shape = a.shape
 
     _manager = dpu.SequentialOrderManager[a_sycl_queue]
+    dep_evs = _manager.submitted_events
```
antonwolfy, line 124: Please move these lines closer o memory allocation at line#131, just in case if in future `dpnp.moveaxis` will produce a dependent event.

dpnp/backend/extensions/lapack/evd_batch_common.hpp
```diff
@@ -0,0 +1,138 @@
...
+    sycl::event ht_ev = dpctl::utils::keep_args_alive(
+        exec_q, {eig_vecs, eig_vals}, {evd_batch_ev});
+
+    return std::make_pair(ht_ev, evd_batch_ev);
+}
+
+template <typename dispatchT,
+          template <typename fnT, typename T, typename RealT>
+          typename factoryT>
+void init_evd_batch_dispatch_table(
```
antonwolfy, line 130: The init functions in batch and non-batch implementations looks exactly the same.

dpnp/backend/extensions/lapack/evd_batch_common.hpp
```diff
@@ -0,0 +1,138 @@
...
+    size_t src_nelems(1);
+
+    for (int i = 0; i < eig_vecs_nd; ++i) {
+        src_nelems *= static_cast<size_t>(eig_vecs_shape[i]);
+    }
+
+    if (src_nelems == 0) {
+        // nothing to do
+        return std::make_pair(sycl::event(), sycl::event());
+    }
```
antonwolfy, line 92: That part is also common, can we move it to separate function, like:
```cpp
    if (get_src_nelems(eig_vecs_nd, eig_vecs_shape) == 0) {
        // nothing to do
        return std::make_pair(sycl::event(), sycl::event());
    }
```

dpnp/backend/extensions/lapack/heevd_batch.cpp
```diff
@@ -0,0 +1,219 @@
...
+        scratchpad = sycl::malloc_device<T>(alloc_scratch_size, exec_q);
+        if (!scratchpad)
+            throw std::runtime_error("Device allocation for scratchpad failed");
+    }
+    else {
+        throw std::runtime_error(
+            "Invalid scratchpad size: must be greater than zero."
+            "Calculated scratchpad size: " +
+            std::to_string(scratchpad_size));
+    }
```
antonwolfy, line 89: Can we move that to a separate function in `evd_btach_common.hpp`, to have like:
```cpp
T *scratchpad = alloc_scratchpad(scratchpad_size, n_linear_streams, exec_q);
```

vlad-perevezentsev: done

dpnp/backend/extensions/lapack/evd_batch_common.hpp
vlad-perevezentsev: I have added `check_zeros_shape` function.
Thanks

dpnp/backend/extensions/lapack/evd_batch_common.hpp
vlad-perevezentsev: Done
Thanks!

dpnp/linalg/dpnp_utils_linalg.py
vlad-perevezentsev: Done.
Thanks!

dpnp/backend/extensions/lapack/evd_batch_common.hpp
```diff
@@ -124,15 +119,33 @@ std::pair<sycl::event, sycl::event>
     return std::make_pair(ht_ev, evd_batch_ev);
 }
 
-template <typename dispatchT,
-          template <typename fnT, typename T, typename RealT>
-          typename factoryT>
-void init_evd_batch_dispatch_table(
-    dispatchT evd_batch_dispatch_table[][dpctl_td_ns::num_types])
+template <typename T>
+T *alloc_scratchpad(std::int64_t scratchpad_size,
```
antonwolfy, line 123: Should we add `inline` here?

dpnp/backend/extensions/lapack/heevd_batch.cpp
vlad-perevezentsev: My check `helper::check_zeros_shape` in `evd_batch_func` ensures that `batch_size` and `n` are non-zero and returns empty events otherwise.
I added a comment with this and I think the additional checks in `syevd_batch_impl` and `heevd_batch_impl` do not make sense

antonwolfy (approved):
Thank you @vlad-perevezentsev . It's really nice changes which can be replicated to other batch functions and which help to get rid of limitations we previously had there.

PR #1940, opened 2024-07-22
Resolve a possible race condition in dpnp.inv() implemenetation

antonwolfy:
The PR proposes to implement a fix preventing possible race condition inside `dpnp.inv` function.

It assumes `_gesv` call depends on both copy event and event produced above `dpnp.eye` call and so it is mandatory to pass `depends=_manager.submitted_events` there.

The issue was caught during the analysis of failed tests in GitHub actions. No need to implement any dedicated test to cover that use case.


vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy for finding and fixing it

PR #1947, opened 2024-07-25
Add conditional run of test which fails with older numpy

antonwolfy:
The PR implements a conditional run of `tests/test_mathematical.py::TestLogSumExp::test_logsumexp_out_dtype` test which fails with `numpy < 1.26.4` due to exception raised:
```python
exp = numpy.logaddexp.reduce(a, out=out, axis=1)
E   OverflowError: cannot convert float infinity to integer
```

The PR proposes to run the test only with `numpy >= 1.26.4` which contains a fix for that.


vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy !

PR #1948, opened 2024-07-25
Reuse `dpctl.tensor._numpy_helper` module

antonwolfy:
As part of the work to support numpy 2.0 we need to align with new import of `normalize_axis_index`, `normalize_axis_tuple` and `AxisError` which was changed in non-backward compatible way since numpy 2.0.

That PR proposes to use the same functionality but from `dpctl.tensor._numpy_helper` module which already adopted to support different numpy versions (including numpy 2.0).


vtavana (approved):
Thank you, @antonwolfy!

PR #1964, opened 2024-08-02
Enable third party tests for `dpnp.ndarray`

antonwolfy:
The PR is about to enable scope of third party tests intended to validate `dpnp.ndarray` constructor.

The PR also proposes some code changes to pass the new test scope and be aligned with NumPy behavior:
- To use only dpctl.tensor constructor for `usm_ndarray` instead of `dpt.asaray` call in code of dpnp array constructor
- To use  `dpnp_array._create_from_usm_ndarray` function in implementation of dpnp container
- To allow passing an integer scalar to shape setter
- To prohibit extending dpnp array with any custom attribute 

The configuration of pre-commit `pylint` hook is extended to disable `c-extension-no-member` check, because it produces only false-positive error in dpnp code. It was already previously disable per specific dpnp interface files, but not moved to global `pylint` configuration.


vlad-perevezentsev (approved):
Looks good
Thank you @antonwolfy

PR #1966, opened 2024-08-02
Implement `dpnp.nan_to_num()`

vlad-perevezentsev:
This PR suggests implementing new math function `dpnp.nan_to_num()` through dpnp functions like `dpnp.isposinf()`, `dpnp.isneginf()` `dpnp.isnan()` and `dpnp.where()`. Also added new tests covering cases not implemented in CuPy tests

Performance measurements on PVC 
image


dpnp/dpnp_iface_mathematical.py
```diff
@@ -2254,6 +2262,119 @@ def modf(x1, **kwargs):
...
+        Value to be used to fill negative infinity values. If no value is
+        passed then negative infinity values will be replaced with a very
+        small (or negative) number.
+        Default: ``None``.
+
+    Returns
+    -------
+    out : dpnp.ndarray
+        `x`, with the non-finite values replaced. If `copy` is False, this may
+        be `x` itself.
```
antonwolfy, line 2312:
```suggestion
        `x`, with the non-finite values replaced. If `copy` is ``False``, this
        may be `x` itself.
```

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2254,6 +2262,119 @@ def modf(x1, **kwargs):
...
+    ----------
+    x : {dpnp.ndarray, usm_ndarray}
+        Input data.
+    copy : bool, optional
+        Whether to create a copy of `x` (True) or to replace values
+        in-place (False). The in-place operation only occurs if
+        casting to an array does not require a copy.
+        Default: ``True``.
+    nan : {int, float}, optional
+        Value to be used to fill NaN values.
```
antonwolfy, line 2295:
```suggestion
        Value to be used to fill ``NaN`` values.
```

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2254,6 +2262,119 @@ def modf(x1, **kwargs):
...
+    Parameters
+    ----------
+    x : {dpnp.ndarray, usm_ndarray}
+        Input data.
+    copy : bool, optional
+        Whether to create a copy of `x` (True) or to replace values
+        in-place (False). The in-place operation only occurs if
+        casting to an array does not require a copy.
+        Default: ``True``.
+    nan : {int, float}, optional
```
antonwolfy, line 2294: The below example works:
```python
import dpnp as np
np.nan_to_num(np.array(np.nan), nan=np.array(0))
# Out: array(0.)
```
but is it suppose to be working?
Based on the parameter description, only scalars are valid.
And do we need an explicit check for `None` value for `nan` keyword?

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2254,6 +2262,119 @@ def modf(x1, **kwargs):
...
+           0.00000000e+000 +0.00000000e+000j,
+           0.00000000e+000 +1.79769313e+308j])
+    >>> np.nan_to_num(y, nan=111111, posinf=222222)
+    array([222222.+111111.j, 111111.     +0.j, 111111.+222222.j])
+
+    """
+
+    dpnp.check_supported_arrays_type(x)
+
+    x = dpnp.array(x, copy=copy)
```
antonwolfy, line 2351:
```suggestion
    if copy:
        x = dpnp.empty_like(x)
```

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2254,6 +2262,119 @@ def modf(x1, **kwargs):
...
+    for part in parts:
+        nan_mask = dpnp.isnan(part)
+        posinf_mask = dpnp.isposinf(part)
+        neginf_mask = dpnp.isneginf(part)
+
+        part = dpnp.where(nan_mask, nan, part, out=part)
+        part = dpnp.where(posinf_mask, max_f, part, out=part)
+        part = dpnp.where(neginf_mask, min_f, part, out=part)
+
+    return x
```
antonwolfy, line 2375: Could you please a performance comparison table to the PR?
It looks there is some place for improvement in current implementation if requires.

tests/test_mathematical.py
```diff
@@ -1445,6 +1445,40 @@ def test_power_scalar(shape, dtype):
     assert_allclose(result, expected, rtol=1e-6)
 
 
+class TestNanToNum:
```
antonwolfy, line 1448: please move new class before `class TestNextafter` to keep lexicographical order

tests/test_mathematical.py
```diff
@@ -1445,6 +1445,40 @@ def test_power_scalar(shape, dtype):
...
+    @pytest.mark.parametrize("dtype", get_all_dtypes())
```
antonwolfy, line 1449:
```suggestion
    @pytest.mark.parametrize("dtype", get_all_dtypes(no_none=True))
```

tests/test_mathematical.py
antonwolfy, line 1448: please add tests to `test_sycl_queue` and `test_usm_type` scope, specially with `copy=True` use case.

dpnp/dpnp_iface_mathematical.py
vlad-perevezentsev: We can't create an empty copy
Did you mean that?
```
if copy:
    x = dpnp.array(x)
```

antonwolfy: No, I mean we do not really need to make a copy. We can allocate memory for `out` array:
```python
out = dpnp.empty_like(x) if copy else x
```
and to pass it later to `dpnp.where(..., out=part_out)` in the loop.

antonwolfy: That way we can avoid launching a copy kernel.

dpnp/dpnp_iface_mathematical.py
vlad-perevezentsev: I have added additional checks for `nan`, `posinf`, `neginf` args.
Thanks

dpnp/dpnp_iface_mathematical.py
vlad-perevezentsev: Your suggestion does improve performance. Thank you

tests/test_sycl_queue.py
```diff
@@ -2329,3 +2329,20 @@ def test_astype(device_x, device_y):
...
+)
+def test_nan_to_num(copy, device):
+    a = dpnp.array([-dpnp.nan, -1, 0, 1, dpnp.nan], device=device)
+    result = dpnp.nan_to_num(a, copy=copy)
+
+    assert_sycl_queue_equal(result.sycl_queue, a.sycl_queue)
+    if copy:
+        assert result is not a
+    else:
+        assert result is a
```
antonwolfy, line 2348: I guess it can be shorter:
```suggestion
    assert copy == (result is not a)
```

tests/third_party/cupy/math_tests/test_misc.py
```diff
@@ -260,27 +261,31 @@ def test_nan_to_num_inf_arg(self):
...
+    @pytest.mark.skip(reason="nan, posinf, neginf as array are not supported")
     @pytest.mark.parametrize("kwarg", ["nan", "posinf", "neginf"])
     def test_nan_to_num_broadcast(self, kwarg):
         for xp in (numpy, cupy):
-            x = xp.asarray([0, 1, xp.nan, 4], dtype=xp.float64)
-            y = xp.zeros((2, 4), dtype=xp.float64)
-            with pytest.raises(ValueError):
+            x = xp.asarray([0, 1, xp.nan, 4], dtype=cupy.default_float_type())
+            y = xp.zeros((2, 4), dtype=cupy.default_float_type())
+            with pytest.raises(TypeError):
```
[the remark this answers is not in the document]
vlad-perevezentsev: Done
Thanks

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2254,6 +2262,134 @@ def modf(x1, **kwargs):
...
+
+    Parameters
+    ----------
+    x : {dpnp.ndarray, usm_ndarray}
+        Input data.
+    copy : bool, optional
+        Whether to create a copy of `x` (``True``) or to replace values
+        in-place (``False``). The in-place operation only occurs if casting to
+        an array does not require a copy.
+    nan : {int, float}, optional
```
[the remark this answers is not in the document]
vlad-perevezentsev: Sound good.
I have added error throwing if types are not `int` and `float`.
Thank you!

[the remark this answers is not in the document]
antonwolfy: But why don't you add support of bool scalar? (instead of raising `TypeError` as for complex)
For complex scalar the function will not work, but it can work with bool scalar. What was the reason to limit that?

PR #1967, opened 2024-08-02
Implement docs for ufunc. Update docs for logic element-wise functions.

npolina4:

### DIFF
--- a/doc/reference/ufunc.rst
+++ b/doc/reference/ufunc.rst
@@ -5,7 +5,49 @@ Universal Functions (ufunc)
 
 .. https://docs.scipy.org/doc/numpy/reference/ufuncs.html
 
-DPNP provides universal functions (a.k.a. ufuncs) to support various element-wise operations.
+A universal function (or ufunc for short) is a function that operates on ndarrays in an element-by-element fashion, \
+supporting array broadcasting, type casting, and several other standard features. That is, a ufunc is a “vectorized” \
+wrapper for a function that takes a fixed number of specific inputs and produces a fixed number of specific outputs. \
+For full documentation refer to :obj:`numpy.ufunc`.
+
+ufuncs
+------
+.. autosummary::
+   :toctree: generated/
+
+   dpnp.ufunc
+
+Attributes
+~~~~~~~~~~
+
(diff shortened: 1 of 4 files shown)


doc/reference/ufunc.rst
```diff
@@ -5,7 +5,49 @@ Universal Functions (ufunc)
 
 .. https://docs.scipy.org/doc/numpy/reference/ufuncs.html
 
-DPNP provides universal functions (a.k.a. ufuncs) to support various element-wise operations.
+A universal function (or ufunc for short) is a function that operates on ndarrays in an element-by-element fashion, \
```
antonwolfy, line 8: Why do we need `\` here? Is it helps somehow to format the output? (and in what way?)
I don't see that we used it previously somewhere and so wonder why do we need it here.

doc/reference/ufunc.rst
```diff
@@ -5,7 +5,49 @@ Universal Functions (ufunc)
...
+supporting array broadcasting, type casting, and several other standard features. That is, a ufunc is a “vectorized” \
+wrapper for a function that takes a fixed number of specific inputs and produces a fixed number of specific outputs. \
+For full documentation refer to :obj:`numpy.ufunc`.
+
+ufuncs
```
antonwolfy, line 13: Can we put here a link to :class:`ufunc` like in numpy with more detailed description of the attributes?

doc/reference/ufunc.rst
```diff
@@ -5,7 +5,49 @@ Universal Functions (ufunc)
...
+------
+.. autosummary::
+   :toctree: generated/
+
+   dpnp.ufunc
+
```
antonwolfy, line 19: What if about `Optional keyword arguments` section present in numpy? Should we add it also?

doc/reference/ufunc.rst
```diff
@@ -5,7 +5,49 @@ Universal Functions (ufunc)
...
+Attributes
+~~~~~~~~~~
+
```
antonwolfy, line 22: Do we need to specify `index` directive? (applicable to different place on the page)
```rst
.. index::
   pair: ufunc; attributes
```

PR #1969, opened 2024-08-05
Leverage on dpctl.tensor implementation in `dpnp.take_along_axis`

antonwolfy:
The PR proposes to leverage on dpctl.tensor implementation in `dpnp.take_along_axis` function.
Additionally, the function signature is extended with new keyword `mode` to provide a control how out-of-bounds indices will be handled.


dpnp/dpnp_iface_indexing.py
```diff
@@ -1579,12 +1588,21 @@ def take_along_axis(a, indices, axis):
...
+        dpnp.check_supported_arrays_type(indices)
+        if indices.ndim != 1:
+            raise ValueError(
+                "when axis=None, `indices` must have a single dimension."
+            )
 
-    return a[_build_along_axis_index(a, indices, axis)]
+        a = dpnp.ravel(a)
+        axis = 0
+
```
vlad-perevezentsev, line 1600: Does it make sense to handle cases like `isinstance(axis, int)` and `if mode not in ("wrap", "clip")` and raise an error from dpnp or raising the error from dpctl is ok?

dpnp/dpnp_iface_indexing.py
```diff
@@ -1511,15 +1511,24 @@ def take_along_axis(a, indices, axis):
         Indices to take along each 1d slice of `a`. This must match the
         dimension of the input array, but dimensions ``Ni`` and ``Nj``
         only need to broadcast against `a`.
-    axis : int
+    axis : {None, int}
```
vlad-perevezentsev, line 1514: Could you add a colon for `axis` in `def put` docstrings  like `axis : {None, int}, optional`

antonwolfy: I don't feel I got the comment.
The `axis` here is not optional, so we can to change to `axis : {None, int}, optional`.
It wasn't implemented as an optional keyword, because in numpy it is not the case.

dpnp/dpnp_iface_indexing.py
antonwolfy: All the error handling is already implemented in dpctl. I don't think we need to duplicate it in dpnp also.

dpnp/dpnp_iface_indexing.py
antonwolfy: Do you mean to make `axis` keyword as optional one?
It was not implemented in that way, because it is not the case for numpy also. I'd propose to keep the behavior aligned.

vlad-perevezentsev: I meant that `put` documentation for axis parameter there is no colon `:`
This does not apply to your changes here, just noticed it

dpnp/dpnp_iface_indexing.py
vlad-perevezentsev: Ok

dpnp/dpnp_iface_indexing.py
antonwolfy: I see, updated.

vlad-perevezentsev (approved):
Looks good!
Thank you @antonwolfy

PR #1983, opened 2024-08-13
Update `dpnp.ediff1d` function

vlad-perevezentsev:
This PR suggests updating the` dpnp.ediff1d()` implementation to make it more readable and adding a new `_process_ediff1d_args` helper function to reduce duplicate code

Additionally it updates tests to unlock the internal CI cover the new helper function.

### DIFF
--- a/dpnp/dpnp_iface_mathematical.py
+++ b/dpnp/dpnp_iface_mathematical.py
@@ -308,6 +308,30 @@ def _gradient_num_diff_edges(
         )
 
 
+def _process_ediff1d_args(arg, arg_name, ary_dtype, ary_sycl_queue, usm_type):
+    """Process the argument for ediff1d."""
+    if not dpnp.is_supported_array_type(arg):
+        arg = dpnp.asarray(arg, usm_type=usm_type, sycl_queue=ary_sycl_queue)
+    else:
+        usm_type = dpu.get_coerced_usm_type([usm_type, arg.usm_type])
+        # check that arrays have the same allocation queue
+        if dpu.get_execution_queue([ary_sycl_queue, arg.sycl_queue]) is None:
+            raise ValueError(
+                f"ary and {arg_name} must be allocated on the same SYCL queue"
+            )
+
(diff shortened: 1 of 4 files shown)


tests/test_mathematical.py
```diff
@@ -2151,6 +2151,14 @@ def test_ediff1d_errors(self):
         to_end = dpnp.array([5], dtype="f4")
         assert_raises(TypeError, dpnp.ediff1d, a_dp, to_end=to_end)
 
+        # another `to_begin` sycl queue
+        to_begin = dpnp.array([-20, -15], sycl_queue=dpctl.SyclQueue())
```
antonwolfy, line 2155: Shouldn't that tests be part of `test_sycl_queue.py` scope?

dpnp/dpnp_iface_mathematical.py
```diff
@@ -308,6 +308,30 @@ def _gradient_num_diff_edges(
...
 
+def _process_ediff1d_args(arg, arg_name, ary_dtype, ary_sycl_queue, usm_type):
+    """Process the argument for ediff1d."""
+    if not dpnp.is_supported_array_type(arg):
+        arg = dpnp.asarray(arg, usm_type=usm_type, sycl_queue=ary_sycl_queue)
+    else:
+        usm_type = dpu.get_coerced_usm_type([usm_type, arg.usm_type])
+        # check that arrays have the same allocation queue
+        if dpu.get_execution_queue([ary_sycl_queue, arg.sycl_queue]) is None:
+            raise ValueError(
```
antonwolfy, line 319: It should be `ExecutionPlacementError` exception raised

tests/test_mathematical.py
vlad-perevezentsev: I would leave it here because `test_sycl_queue.py` is used to test function works with only valid queues. 
We do not check there errors with different queues

antonwolfy (approved):
Thank you @vlad-perevezentsev

PR #1987, opened 2024-08-15
Removed sycl extensions backend

npolina4:
Removed sycl extensions backend.
Using dpctl.tensor.sum for all cases.
Improved build speed


tests/test_extensions.py
```diff
@@ -1,252 +0,0 @@
-from itertools import product
```
antonwolfy, line 1: Please remove `test_extensions.py` from scope of public CI

antonwolfy (approved):
Thank you @npolina4

PR #2002, opened 2024-08-20
Add implementation of `dpnp.real_if_close`

antonwolfy:
The PR proposes to implement `dpnp.real_if_close` function.

The implementation is done through existing python calls.
Existing third party tests are unmuted. And new tests are added to cover different use cases.

### DIFF
--- a/dpnp/dpnp_iface_mathematical.py
+++ b/dpnp/dpnp_iface_mathematical.py
@@ -3081,6 +3081,7 @@ def real_if_close(a, tol=100):
     tol : scalar, optional
         Tolerance in machine epsilons for the complex part of the elements in
         the array. If the tolerance is <=1, then the absolute tolerance is used.
+        Default: ``100``.
 
     Returns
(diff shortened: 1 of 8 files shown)


dpnp/dpnp_iface_mathematical.py
```diff
@@ -3064,6 +3065,67 @@ def prod(
...
+
+    Returns
+    -------
+    out : dpnp.ndarray
+        If `a` is real, the type of `a` is used for the output. If `a` has
+        complex elements, the returned type is float.
+
+    See Also
+    --------
+    real, imag, angle
```
vlad-perevezentsev, line 3094: Could you expand the description of these functions to follow the general style?
Like this:
```
:obj:`dpnp.real` : The real part of the array.
...
```

antonwolfy: Done. Thank you for noticing that.

PR #2055, opened 2024-09-17
Implement asynchronous `fill` method using `dpctl` kernels

ndgrigorian:
This PR proposes a change to `dpnp_array.fill` method which leverages dpctl kernels to make `fill` asynchronous and more efficient, avoiding repeated calls to index the array and copying scalars to the device for each element.

Shows significant performance gains on Iris Xe in WSL

Before
```
In [1]: import dpnp as dnp

In [2]: x_dnp = dnp.empty(10000, dtype="c8")

In [3]: %timeit x_dnp.fill(10)
1.25 s ± 47.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [4]: %timeit x_dnp.fill(10)
1.26 s ± 27.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

After
```
In [8]: %timeit x_dnp.fill(10); q.wait()
229 μs ± 37.8 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

- [X] If this PR is a work in progress, are you filing the PR as a draft?


ndgrigorian:
@antonwolfy @vtavana @vlad-perevezentsev 
I've added a commit to skip `test_fill_with_numpy_scalar_ndarray` from the CuPy tests.

`dpnp.fill_diagonal` does not permit NumPy arrays, so it made sense to do the same here. If it would be preferred to keep this feature, it can be implemented, but I would argue that for consistency, the two should behave the same.

tests/third_party/cupy/core_tests/test_ndarray_copy_and_view.py
```diff
@@ -274,9 +274,10 @@ def test_fill(self, xp, dtype):
...
         return a
 
-    @testing.with_requires("numpy>=1.24.0")
-    @testing.for_all_dtypes_combination(("dtype1", "dtype2"))
-    @testing.numpy_cupy_array_equal(accept_error=ComplexWarning)
+    @pytest.mark.skip(
+        "asynchronous fill does not allow numpy scalar array for consistency"
+        "with dpnp.fill_diagonal"
+    )
     def test_fill_with_numpy_scalar_ndarray(self, xp, dtype1, dtype2):
```
antonwolfy, line 281: Could you please update `test_fill_with_nonscalar_ndarray` test to remove skip there? (need to accept both exception type: TypeError and ValueError)

dpnp/dpnp_array.py
```diff
@@ -903,8 +903,10 @@ def fill(self, value):
```
antonwolfy, line 903: Could you please update docstring a bit, like:
```python
        """
        Fill the array with a scalar value.

        For full documentation refer to :obj:`numpy.ndarray.fill`.

        Parameters
        ----------
        value : {dpnp.ndarray, usm_ndarray, scalar}
            All elements of `a` will be assigned this value.

        Examples
        --------
        >>> import numpy as np
        >>> a = np.array([1, 2])
        >>> a.fill(0)
        >>> a
        array([0, 0])
        >>> a = np.empty(2)
        >>> a.fill(1)
        >>> a
        array([1.,  1.])

        """
```

dpnp/dpnp_algo/dpnp_fill.py
```diff
@@ -0,0 +1,85 @@
...
+    dpnp.check_supported_arrays_type(arr)
+    arr = dpnp.get_usm_ndarray(arr)
+    exec_q = arr.sycl_queue
+
+    dpnp.check_supported_arrays_type(val, scalar_type=True, all_scalars=True)
+    # if val is an array, process it
+    if isinstance(val, (dpnp_array, dpt.usm_ndarray)):
+        val = dpnp.get_usm_ndarray(val)
+        if val.shape != ():
+            raise ValueError("`val` must be a scalar")
```
antonwolfy, line 50: Could you please add tests to cover lines: 50, 71, 73?

antonwolfy (commented):
@ndgrigorian, thank you for implementing so great improvement.
Please find some comments below.

dpnp/dpnp_algo/dpnp_fill.py
```diff
@@ -0,0 +1,85 @@
...
+        # asarray moves scalar to the correct device
+        # and casts to the expected dtype
+        a_val = dpt.asarray(
```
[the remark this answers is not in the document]
ndgrigorian: Good point, my logic was that this case is allowed for both dpnp and dpctl in `fill`, but in retrospect, that function allows it because of `device` keyword.

I'll make a change to to disallow this case.

ndgrigorian:
@antonwolfy 
I've made the suggested changes, added the new tests, and switched back to skipping the one CuPy test.

The CuPy test looks for filling the array with a NumPy `ndarray`, but the `accept_error` parameter only works where CuPy and NumPy both raise the error, and of course NumPy does not raise in that case. Since we are specifically disallowing NumPy arrays as the fill value, I think the test is best skipped.

I've also updated docstring per request.

dpnp/dpnp_algo/dpnp_fill.py
```diff
@@ -0,0 +1,89 @@
...
+        val = np.asarray(val, dtype=dt)[()]
+
+    _manager = dpu.SequentialOrderManager[exec_q]
+    dep_evs = _manager.submitted_events
+    # can leverage efficient memset when val is 0
+    if arr.flags["FORC"] and val == 0:
+        h_ev, zeros_ev = _zeros_usm_ndarray(arr, exec_q, depends=dep_evs)
+        _manager.add_event_pair(h_ev, zeros_ev)
+    else:
+        h_ev, fill_ev = _full_usm_ndarray(val, arr, exec_q, depends=dep_evs)
```
antonwolfy, line 88: below example produce an unexpected error inside the extension:
```python
a = dpnp.ones(10, dtype='i')
a.fill('0')

Cell In[9], line 1
----> 1 a.fill('0')

File ~/code/dpnp/dpnp/dpnp_array.py:912, in dpnp_array.fill(self, value)
    909 # lazy import avoids circular imports
    910 from .dpnp_algo.dpnp_fill import dpnp_fill
--> 912 dpnp_fill(self, value)

File ~/code/dpnp/dpnp/dpnp_algo/dpnp_fill.py:92, in dpnp_fill(arr, val)
     90     _manager.add_event_pair(h_ev, zeros_ev)
     91 else:
---> 92     h_ev, fill_ev = _full_usm_ndarray(val, arr, exec_q, depends=dep_evs)
     93     _manager.add_event_pair(h_ev, fill_ev)

RuntimeError: Unable to cast Python instance of type <class 'str'> to C++ type '?' (#define PYBIND11_DETAILED_ERROR_MESSAGES or compile in debug mode for details)
```

tests/test_fill.py
```diff
@@ -0,0 +1,61 @@
+import dpctl
+import numpy as np
```
antonwolfy, line 2: Could you please extend `TEST_SCOPE` in `.github/workflows/conda-package.yml` with `test_fill.py`

dpnp/dpnp_algo/dpnp_fill.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+    dt = arr.dtype
+    val_type = type(val)
+    if val_type in [float, complex] and dpnp.issubdtype(dt, dpnp.integer):
+        val = int(val.real)
+    elif val_type is complex and dpnp.issubdtype(dt, dpnp.floating):
+        val = val.real
+    elif val_type is int and dpnp.issubdtype(dt, dpnp.integer):
+        val = np.asarray(val, dtype=dt)[()]
```
antonwolfy, line 79: What is the use case when it is required? Both input array and scalar have integer type, but casting to numpy.ndarray is done.

dpnp/dpnp_algo/dpnp_fill.py
```diff
@@ -0,0 +1,89 @@
...
+        return
+    elif not dpnp.isscalar(val):
+        raise TypeError(
+            f"Expected `val` to be a 0D-array or Python scalar, got {type(val)}"
+        )
+
+    dt = arr.dtype
+    val_type = type(val)
+    if val_type in [float, complex] and dpnp.issubdtype(dt, dpnp.integer):
+        val = int(val.real)
```
antonwolfy, line 75: Could you please cover line#75, line#77 by a test.

dpnp/dpnp_algo/dpnp_fill.py
ndgrigorian: It handles out-of-bounds integers for a data type. This is how dpctl handles it in `full` as well.

```
In [1]: import dpctl.tensor as dpt, dpctl, numpy as np

In [2]: sc = dpt.iinfo(dpt.int32).max + 1

In [3]: # Numpy 2 raises for out-of-bounds integer for type

In [4]: dpt.full((), sc, dtype="i4")
OverflowError                             Traceback (most recent call last)
Cell In[4], line 1
----> 1 dpt.full((), sc, dtype="i4")

File ~/repos/dpctl/dpctl/tensor/_ctors.py:1124, in full(shape, fill_value, dtype, order, device, usm_type, sycl_queue)
   1116 dtype = _get_dtype(dtype, sycl_queue, ref_type=type(fill_value))
   1117 res = dpt.usm_ndarray(
   1118     shape,
   1119     dtype=dtype,
   (...)
   1122     buffer_ctor_kwargs={"queue": sycl_queue},
   1123 )
-> 1124 fill_value = _cast_fill_val(fill_value, dtype)
   1126 _manager = dpctl.utils.SequentialOrderManager[sycl_queue]
   1127 # populating new allocation, no dependent events

File ~/repos/dpctl/dpctl/tensor/_ctors.py:1035, in _cast_fill_val(fill_val, dt)
   1033     return fill_val.real
   1034 elif val_type is int and np.issubdtype(dt, np.integer):
-> 1035     return _to_scalar(fill_val, dt)
   1036 else:
   1037     return fill_val

File ~/repos/dpctl/dpctl/tensor/_ctors.py:772, in _to_scalar(obj, sc_ty)
    767 def _to_scalar(obj, sc_ty):
    768     """A way to convert object to NumPy scalar type.
    769     Raises OverflowError if obj can not be represented
    770     using the requested scalar type.
    771     """
--> 772     zd_arr = np.asarray(obj, dtype=sc_ty)
    773     return zd_arr[tuple()]

OverflowError: Python integer 2147483648 out of bounds for int32
```

antonwolfy: Do you think we can directly call `_cast_fill_val` from dpctl then? (just to avoid duplication)

ndgrigorian: We can, I only avoided here to avoid using a function dpctl considers private, so as long as that's acceptable practice for dpnp I can make the change.

dpnp/dpnp_algo/dpnp_fill.py
```diff
@@ -0,0 +1,89 @@
...
+    elif val_type is complex and dpnp.issubdtype(dt, dpnp.floating):
+        val = val.real
+    elif val_type is int and dpnp.issubdtype(dt, dpnp.integer):
+        val = np.asarray(val, dtype=dt)[()]
+
+    _manager = dpu.SequentialOrderManager[exec_q]
+    dep_evs = _manager.submitted_events
+    # can leverage efficient memset when val is 0
+    if arr.flags["FORC"] and val == 0:
+        h_ev, zeros_ev = _zeros_usm_ndarray(arr, exec_q, depends=dep_evs)
```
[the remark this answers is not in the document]
ndgrigorian: We could but the `zeros` array would have to be brought to host to check for equality with 0.

Even using `if dpt.equal(a_zeros, 0)` involves two extra memory transfers (0 to device and `a_zeros` to host).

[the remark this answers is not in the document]
antonwolfy: I see, then it sounds the comment can be ignored.

dpnp/dpnp_algo/dpnp_fill.py
antonwolfy: I think it would be fine to use private method here, because we did that in other places also.

dpnp/dpnp_algo/dpnp_fill.py
ndgrigorian: `dpnp.full` produces this error, too.

```
In [1]: import dpnp as dnp

In [2]: dnp.full(10, "0", dtype="i4")
Cell In[2], line 1
----> 1 dnp.full(10, "0", dtype="i4")

File ~/repos/dpnp-fork/dpnp/dpnp_iface_arraycreation.py:2120, in full(shape, fill_value, dtype, order, like, device, usm_type, sycl_queue)
   2041 """
   2042 Return a new array of given shape and type, filled with `fill_value`.
   2043
   (...)
   2115
   2116 """
   2118 dpnp.check_limitations(like=like)
-> 2120 return dpnp_container.full(
   2121     shape,
   2122     fill_value,
   2123     dtype=dtype,
   2124     order=order,
   2125     device=device,
   2126     usm_type=usm_type,
   2127     sycl_queue=sycl_queue,
   2128 )

File ~/repos/dpnp-fork/dpnp/dpnp_container.py:230, in full(shape, fill_value, dtype, order, device, usm_type, sycl_queue)
    227     fill_value = fill_value.get_array()
    229 """Creates `dpnp_array` having a specified shape, filled with fill_value."""
--> 230 array_obj = dpt.full(
    231     shape,
    232     fill_value,
    233     dtype=dtype,
    234     order=order,
    235     usm_type=usm_type,
    236     sycl_queue=sycl_queue_normalized,
    237 )
    238 return dpnp_array._create_from_usm_ndarray(array_obj)

File ~/repos/dpctl/dpctl/tensor/_ctors.py:1128, in full(shape, fill_value, dtype, order, device, usm_type, sycl_queue)
   1126 _manager = dpctl.utils.SequentialOrderManager[sycl_queue]
   1127 # populating new allocation, no dependent events
-> 1128 hev, full_ev = ti._full_usm_ndarray(fill_value, res, sycl_queue)
   1129 _manager.add_event_pair(hev, full_ev)
   1130 return res

RuntimeError: Unable to cast Python instance of type <class 'str'> to C++ type '?' (#define PYBIND11_DETAILED_ERROR_MESSAGES or compile in debug mode for details)
```
For this case to be covered, `dpnp` will need a utility for extracting Python scalar from Python string.

That can be handled here or in a separate PR.

antonwolfy: As for now, we state only support for scalar or an array as an input `fill_value` to `dpnp.full` function and doesn't support string input.
Thus another alternative might be to have a validation check with a proper error message in both `dpt.full` and `dpnp.ndarray.fill` functions before calling `_full_usm_ndarray` from the extension. Since I guess `RuntimeError` is not that we are mainly expecting to have in such use case.

ndgrigorian: I have opened https://github.com/IntelPython/dpctl/pull/1878 to resolve this in `full` and `full_like`.

I will apply a similar validation check to `fill`.

tests/test_fill.py
```diff
@@ -0,0 +1,78 @@
...
+
+def test_fill_non_scalar():
+    a = dnp.ones(5, dtype="i4")
+    val = dnp.ones(2, dtype="i4")
+
+    with pytest.raises(ValueError):
+        a.fill(val)
+
+    with pytest.raises(TypeError):
+        a.fill(dict())
```
antonwolfy, line 18: It'd be helpful to cover a string value also
```suggestion
@pytest.mark.parametrize(
    "val, error",
    [
        pytest.param(dnp.ones(2, dtype="i4"), ValueError, id="array"),
        pytest.param(dict(), TypeError, id="dictionary"),
        pytest.param("0", TypeError, id="string"),
    ]
)
def test_fill_non_scalar(val, error):
    a = dnp.ones(5, dtype="i4")

    with pytest.raises(error):
        a.fill(val)
```

tests/test_fill.py
```diff
@@ -0,0 +1,78 @@
+import dpctl
+import numpy as np
```
antonwolfy, line 2: Unused import
```suggestion
```

tests/test_fill.py
```diff
@@ -0,0 +1,78 @@
...
+    a = dnp.ones(5, dtype="i4")
+    val = dnp.ones(2, dtype="i4")
+
+    with pytest.raises(ValueError):
+        a.fill(val)
+
+    with pytest.raises(TypeError):
+        a.fill(dict())
+
+
```
antonwolfy, line 20: and also it'd meaningful to cover bool fill value:
```suggestion

def test_fill_bool():
    a = dnp.full(5, fill_value=7, dtype="i4")

    a.fill(True)
    assert_array_equal(a, 1)

```

antonwolfy (approved):
Thank you @ndgrigorian for significant improvement of the `fill` method. No more comments from me.

ndgrigorian:
@antonwolfy 
I don't have the write access so please merge when you have the chance

PR #2063, opened 2024-09-18
Resolve crash identified in backend implementation of `dpnp.choose`

antonwolfy:
The PR resolves crash in `dpnp.choose` observing in internal CI/CD flow on PVC machine with Linux OS.

The issue was caused by missing dependency of temporary allocated device memory on a computing event. Thus the memory was released before the execution kernel completes work,
While allocation of temporary memory was implicit, because `DPNPC_ptr_adapter` constructor identified the input pointer `choices1` as of unknown USM allocation type. And it was valid classification, since the pointer was allocated in host memory for holding a list of pointers on arrays in device memory.

The change assumes to request explicit memory transfer for `choices1` pointer from host to device memory and to add a dependency from the temporary allocated device memory on the event from the submitted kernel. 


vtavana (approved):
Thank you, @antonwolfy!

PR #2096, opened 2024-10-08
Update a workaround to gemm issue in OneMKL

antonwolfy:
The PR proposes to update the workaround implemented in #2082, based on new input from OneMKL team.
It assumes to enable the w/a only in case when an input array does not have 16 bytes alignment in the memory and to remove any check on size of input array.
Also the w/a is extended to `gemm_batch` functionl, since it might be impacted.

Additionally an explicit test is added to verify the w/a scenario.


antonwolfy:
The issue has been resolved by OneMKL team, no need for a workaround anymore.

PR #2153, opened 2024-11-06
Add dpnp.broadcast_shapes implementation

ekomarova:
This PR adds the implementation of `dpnp.broadcast_shapes`

### DIFF
--- a/dpnp/dpnp_iface_manipulation.py
+++ b/dpnp/dpnp_iface_manipulation.py
@@ -63,6 +63,7 @@ __all__ = [
     "atleast_2d",
     "atleast_3d",
     "broadcast_arrays",
+    "broadcast_shapes",
     "broadcast_to",
     "can_cast",
     "column_stack",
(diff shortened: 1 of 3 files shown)


tests/test_manipulation.py
```diff
@@ -332,6 +332,14 @@ def test_no_copy(self):
         assert_array_equal(b, a)
 
 
+class TestBroadcast:
+    @pytest.mark.parametrize("shape", [(1, 1), (0, 1)])
```
antonwolfy, line 336: I guess it needs to be updated like:
```suggestion
    @pytest.mark.parametrize(
        "shape",
        [
            [(1,), (3,)],
            [(1, 3), (3, 3)],
            [(3, 1), (3, 3)],
            [(1, 3), (3, 1)],
            [(1, 1), (3, 3)],
            [(1, 1), (1, 3)],
            [(1, 1), (3, 1)],
            [(1, 0), (0, 0)],
            [(0, 1), (0, 0)],
            [(1, 0), (0, 1)],
            [(1, 1), (0, 0)],
            [(1, 1), (1, 0)],
            [(1, 1), (0, 1)],
        ],
    )
```

dpnp/dpnp_iface_manipulation.py
```diff
@@ -967,6 +969,44 @@ def broadcast_arrays(*args, subok=False):
...
+    (3, 2)
+    >>> np.broadcast_shapes((6, 7), (5, 6, 1), (7,), (5, 1, 7))
+    (5, 6, 7)
+
+    """
+
+    if hasattr(numpy, "broadcast_shapes"):
+        return numpy.broadcast_shapes(*args)
+
+    return _broadcast_shapes(*args)
```
antonwolfy, line 1007: `_broadcast_shapes` requires a list of array as an input, which is not the case here.
We should either replace it with `_broadcast_shape_impl(args)` from dpctl or use `numpy.broadcast_shapes` only, which is also an option, because it is available since 1.20 version (which is EOL already and so any actual older numpy version has it).

antonwolfy (approved):
Many thanks @ekomarova for your effort to reduce functional gap in dpnp.

PR #2177, opened 2024-11-18
Implement `dpnp.compress` and `dpnp_array.compress` method

ndgrigorian:
This PR proposes an implementation for `dpnp.compress` using `dpctl.tensor` kernels.

`nonzero` is used to accumulate the `condition` into 1d array of indices, which are then used to take the appropriate slices of the array. When `axis` is `None`, the operand array is flattened.

The coupled `dpnp_array.compress` method is implemented through a simple call to the `compress` function,

- [X] If this PR is a work in progress, are you filing the PR as a draft?

### DIFF
--- a/dpnp/dpnp_array.py
+++ b/dpnp/dpnp_array.py
@@ -786,7 +786,14 @@ class dpnp_array:
 
         return dpnp.clip(self, min, max, out=out, **kwargs)
 
-    # 'compress',
+    def compress(self, condition, axis=None, out=None):
+        """
+        Select slices of an array along a given axis.
+
+        Refer to :obj:`dpnp.compress` for full documentation.
+        """
+
+        return dpnp.compress(condition, self, axis=axis, out=out)
 
     def conj(self):
         """
--- a/dpnp/dpnp_iface_indexing.py
+++ b/dpnp/dpnp_iface_indexing.py
@@ -46,6 +46,7 @@ import dpctl.tensor._tensor_impl as ti
 import dpctl.utils as dpu
 import numpy
 from dpctl.tensor._copy_utils import _nonzero_impl
+from dpctl.tensor._indexing_functions import _get_indexing_mode
 from dpctl.tensor._numpy_helper import normalize_axis_index
 
 import dpnp
@@ -161,14 +162,13 @@ def choose(x1, choices, out=None, mode="raise"):
     return call_origin(numpy.choose, x1, choices, out, mode)
 
 
-def _take_1d_index(x, inds, axis, q, usm_type, out=None):
+def _take_index(x, inds, axis, q, usm_type, out=None, mode=0):
     # arg validation assumed done by caller
     x_sh = x.shape
-    ind0 = inds[0]
     axis_end = axis + 1
-    if 0 in x_sh[axis:axis_end] and ind0.size != 0:
+    if 0 in x_sh[axis:axis_end] and inds.size != 0:
         raise IndexError("cannot take non-empty indices from an empty axis")
-    res_sh = x_sh[:axis] + ind0.shape + x_sh[axis_end:]
+    res_sh = x_sh[:axis] + inds.shape + x_sh[axis_end:]
 
     orig_out = None
     if out is not None:
@@ -202,13 +202,12 @@ def _take_1d_index(x, inds, axis, q, usm_type, out=None):
     _manager = dpu.SequentialOrderManager[q]
     dep_evs = _manager.submitted_events
 
-    # always use wrap mode here
     h_ev, take_ev = ti._take(
         src=x,
-        ind=inds,
+        ind=(inds,),
         dst=out,
         axis_start=axis,
-        mode=0,
+        mode=mode,
         sycl_queue=q,
         depends=dep_evs,
     )
@@ -319,7 +318,8 @@ def compress(condition, a, axis=None, out=None):
     inds = _nonzero_impl(cond_ary)
 
     return dpnp.get_result_array(
-        _take_1d_index(a_ary, inds, axis, exec_q, res_usm_type, out), out=out
+        _take_index(a_ary, inds[0], axis, exec_q, res_usm_type, out=out),
+        out=out,
     )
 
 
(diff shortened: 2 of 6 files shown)


dpnp/dpnp_iface_indexing.py
```diff
@@ -154,6 +160,146 @@ def choose(x1, choices, out=None, mode="raise"):
     return call_origin(numpy.choose, x1, choices, out, mode)
 
 
+def _take_1d_index(x, inds, axis, q, usm_type, out=None):
```
antonwolfy, line 163: Should we reuse that function in below ``dpnp.take` implementation?

dpnp/dpnp_iface_indexing.py
```diff
@@ -154,6 +160,146 @@ def choose(x1, choices, out=None, mode="raise"):
...
+            dtype=dpnp.bool,
+            usm_type=usm_type,
+            sycl_queue=q,
+        )
+        queues_ = [q]
+        usm_types_ = [usm_type]
+    else:
+        cond_ary = dpnp.get_usm_ndarray(condition)
+        queues_ = [a_ary.sycl_queue, cond_ary.sycl_queue]
+        usm_types_ = [a_ary.usm_type, cond_ary.usm_type]
```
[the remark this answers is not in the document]
ndgrigorian: It can be, but it should still be somewhat broken up, because if it is an array, `dtype=dpnp.bool` is an unnecessary copy, as `nonzero` can accept any dtype input.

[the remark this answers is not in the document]
ndgrigorian: Another issue is that if `condition` is an array, it will go against compute follows data principle here: implicit data copy from the device of `condition` to the device of `a`.

I think it should stay split up between array and non-array cases for this reason.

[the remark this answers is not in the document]
antonwolfy: Yes, and that is done inside `dpnp.as_usm_ndarray` function.
It will fallback on `dpnp.get_usm_ndarray` for supported array type or will call `dpt.asarray(...)` otherwise.

dpnp/dpnp_iface_indexing.py
```diff
@@ -227,7 +226,8 @@ def _take_1d_index(x, inds, axis, q, usm_type, out=None):
 
 def compress(condition, a, axis=None, out=None):
     """
-    Return selected slices of an array along given axis.
+    A copy of `a` without the slices along `axis` for which `condition` is
```
antonwolfy, line 229: Should it be moved to `Returns`? It seems in numpy that exact description is used to describe the returned result.

dpnp/dpnp_iface_indexing.py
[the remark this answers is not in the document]
ndgrigorian: I see, then there should be no problem. I'll apply the suggestion.

dpnp/dpnp_iface_indexing.py
ndgrigorian: CuPy is also pretty particular with the language in this way, I'll apply this change

dpnp/tests/test_indexing.py
```diff
@@ -1307,3 +1309,51 @@ def test_error(self):
             dpnp.select([x0], [x1], default=x1)
         with pytest.raises(TypeError):
             dpnp.select([x1], [x1])
+
+
+def test_compress_basic():
+    a = dpnp.arange(16).reshape(4, 4)
+    condition = dpnp.asarray([True, False, True])
+    r = dpnp.compress(condition, a, axis=0)
+    assert_array_equal(r[0], a[0])
```
antonwolfy, line 1318: Usually we compare result against numpy output. It helps to simplify the test logic, to secure alignment in result dtypes, shapes and etc.

dpnp/tests/test_indexing.py
```diff
@@ -1307,3 +1309,51 @@ def test_error(self):
...
+    assert_array_equal(r[1], a[2])
+
+
+@pytest.mark.parametrize("dtype", get_all_dtypes())
```
antonwolfy, line 1322:
```suggestion
@pytest.mark.parametrize("dtype", get_all_dtypes(no_none=True))
```

dpnp/dpnp_iface_indexing.py
```diff
@@ -155,6 +162,157 @@ def choose(x1, choices, out=None, mode="raise"):
...
+        Default: ``None``.
+    out : {None, dpnp.ndarray, usm_ndarray}, optional
+        If provided, the result will be placed in this array. It should
+        be of the appropriate shape and dtype.
+        Default: ``None``.
+
+    Returns
+    -------
+    out : dpnp.ndarray
+        A copy of the slices of `a` where `condition` is True.
```
antonwolfy, line 247:
```suggestion
        A copy of the slices of `a` where `condition` is ``True``.
```

dpnp/tests/test_usm_type.py
```diff
@@ -1746,3 +1746,17 @@ def test_ix(usm_type_0, usm_type_1):
     ixgrid = dp.ix_(x0, x1)
     assert ixgrid[0].usm_type == x0.usm_type
     assert ixgrid[1].usm_type == x1.usm_type
+
+
+@pytest.mark.parametrize("usm_type_a", list_of_usm_types, ids=list_of_usm_types)
+@pytest.mark.parametrize(
+    "usm_type_cond", list_of_usm_types, ids=list_of_usm_types
+)
+def test_compress(usm_type_a, usm_type_cond):
```
antonwolfy, line 1755: Will it be possible to move these tests as part of parametrization for `test_2in_1out` per both `test_usm_type.py` and `test_sycl_queue.py` rather than a separate test?

dpnp/dpnp_array.py
```diff
@@ -786,7 +786,14 @@ def clip(self, min=None, max=None, out=None, **kwargs):
 
         return dpnp.clip(self, min, max, out=out, **kwargs)
 
-    # 'compress',
+    def compress(self, condition, axis=None, out=None):
```
antonwolfy, line 789: please add a test to cover the array method

dpnp/tests/test_usm_type.py
antonwolfy, line 1755: Will it be possible to do  the same as in `dpnp/tests/test_sycl_queue.py` and to move the test as part of `test_2in_1out`?

antonwolfy (approved):
Thank you @ndgrigorian. I have only a minor comment, but in overall LGTM!

antonwolfy (approved):
Thank you @ndgrigorian

ndgrigorian:
@antonwolfy CI is passed, please merge at your convenience

PR #2179, opened 2024-11-18
Do not use bash as default shell on Windows in Upload job

antonwolfy:
`Upload` job has been failing since version of `conda-incubator/setup-miniconda` was bumped up to `3.1.0`.
It seems `mamba install` doesn't properly work on Windows if the default job's shell is specified as `bash -el {0}`.

Since it doesn't look proper to use `bash -el {0}` as the Windows default shell, this PR proposes to remove that and to rely on default shell setting for a job.

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -337,10 +337,6 @@ jobs:
 
     runs-on: ${{ matrix.os }}
 
-    defaults:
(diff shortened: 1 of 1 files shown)


vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy for the timely implementation of this solution

PR #2190, opened 2024-11-21
Add a workflow to run nightly tests

antonwolfy:
The PR proposes to add a workflow which runs nightly and includes installing dpnp from `dppy/label/dev` and running tests. It will cover both Linux and Windows.
This PR ensures tests are run even if no development happens. The intent is it might catch issues due to external dependency updates, and such.

The workflow can also be triggered manually.


vtavana (approved):
Thank you, @antonwolfy!

PR #2202, opened 2024-11-27
Fix FFT negative strides case

AlexanderKalistratov:
FFT implementation validate arrays for supported strides/dtypes and if needed make a copy and cast (`_copy_array`).
Old implementation didn't validate for dtype in case of negative strides resulting in incorrect type selection in `_commit_descriptor` and failures on systems without `float64` support.

### DIFF
--- a/dpnp/tests/test_fft.py
+++ b/dpnp/tests/test_fft.py
@@ -378,6 +378,16 @@ class TestFft:
         out = dpnp.empty((10,), dtype=dpnp.float32)
         assert_raises(TypeError, dpnp.fft.fft, a, out=out)
 
+    @pytest.mark.parametrize(
+        "dtype", get_all_dtypes(no_none=True, no_bool=True)
+    )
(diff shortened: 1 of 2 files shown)


dpnp/tests/test_fft.py
```diff
@@ -378,6 +378,16 @@ def test_fft_validate_out(self):
         out = dpnp.empty((10,), dtype=dpnp.float32)
         assert_raises(TypeError, dpnp.fft.fft, a, out=out)
 
+    @pytest.mark.parametrize(
+        "dtype", get_all_dtypes(no_none=True, no_bool=True)
+    )
+    def test_negative_stride(self, dtype):
+        a = dpnp.arange(10).astype(dtype)
```
antonwolfy, line 385:
```suggestion
        a = dpnp.arange(10, dtype=dtype)
```

antonwolfy (approved):
Thank you @AlexanderKalistratov. I have only a minor comment.

PR #2203, opened 2024-11-27
Correlation via fft implementation

AlexanderKalistratov:
* Added keyword `method` to `correlate` function similar to [scipy correlate](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.correlate.html)
* If `method == 'auto'` method is choosing automatically between `direct` and `fft`
* Added implementation of fft-based correlation
* fft-based implementation may have accuracy issues, so it is validated in non-standard way.

Depends on: https://github.com/IntelPython/dpnp/pull/2180, https://github.com/IntelPython/dpnp/pull/2202


dpnp/dpnp_iface_statistics.py
```diff
@@ -506,7 +550,30 @@ def _run_native_sliding_dot_product1d(a, v, l_pad, r_pad):
     return out
 
 
-def correlate(a, v, mode="valid"):
+def _convolve_fft(a, v, l_pad, r_pad, rtype):
+    assert a.size >= v.size
+    assert l_pad < v.size
```
oleksandr-pavlyk, line 555: These assertions should be turned to checks that raise exceptions if unmet. 

This is because Python interpreter allows to turn off assertion checks with `-O` option:

```
(dev_dpctl) opavlyk@opavlyk-mobl:~/repos/dpctl$ python -O -c "assert False"
(dev_dpctl) opavlyk@opavlyk-mobl:~/repos/dpctl$ python  -c "assert False"
  File "<string>", line 1, in <module>
AssertionError
```

dpnp/dpnp_iface_statistics.py
```diff
@@ -506,7 +550,30 @@ def _run_native_sliding_dot_product1d(a, v, l_pad, r_pad):
...
+
+    # +1 is needed to avoid circular convolution
+    padded_size = a.size + r_pad + 1
+    fft_size = 2 ** math.ceil(math.log2(padded_size))
```
oleksandr-pavlyk, line 559: Since `math.ceil` returns `float`, it should be converted to integer:

```suggestion
    fft_size = 2 ** int(math.ceil(math.log2(padded_size)))
```

AlexanderKalistratov:
@antonwolfy @vlad-perevezentsev please review

dpnp/dpnp_iface_statistics.py
```diff
@@ -451,16 +455,55 @@ def _get_padding(a_size, v_size, mode):
...
+        return "direct"
+
+    if dpnp.issubdtype(rdtype, dpnp.integer):
+        max_a = int(dpnp.max(dpnp.abs(a)))
+        sum_v = int(dpnp.sum(dpnp.abs(v)))
+        max_value = int(max_a * sum_v)
+
+        default_float = dpnp.default_float_type(a.sycl_device)
+        if max_value > 2 ** numpy.finfo(default_float).nmant - 1:
+            return "direct"
```
antonwolfy, line 473: It seems there is not any test covering that line in public CI. Is there any way to add it?

dpnp/dpnp_iface_statistics.py
```diff
@@ -451,16 +455,55 @@ def _get_padding(a_size, v_size, mode):
...
+def _run_native_sliding_dot_product1d(a, v, l_pad, r_pad, rdtype):
     queue = a.sycl_queue
+    device = a.sycl_device
 
-    usm_type = dpu.get_coerced_usm_type([a.usm_type, v.usm_type])
-    out_size = l_pad + r_pad + a.size - v.size + 1
+    supported_types = statistics_ext.sliding_dot_product1d_dtypes()
+    supported_dtype = to_supported_dtypes(rdtype, supported_types, device)
+
+    if supported_dtype is None:
```
antonwolfy, line 488: dpctl implements an approach to omit certain branches, certain lines from coverage stats in [dpctl#1970](https://github.com/IntelPython/dpctl/pull/1970).
Can we do something similar?
```suggestion
    if supported_dtype is None:  # pragma: no cover
```

dpnp/dpnp_iface_statistics.py
```diff
@@ -478,7 +521,30 @@ def _run_native_sliding_dot_product1d(a, v, l_pad, r_pad):
...
+    vf = dpnp.fft.fft(v, fft_size)  # pylint: disable=no-member
+
+    r = dpnp.fft.ifft(af * vf)  # pylint: disable=no-member
+    if dpnp.issubdtype(rtype, dpnp.floating):
+        r = r.real
+    elif dpnp.issubdtype(rtype, dpnp.integer) or rtype == dpnp.bool:
+        r = r.real.round()
+
+    start = v.size - 1 - l_pad
+    end = padded_size - 1
```
antonwolfy, line 542: Should `start` and `end` depend on `mode` keyword of `correlate`?
It seem cupy has such logic.

dpnp/dpnp_iface_statistics.py
```diff
@@ -595,10 +669,15 @@ def correlate(a, v, mode="valid"):
...
-    v_casted = dpnp.asarray(v, dtype=supported_dtype, order="C")
+    if method == "auto":
+        method = _choose_conv_method(a, v, rdtype)
 
-    r = _run_native_sliding_dot_product1d(a_casted, v_casted, l_pad, r_pad)
+    if method == "direct":
+        r = _run_native_sliding_dot_product1d(a, v, l_pad, r_pad, rdtype)
+    elif method == "fft":
+        r = _convolve_fft(a, v[::-1], l_pad, r_pad, rdtype)
+    else:
```
antonwolfy, line 679: Should we mark it like
```suggestion
    else:  # pragma: no cover
```

dpnp/dpnp_iface_statistics.py
```diff
@@ -503,10 +569,24 @@ def correlate(a, v, mode="valid"):
...
+        Note: Use of the FFT convolution on input containing NAN or INF
+        will lead to the entire output being NAN or INF.
+        Use method='direct' when your input contains NAN or INF values.
+
+        Default: ``"auto"``.
 
     Returns
     -------
-    out : dpnp.ndarray
+    out : {dpnp.ndarray}
```
antonwolfy, line 589: Why do we need brackets here?

dpnp/tests/test_statistics.py
```diff
@@ -583,26 +583,104 @@ def test_corrcoef_scalar(self):
...
-            expected = numpy.correlate(an, vn, mode=mode)
-            result = dpnp.correlate(ad, vd, mode=mode)
+        dpnp_kwargs = {}
+        numpy_kwargs = {}
+        if mode is not None:
+            dpnp_kwargs["mode"] = mode
+            numpy_kwargs["mode"] = mode
+        if method is not None:
+            dpnp_kwargs["method"] = method
+
```
antonwolfy, line 608: We can make that as a method:
```python
    @staticmethod
    def _get_kwargs(mode=None, method=None):
        dpnp_kwargs = {}
        numpy_kwargs = {}
        if mode is not None:
            dpnp_kwargs["mode"] = mode
            numpy_kwargs["mode"] = mode
        if method is not None:
            dpnp_kwargs["method"] = method
        return dpnp_kwargs, numpy_kwargs
```
and then it would help to reduce code duplication:
```suggestion
        dpnp_kwargs, numpy_kwargs = self._get_kwargs(mode=mode, method=method)
```

dpnp/dpnp_iface_statistics.py
```diff
@@ -451,16 +455,55 @@ def _get_padding(a_size, v_size, mode):
     return l_pad, r_pad
 
 
-def _run_native_sliding_dot_product1d(a, v, l_pad, r_pad):
+def _choose_conv_method(a, v, rdtype):
```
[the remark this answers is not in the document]
AlexanderKalistratov: Added

dpnp/dpnp_iface_statistics.py
```diff
@@ -451,16 +455,55 @@ def _get_padding(a_size, v_size, mode):
...
+    if dpnp.issubdtype(rdtype, dpnp.integer):
+        max_a = int(dpnp.max(dpnp.abs(a)))
+        sum_v = int(dpnp.sum(dpnp.abs(v)))
+        max_value = int(max_a * sum_v)
+
+        default_float = dpnp.default_float_type(a.sycl_device)
+        if max_value > 2 ** numpy.finfo(default_float).nmant - 1:
+            return "direct"
+
+    if dpnp.issubdtype(rdtype, dpnp.number):
```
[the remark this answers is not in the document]
AlexanderKalistratov: It is here just to be sure fft would work and we would get meaningful error otherwise.
Don't think this check makes any harm

dpnp/dpnp_iface_statistics.py
AlexanderKalistratov: No, we are handling it by `l_pad` and `r_pad`

dpnp/dpnp_iface_statistics.py
```diff
@@ -55,6 +57,8 @@
 from .dpnp_utils.dpnp_utils_reduction import dpnp_wrap_reduction_call
 from .dpnp_utils.dpnp_utils_statistics import dpnp_cov, dpnp_median
 
+min_ = min  # pylint: disable=used-before-assignment
```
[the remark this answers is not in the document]
AlexanderKalistratov: removed

dpnp/dpnp_iface_statistics.py
AlexanderKalistratov: It is hard to add test for it.
Probably it is better to have separate tests for `_choose_conv_method` as scipy do. But it has [choose_conv_method](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.choose_conv_method.html#scipy.signal.choose_conv_method) as part of public API which actually used only inside `correlate`/`convolve` functions

dpnp/dpnp_iface_statistics.py
AlexanderKalistratov: Added

dpnp/dpnp_iface_statistics.py
AlexanderKalistratov: Added

dpnp/dpnp_iface_statistics.py
AlexanderKalistratov: removed

dpnp/tests/test_statistics.py
AlexanderKalistratov: Done

dpnp/dpnp_iface_statistics.py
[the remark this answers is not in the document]
antonwolfy: Would you mind to add some comments to the function? Seems still missing.

dpnp/dpnp_iface_statistics.py
```diff
@@ -503,6 +570,20 @@ def correlate(a, v, mode="valid"):
...
 
         Default: ``"valid"``.
+    method : {"auto", "direct", "fft"}, optional
+        `"direct"`: The correlation is determined directly from sums.
+
+        `"fft"`: The Fourier Transform is used to perform the calculations.
+        This method is faster for long sequences but can have accuracy issues.
+
+        `"auto"`: Automatically chooses direct or Fourier method based on
+        an estimate of which is faster.
```
antonwolfy, line 580: It would be better to format the description like:
```suggestion
        Specifies which method to use to calculate the correlation:

        - `"direct"` : The correlation is determined directly from sums.
        - `"fft"` : The Fourier Transform is used to perform the calculations.
          This method is faster for long sequences but can have accuracy issues.
        - `"auto"` : Automatically chooses direct or Fourier method based on
          an estimate of which is faster.
```
You can look into the docstring of `put_along_axis` as a reference.

dpnp/tests/test_statistics.py
```diff
@@ -583,26 +583,103 @@ def test_corrcoef_scalar(self):
...
+        if dtype == dpnp.bool:
+            an = numpy.random.rand(a_size) > 0.9
+            vn = numpy.random.rand(v_size) > 0.9
+        else:
+            an = 100 * numpy.random.rand(a_size).astype(dtype)
+            vn = 100 * numpy.random.rand(v_size).astype(dtype)
+
+            if dpnp.issubdtype(dtype, dpnp.complexfloating):
+                an = an + 100j * numpy.random.rand(a_size).astype(dtype)
+                vn = vn + 100j * numpy.random.rand(v_size).astype(dtype)
```
antonwolfy, line 634: It might be helpful to use `generate_random_numpy_array` helper utility if it fits the needs.

dpnp/dpnp_iface_statistics.py
antonwolfy: I believe the current set of tests is more than enough then.

PR #2206, opened 2024-11-29
Implement `dpnp.ndarray.__iter__` method

antonwolfy:
The PR proposes to add implementation of `dpnp.ndarray.__iter__` method.
The new code is assumed to be covered by third party tests.

- [x] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?

PR #2208, opened 2024-12-03
Add `dpnp.iterable` function

antonwolfy:
There was `__iter__` method of dpnp.ndarray implemented in scope of #2206.

While this PR proposes to add implementation of `dpnp.iterable` function which is intended to check if the object (including dpnp.ndarray) can be iterated over.
The implementation leverages on NumPy call, which is based on attempt to pass the object to `iter` function.


vtavana (approved):
Thank you, @antonwolfy!

PR #2221, opened 2024-12-06
Remove test_cond_nan_input from TestCond

vlad-perevezentsev:
This PR suggests removing `test_cond_input_nan` due to undefined behavior for different backend in different versions of OpenBLAS and OneMKL to avoid possible future falls

_Checking all input for nan's is expensive, so MKL lapack functions don't do it_


antonwolfy (approved):
Thank you @vlad-perevezentsev

PR #2225, opened 2024-12-09
Fix warnings in blas extensions during build on CUDA

vlad-perevezentsev:
This PR suggests using conditional logic in blas extensions **(gemm, gemm_batch, gemv**) to fix the warning about `unused 'is_row_major' parameter` during dpnp build on CUDA. The changes ensure that `is_row_major` parameter is only passed when the backend requires it.


antonwolfy (approved):
Thank you @vlad-perevezentsev, LGTM

PR #2236, opened 2024-12-16
mamba search does not accept --info option

antonwolfy:
The PR proposes to resolve the issue observing in GitHub actions where `mamba search --info` results into `The following argument was not expected: --info` error. The error started to be reported since `mamba==2.0.5` release.

As a workaround it is proposed to switch to `conda search --info` command which works properly.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy !

PR #2249, opened 2025-01-07
Support python 3.13

antonwolfy:
The PR proposes to build and test with Python 3.13 in conda package workflow.

Also default python version used by other GH actions is bumped up to 3.13 except:
- building docs because there is no support of python 3.13 in CuPy
- coverage because there is no support of python 3.13 in `coveralls`

Note, `conda-build` does not support python 3.13, but it is still possible to install it into env with python 3.12 and to run conda build with `--python 3.13` option.

Also it forces to pin `numpy >= 1.23` in `meta.yaml` otherwise it is impossible to use the latest numpy version, because it is pinned to 2.0 by `--numpy 2.0` option (while numpy 2.0 does not support python 3.13 yet, it has been added later).

### DIFF
--- a/.github/workflows/build-sphinx.yml
+++ b/.github/workflows/build-sphinx.yml
@@ -36,7 +36,7 @@ jobs:
       pull-requests: write
 
     env:
-      python-ver: '3.13'
+      python-ver: '3.12'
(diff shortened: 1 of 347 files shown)


vlad-perevezentsev (approved):
Looks good
Thank you @antonwolfy !

PR #2251, opened 2025-01-08
Update `TestSpacing::test_zeros` 

vlad-perevezentsev:
This PR suggests using `assert_dtype_allclose` instead of `assert_equal` to fix failures on CUDA tests caused by excessive precision for float32.

```
Mismatched elements: 2 / 2 (100%)
E           Max absolute difference among violations: 1.1754942e-38
E           Max relative difference among violations: 8388607.
E            ACTUAL: array([ 1.175494e-38, -1.175494e-38], dtype=float32)
E            DESIRED: array([ 1.e-45, -1.e-45], dtype=float32)
```

```
$ ONEAPI_DEVICE_SELECTOR=cuda pytest dpnp/tests/test_mathematical.py::TestSpacing::test_zeros
==============================2 passed in 1.18s =================================
```

### DIFF
--- a/dpnp/tests/test_mathematical.py
+++ b/dpnp/tests/test_mathematical.py
@@ -2112,13 +2112,13 @@ class TestSpacing:
         result = dpnp.spacing(ia)
         expected = numpy.spacing(a)
         if numpy.lib.NumpyVersion(numpy.__version__) < "2.0.0":
-            assert_equal(result, expected)
+            assert_dtype_allclose(result, expected)
         else:
             # numpy.spacing(-0.0) == numpy.spacing(0.0), i.e. NumPy returns
             # positive value (looks as a bug in NumPy), because for any other
             # negative input the NumPy result will be also a negative value.
             expected[1] *= -1
-            assert_equal(result, expected)
+            assert_dtype_allclose(result, expected)
 
     @pytest.mark.parametrize("dt", get_float_dtypes(no_float16=False))
     @pytest.mark.parametrize("val", [1, 1e-5, 1000])


dpnp/tests/test_mathematical.py
```diff
@@ -2112,13 +2112,13 @@ def test_zeros(self, dt):
         result = dpnp.spacing(ia)
         expected = numpy.spacing(a)
         if numpy.lib.NumpyVersion(numpy.__version__) < "2.0.0":
-            assert_equal(result, expected)
+            assert_dtype_allclose(result, expected)
```
antonwolfy, line 2115: The result dtype will always follow input dtype for floating types, so no need to use `assert_dtype_allclose` but `assert_allclose` seems to be enough.

vlad-perevezentsev: I used this function because it calculates `atol` for the input dtype and passes it to `assert_allclose()`
Using `assert_allclose` with default `atol=0` causes the test to fail
```
AssertionError:
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 2 / 2 (100%)
Max absolute difference among violations: 1.1754942e-38
Max relative difference among violations: 0.9999999
 ACTUAL: array([ 1.e-45, -1.e-45], dtype=float32)
 DESIRED: array([ 1.175494e-38, -1.175494e-38], dtype=float32)
```

antonwolfy: You can do the same, like:
```python
tol = numpy.finfo(expected.dtype).resolution
assert_allclose(result, expected, rtol=tol, atol=tol)
```
or, alternatively, you can:
```python
assert_dtype_allclose(result, expected, factor=1)
```
which is exactly the same, but more precise.
In general `assert_dtype_allclose` default is intended for a case when input arrays might have different dtype due to limited fp support be a device.

vlad-perevezentsev: Done
Thank you!

antonwolfy (approved):

PR #2252, opened 2025-01-08
Add `__array_namespace__` method

antonwolfy:
The PR proposes to implement `__array_namespace__` method of dpnp ndarray, which is required to be compliant with python array API.
The method will return dpnp as an array namespace, member functions of which implement data API.

The array namespace is assumed to be stored inside dpctl tensor. So dpnp ndarray constructor is updated to explicitly pass `array_namespace=dpnp` into `dpt.usm_ndarray` call. And also to set the namespace through `_set_namespace(dpnp)` every time dpnp ndarray is created from usm_ndarray.

### DIFF
--- a/dpnp/dpnp_array.py
+++ b/dpnp/dpnp_array.py
@@ -353,15 +353,7 @@ class dpnp_array:
         key = _get_unwrapped_index_key(key)
 
         item = self._array_obj.__getitem__(key)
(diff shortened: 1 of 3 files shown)


vlad-perevezentsev (approved):
Looks good!
Thank you @antonwolfy

PR #2253, opened 2025-01-10
Add GitHub job with array API conformity run

antonwolfy:
The PR adds a GitHub job to `Conda package` workflow with array API conformity.

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -210,7 +210,7 @@ jobs:
 
       - name: Smoke test
         run: |
-          python -c "import dpnp, dpctl; dpctl.lsplatform()"
+          python -c "import dpctl; dpctl.lsplatform()"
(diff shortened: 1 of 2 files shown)


.github/workflows/conda-package.yml
```diff
@@ -432,6 +432,190 @@ jobs:
...
 
+  array-api-conformity:
+    name: Array API conformity
+
+    needs: build
+
+    permissions:
+      # Needed to add a comment to a pull request's issue
+      pull-requests: write
+
```
vlad-perevezentsev, line 443: Could you set a timeout in case there is a hang during test runs?

antonwolfy: It has to be a general improvement across all workflows, because currently we are not defining any timeout at all.

vlad-perevezentsev (approved):
Looks good!
Thank you @antonwolfy

PR #2258, opened 2025-01-13
Add environment file to GitHub workflows

antonwolfy:
The PR proposes to update GitHub workflows which used miniconda and to install all required conda packages through passing an environment file by single step. It will make the jibs more robust against connectivity issues which occur often during downloading conda packages.

There are dedicated yaml files with min set of required packages introduced to reuse them in different jobs/workflows with common requirements. In case when few yaml files needs to be merged into one `environment.yml` there is `conda-merge` tool used.


.github/workflows/build-sphinx.yml
```diff
@@ -36,11 +36,11 @@ jobs:
       pull-requests: write
 
     env:
-      python-ver: '3.12'
-      CHANNELS: '-c dppy/label/dev -c intel -c conda-forge --override-channels'
-      NO_INTEL_CHANNELS: '-c dppy/label/dev -c conda-forge --override-channels'
-      # Install the latest oneAPI compiler to work around an issue
-      INSTALL_ONE_API: 'yes'
+      environment-file: 'environments/environment.yml'
```
AlexanderKalistratov, line 39: Is `environment.yml` missing?

antonwolfy: It's output from the above command: `conda-merge ${{ env.base-doc-env }} ${{ env.build-pkgs-env }} > ${{ env.environment-file }}`

AlexanderKalistratov: Cool. Do we need to hardcode `python=3.12` dependency in [base_duild_docs.yml](https://github.com/IntelPython/dpnp/pull/2258/files#diff-0eacc7e983d109ddf0ae896b0da58407221f03bda76ffd0399349dd83497bed0R6) ?

Taking into consideration that we can dynamically create environment files using `conda-merge`?

Or am I missing something?

antonwolfy: The file will be split later, because part of packages are common with `coverage` workflow.
The PR is work in progress still :)

In general, the python is pinned as 3.12, because cupy package doesn't support 3.13, i.e. max python version depends on cupy package and so I linked them there in one file.

Alternatively we can miss python in `base_duild_docs.yml` and to pass as option of `conda-incubator/setup-miniconda` step, but I preferred to do that through `base_duild_docs.yml` file.

vlad-perevezentsev (approved):
Great changes!
Thank you @antonwolfy for optimizing dependency management for jobs

PR #2263, opened 2025-01-15
Add missing `stream` keyword argument to `dpnp.ndarray.to_device` method

antonwolfy:
The PR proposes to align `dpnp.ndarray.to_device` method with python array API and to add support of `stream` keyword argument.
The tests coverage is extended to cover new argument. Previously muted array-api test is enabled back.

Note, tests for `dpnp.__dlpack__` are updated due to recent changes done in [dpctl#1969](https://github.com/IntelPython/dpctl/pull/1969).

### DIFF
--- a/dpnp/tests/test_sycl_queue.py
+++ b/dpnp/tests/test_sycl_queue.py
@@ -1932,23 +1932,51 @@ def test_svd(shape, full_matrices, compute_uv, device):
         assert_sycl_queue_equal(dpnp_s_queue, expected_queue)
 
 
-@pytest.mark.parametrize(
-    "device_from",
-    valid_devices,
-    ids=[device.filter_string for device in valid_devices],
-)
-@pytest.mark.parametrize(
-    "device_to",
-    valid_devices,
-    ids=[device.filter_string for device in valid_devices],
-)
-def test_to_device(device_from, device_to):
-    data = [1.0, 1.0, 1.0, 1.0, 1.0]
+class TestToDevice:
+    @pytest.mark.parametrize(
+        "device_from",
+        valid_devices,
+        ids=[device.filter_string for device in valid_devices],
+    )
+    @pytest.mark.parametrize(
+        "device_to",
+        valid_devices,
+        ids=[device.filter_string for device in valid_devices],
+    )
+    def test_basic(self, device_from, device_to):
+        data = [1.0, 1.0, 1.0, 1.0, 1.0]
+        x = dpnp.array(data, dtype=dpnp.float32, device=device_from)
 
-    x = dpnp.array(data, dtype=dpnp.float32, device=device_from)
-    y = x.to_device(device_to)
+        y = x.to_device(device_to)
+        assert y.sycl_device == device_to
 
-    assert y.sycl_device == device_to
+    def test_to_queue(self):
+        x = dpnp.full(100, 2, dtype=dpnp.int64)
+        q_prof = dpctl.SyclQueue(x.sycl_device, property="enable_profiling")
+
+        y = x.to_device(q_prof)
+        assert_sycl_queue_equal(y.sycl_queue, q_prof)
+
+    def test_stream(self):
+        x = dpnp.full(100, 2, dtype=dpnp.int64)
+        q_prof = dpctl.SyclQueue(x.sycl_device, property="enable_profiling")
+        q_exec = dpctl.SyclQueue(x.sycl_device)
+
+        y = x.to_device(q_prof, stream=q_exec)
+        assert_sycl_queue_equal(y.sycl_queue, q_prof)
+
+        q_exec = dpctl.SyclQueue(x.sycl_device)
+        _ = dpnp.linspace(0, 20, num=10**5, sycl_queue=q_exec)
+        y = x.to_device(q_prof, stream=q_exec)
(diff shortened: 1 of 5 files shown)


dpnp/tests/test_sycl_queue.py
```diff
@@ -1932,23 +1932,51 @@ def test_svd(shape, full_matrices, compute_uv, device):
...
+        "device_to",
+        valid_devices,
+        ids=[device.filter_string for device in valid_devices],
+    )
+    def test_basic(self, device_from, device_to):
+        data = [1.0, 1.0, 1.0, 1.0, 1.0]
+        x = dpnp.array(data, dtype=dpnp.float32, device=device_from)
+
+        y = x.to_device(device_to)
+        assert y.sycl_device == device_to
```
AlexanderKalistratov, line 1951: Shouldn't we also compare content of the arrays?
Like ```assert_array_equal(x, y)```

dpnp/tests/test_sycl_queue.py
```diff
@@ -1932,23 +1932,51 @@ def test_svd(shape, full_matrices, compute_uv, device):
...
+        assert_sycl_queue_equal(y.sycl_queue, q_prof)
 
-    x = dpnp.array(data, dtype=dpnp.float32, device=device_from)
-    y = x.to_device(device_to)
+    def test_stream_no_sync(self):
+        x = dpnp.full(100, 2, dtype=dpnp.int64)
+        q_prof = dpctl.SyclQueue(x.sycl_device, property="enable_profiling")
 
-    assert y.sycl_device == device_to
+        for stream in [None, 1, dpctl.SyclDevice(), x.sycl_queue]:
```
AlexanderKalistratov, line 1977: And shouldn't passing `1` as stream cause an exception?

dpnp/tests/test_sycl_queue.py
```diff
@@ -1932,23 +1932,51 @@ def test_svd(shape, full_matrices, compute_uv, device):
...
+        q_prof = dpctl.SyclQueue(x.sycl_device, property="enable_profiling")
+        q_exec = dpctl.SyclQueue(x.sycl_device)
+
+        y = x.to_device(q_prof, stream=q_exec)
+        assert_sycl_queue_equal(y.sycl_queue, q_prof)
+
+        q_exec = dpctl.SyclQueue(x.sycl_device)
+        _ = dpnp.linspace(0, 20, num=10**5, sycl_queue=q_exec)
+        y = x.to_device(q_prof, stream=q_exec)
+        assert_sycl_queue_equal(y.sycl_queue, q_prof)
```
AlexanderKalistratov, line 1971: Well, it seems we are not testing actual functionality of `stream` keyword?

AlexanderKalistratov (commented):
Everything is good, but we are not testing actual functionality of `stream` keyword which purpose is to do synchronization.

We need to figure out how to properly test it.

dpnp/tests/test_sycl_queue.py
antonwolfy: Based on python array API [specification](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.to_device.html) `stream` can have any type:
> array.to_device(device: device, /, *, stream: int | Any | None = None) → array

I will update the description of `stream` argument to clarify that synchronization will be only in case of `SyclQueue` type.

In general the behavior of `stream` argument is implementation-dependent, but must be clearly explained in the library's documentation.
The PR assumes that dpnp implementation fully relies on dpctl behavior here.

AlexanderKalistratov: `stream` can have any type because it is general API and `Array API` has no idea of your actual `stream` type. It is just signature for static type checkers.
It doesn't mean we must accept any object.
And also according to [array_api.array.__dlpack__](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html#array_api.array.__dlpack__ ) `int` is used to pass pointers.
If we are not accepting pointers or special values (like `-1`), we shouldn't accepts `int` either.

AlexanderKalistratov: from [array_api.array.dlpack](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html#array_api.array.__dlpack__)

```
Support for a stream value other than None is optional and implementation-dependent.
```

antonwolfy: I guess these questions need to addressed to dpctl implementation mainly, because dpnp fully leverages dpctl logic here.
I don't think we need to add extra checks into dpnp implementation. If there is type check required, it will be applicable to dpctl also.

antonwolfy: The fix was implemented in [dpctl#1969](https://github.com/IntelPython/dpctl/pull/1969).
I will align dpnp implementation.

dpnp/tests/test_sycl_queue.py
antonwolfy: There is no good way to test that.

dpnp/tests/test_sycl_queue.py
antonwolfy: I'll add the assertion to test that.

antonwolfy:
@AlexanderKalistratov , I've applied the comments, please rereview.

PR #2075, opened 2024-09-24
Handle dpnp functions and tests to run on CUDA devices

npolina4:
This PR suggests updating some function implementations and tests to run on CUDA devices
The PR includes:

1. Raise NotImplemetedError for unsupported functions
2. Create **skipped_tests_cuda.tbl** for unsupported functions
3. Skipped tests with a bug ticket for functions that should run on CUDA 
4. Implement **is_cuda_backend()** and **is_cuda_device()** to determine that the object has a CUDA backend and that the tests run on CUDA

- [x] Have you provided a meaningful PR description?
- [x] Have you added a test, reproducer or referred to issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?

PR #2265, opened 2025-01-17
Workaround for `dpnp.linalg.qr()` to run on CUDA

vlad-perevezentsev:
This PR suggests adding a workaround like waiting for host task after calling `geqrf` to avoid a race condition due to an issue in oneMath ([626](https://github.com/IntelPython/dpnp/pull/2075)) 

Also updates tests by removing old skips and adds `test_qr_large` in `TestQr` 

### DIFF
--- a/dpnp/linalg/dpnp_utils_linalg.py
+++ b/dpnp/linalg/dpnp_utils_linalg.py
@@ -399,6 +399,12 @@ def _batched_qr(a, mode="reduced"):
     )
     _manager.add_event_pair(ht_ev, geqrf_ev)
 
+    # w/a to avoid raice conditional on CUDA during multiple runs
+    # TODO: Remove it ones the OneMath issue is resolved
+    # https://github.com/uxlfoundation/oneMath/issues/626
+    if dpnp.is_cuda_backend(a_sycl_queue):
+        ht_ev.wait()
+
     if mode in ["r", "raw"]:
         if mode == "r":
             r = a_t[..., :k].swapaxes(-2, -1)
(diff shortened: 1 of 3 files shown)


dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -2470,6 +2476,12 @@ def dpnp_qr(a, mode="reduced"):
     )
     _manager.add_event_pair(ht_ev, geqrf_ev)
 
+    # w/a to avoid raice conditional on CUDA during multiple runs
+    # TODO: Remove it ones the OneMath issue is resolved
+    # https://github.com/uxlfoundation/oneMath/issues/626
+    if dpnp.is_cuda_backend(a_sycl_queue):
+        ht_ev.wait()
```
antonwolfy, line 2483: Does it still make sense to add the events to the manager list (above call `_manager.add_event_pair(ht_ev, geqrf_ev)`) when we are going make a sync point here in case of a w/a?

vlad-perevezentsev: It is a good point.
I think it makes no sense to add the events to the manager list here
I will update it

antonwolfy (approved):
LGTM! Thank you @vlad-perevezentsev

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -397,7 +397,14 @@ def _batched_qr(a, mode="reduced"):
         batch_size,
         depends=[copy_ev],
     )
-    _manager.add_event_pair(ht_ev, geqrf_ev)
+
+    # w/a to avoid raice conditional on CUDA during multiple runs
+    # TODO: Remove it ones the OneMath issue is resolved
+    # https://github.com/uxlfoundation/oneMath/issues/626
+    if dpnp.is_cuda_backend(a_sycl_queue):
```
antonwolfy, line 404: dpctl implements an approach to omit certain branches, certain lines from coverage stats in [dpctl#1970](https://github.com/IntelPython/dpctl/pull/1970).
Should we do something similar here?
```suggestion  # pragma: no cover
    if dpnp.is_cuda_backend(a_sycl_queue):
```

vlad-perevezentsev: I can add this.
But I would prefer to create a separate PR to omit lines with this check from coverage report

antonwolfy: I don't mind to handle that separately.

PR #2272, opened 2025-01-22
Exclude code from coverage

vlad-perevezentsev:
This PR continues #2268  and suggests excluding some lines of code from coverage using `pragma:: no cover` to increase the coverage statistic

- [X] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to issue with a reproducer?
- [ ] Have you tested your changes locally for CPU and GPU devices?
- [X] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?

PR #2268, opened 2025-01-20
Increase code coverage for `.py` files

vlad-perevezentsev:
This PR suggests increasing code coverage by adding missing tests for uncovered lines of code for `python` files

- [X] Have you provided a meaningful PR description?
- [X] Have you added a test, reproducer or referred to issue with a reproducer?
- [X] Have you tested your changes locally for CPU and GPU devices?
- [X] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] If this PR is a work in progress, are you filing the PR as a draft?

PR #2277, opened 2025-01-24
Add new tests to improve coverage stats

vlad-perevezentsev:
This PR continues #2268 and #2272 and suggests adding a couple new tests to slightly improve code coverage stats 

### DIFF
--- a/dpnp/tests/test_mathematical.py
+++ b/dpnp/tests/test_mathematical.py
@@ -3320,30 +3320,6 @@ class TestMatmul:
         expected = a[1] @ b[1]
         assert_dtype_allclose(result, expected)
 
-    def test_rmatmul_dpnp_array(self):
-        a = dpnp.ones(10)
-        b = dpnp.ones(10)
-
-        class A(dpnp.ndarray):
-            def __init__(self, x):
-                self._array_obj = x.get_array()
-
-            def __matmul__(self, other):
-                return NotImplemented
-
-        a1 = A(a)
-
-        result = a1 @ b
-        expected = dpnp.matmul(a, b)
-        assert_dtype_allclose(result, expected)
-
-    def test_rmatmul_numpy_array(self):
-        a = dpnp.ones(10)
-        b = numpy.ones(10)
-
-        with pytest.raises(TypeError):
-            b @ a
-
 
 class TestMatmulInplace:
     ALL_DTYPES = get_all_dtypes(no_none=True)
--- a/dpnp/tests/test_mathematical.py
+++ b/dpnp/tests/test_mathematical.py
@@ -3320,6 +3320,30 @@ class TestMatmul:
         expected = a[1] @ b[1]
         assert_dtype_allclose(result, expected)
 
+    def test_rmatmul_dpnp_array(self):
+        a = dpnp.ones(10)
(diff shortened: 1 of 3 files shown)


dpnp/tests/test_mathematical.py
```diff
@@ -3320,6 +3320,30 @@ def test_matmul_with_offsets(self, sh1, sh2):
...
+        class A(dpnp.ndarray):
+            def __init__(self, x):
+                self._array_obj = x.get_array()
+
+            def __matmul__(self, other):
+                return NotImplemented
+
+        a1 = A(a)
+
+        result = a1 @ b
```
antonwolfy, line 3336:
```suggestion
        class Dummy(dpnp.ndarray):
            def __init__(self, x):
                self._array_obj = x.get_array()

            def __matmul__(self, other):
                return NotImplemented

        d = Dummy(a)

        result = d @ b
```

dpnp/tests/test_mathematical.py
```diff
@@ -3320,6 +3320,30 @@ def test_matmul_with_offsets(self, sh1, sh2):
         expected = a[1] @ b[1]
         assert_dtype_allclose(result, expected)
 
+    def test_rmatmul_dpnp_array(self):
```
antonwolfy, line 3323: That tests have to be part of `test_ndarray,py`, because they are testing method of ndarray.

dpnp/tests/test_utils.py
```diff
@@ -0,0 +1,67 @@
...
+    @pytest.mark.parametrize(
+        "value",
+        [
+            42,
+            True,
+            "1",
+        ],
+    )
+    def test_valid_scalars(self, value):
+        assert dpnp.is_supported_array_or_scalar(value)
```
antonwolfy, line 29: Explicit assertion seems more clear about testing scenario.
```suggestion
        assert dpnp.is_supported_array_or_scalar(value) is True
```

dpnp/tests/test_utils.py
```diff
@@ -0,0 +1,67 @@
...
+    @pytest.mark.parametrize(
+        "array",
+        [
+            dpnp.array([1, 2, 3]),
+            dpt.asarray([1, 2, 3]),
+        ],
+    )
+    def test_synchronize_array_data(self, array):
+        try:
+            dpnp.synchronize_array_data(array)
```
antonwolfy, line 54: It would be better to submit something to the queue before calling synchronization.

dpnp/tests/test_ndarray.py
```diff
@@ -460,3 +460,29 @@ def test_clip():
...
+            self._array_obj = x.get_array()
+
+        def __matmul__(self, other):
+            return NotImplemented
+
+    d = Dummy(a)
+
+    result = d @ b
+    expected = dpnp.matmul(a, b)
+    assert_allclose(result, expected)
```
antonwolfy, line 480: both output arrays are dpnp.ndarray, so we no need to use assert util function from numpy here:
```suggestion
    expected = a @ b
    assert (result == expected).all()
```

antonwolfy (approved):
LGTM, thank you @vlad-perevezentsev

PR #2282, opened 2025-01-27
No need to extract dpnp conda package in GitHub actions

antonwolfy:
Now it is not required to extract dpnp conda package to run the test scope, because the tests will be installed during the conda package installation and so can be run with `pytest -ra --pyargs dpnp.tests` command.

The PR propose to remove obsolete steps from `Conda package` workflow.


vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy

PR #2284, opened 2025-01-27
Update third party tests to pass without fp64 aspect

antonwolfy:
Some of third party tests are intended to run towards the latest NumPy, but previously it was run only in public CI on OpenCL CPU device.
With the new `stable` workflow available in the internal CI, such scenario is also covered for L0 GPU device.
Thus the tests has to consider dtype check of result based on fp64 aspect of the device (some of GPU does not provide it).

The PR proposes to update the affected tests and to modify dtype check based on device capabilities where the result is allocating.


vtavana (approved):
Thank you, @antonwolfy!

PR #2285, opened 2025-01-27
debug: errorlevel conditionals

jharlow-intel:
Wheels tests were failing because some python abrupt failures (or windows access violations) would use an exit code that was both non-zero and non-1.

This should capture everything that isn't an exit code of 0

### DIFF
--- a/conda-recipe/run_test.bat
+++ b/conda-recipe/run_test.bat
@@ -32,10 +32,10 @@ if not defined PYTHON (
 
 
 "%PYTHON%" -c "import dpnp; print(dpnp.__version__)"
-if errorlevel 1 exit 1
+if errorlevel neq 0 exit 1
 
 "%PYTHON%" -m dpctl -f
-if errorlevel 1 exit 1
+if errorlevel neq 0 exit 1
 
 "%PYTHON%" -m pytest -ra --pyargs dpnp
-if errorlevel 1 exit 1
+if errorlevel neq 0 exit 1
--- a/conda-recipe/run_test.bat
+++ b/conda-recipe/run_test.bat
@@ -32,10 +32,10 @@ if not defined PYTHON (
 
 
 "%PYTHON%" -c "import dpnp; print(dpnp.__version__)"
-if errorlevel neq 0 exit 1
+if %errorlevel% neq 0 exit 1
 
 "%PYTHON%" -m dpctl -f
-if errorlevel neq 0 exit 1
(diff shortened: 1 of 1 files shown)


antonwolfy (approved):
Thank you @jharlow-intel for resolving that 👍

conda-recipe/run_test.bat
```diff
@@ -32,10 +32,10 @@ if not defined PYTHON (
 
 
 "%PYTHON%" -c "import dpnp; print(dpnp.__version__)"
-if errorlevel 1 exit 1
+if errorlevel neq 0 exit 1
```
antonwolfy, line 35: It seems doesn't work properly.
All tests passed there:
https://cje-fm-owrp-prod04.devtools.intel.com/satg-dap-intelpython/job/intel-packages/job/dpnp/job/dev-windows-py3.11/job/test/1493/cloudbees-pipeline-explorer/
but job returned error status.

jharlow-intel:
Hmm, I was a little afraid of that. I guess the tests can all "pass" - but due to how some tests are muted I guess it doesn't always return an exit code of 0 even when it passes.

I'll need to experiment more, and I'll share my findings with you in Teams.

We at least now know it is indeed the `run_test.bat` script that was returning a 0 code. We just need to adjust that or the pytest command within maybe.

jharlow-intel:
https://cje-fm-owrp-prod04.devtools.intel.com/satg-dap-intelpython/job/intel-packages/job/dpnp/job/dev-windows-py3.12/job/test/1194/console

Looks like it's passing _and_ failing when appropriate now 😄

antonwolfy (approved):
LGTM! Works as expected now. Thank you @jharlow-intel

jharlow-intel:
It looks like my conditional is basically identical to the ones used previously, but their is a nuance to `errorlevel` vs %errorlevel%` 

TLDR; these variables are technically not the same and `errorlevel` is sytem-level and `%errorlevel%` will _default_ to the system-level value, unless it's set manually by something else (for example, `pytest` or python itself could be setting `%errorlevel%`)

Links I read:
https://stackoverflow.com/questions/42967840/errorlevel-vs-errorlevel-vs-exclamation-mark-errorlevel-exclamation-mark
https://devblogs.microsoft.com/oldnewthing/20080926-00/?p=20743

I also peeked at conda-forge recipes and they also seem to lean into the `%ERRORLEVEL% neq 0` syntax so I think this is for the best here

PR #2302, opened 2025-02-10
Enable dpnp build on AMD GPU

vlad-perevezentsev:
This PR updates `СMakeLists` files and `build_locally.py` to enable building dpnp for AMD targets.

To build dpnp on AMD:
```
python scripts/build_locally.py --target=hip --arch=gfx90a
```

`--arch` is a required argument when `target=hip`. 
where `arch`` is the architecture of the AMD GPU.

To find the architecture, use
```
rocminfo | grep 'Name: *gfx.*'
```

To build OneMKL Interfaces on AMD , it is necessary to add `rocm` root directory to `CMAKE_PREFIX_PATH`
If `rocm` is not installed by default in `/opt/rocm` , it is necessary to set `ROCM_PATH` environment variable


CMakeLists.txt
```diff
@@ -71,31 +71,61 @@ option(DPNP_TARGET_CUDA
...
-   set(_dpnp_sycl_targets ${DPNP_SYCL_TARGETS})
+    set(_dpnp_sycl_targets ${DPNP_SYCL_TARGETS})
+    if (NOT "x${HIP_TARGETS}" STREQUAL "x")
+        set(_dpnp_amd_targets ${HIP_TARGETS})
+        set(_use_onemkl_interfaces_hip ON)
+    endif()
 endif()
 
-if(_dpnp_sycl_targets)
+if (_dpnp_sycl_targets)
```
antonwolfy, line 119: Do we support a use case when the flag is not set to any value?

dpnp/backend/extensions/indexing/CMakeLists.txt
```diff
@@ -38,12 +38,12 @@ if(_dpnp_sycl_targets)
     target_compile_options(
         ${python_module_name}
         PRIVATE
-        -fsycl-targets=${_dpnp_sycl_targets}
+        ${_dpnp_sycl_target_compile_options}
```
antonwolfy, line 41: Should we do the same for each CMakeList.txt per every extension?

scripts/build_locally.py
```diff
@@ -104,6 +105,16 @@ def run(
         # Always builds using oneMKL interfaces for the cuda target
         onemkl_interfaces = True
 
+    if target == "hip":
```
antonwolfy, line 108: What is about documentation updates?
We need to describe and provide an example how to enable build for HIP target.

scripts/build_locally.py
```diff
@@ -104,6 +105,16 @@ def run(
...
+        if not arch:
+            raise ValueError("--arch is required when --target=hip")
+        cmake_args += [
+            "-DDPNP_TARGET_HIP=ON",
```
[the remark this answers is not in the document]
ndgrigorian: Additionally, `--target=cuda` is current dpnp approach, but:

1. dpctl and dpnp should consider supporting targeting specific CUDA architectures
2. `--target=hip` means that there is no way to build simultaneously for HIP and CUDA (which is very, very much an edge case, but should be considered)

For these reasons, I think it is most sensible to move away from `--target=` universal approach to `--target-cuda=` and `--target-hip=`

scripts/build_locally.py
vlad-perevezentsev: done

scripts/build_locally.py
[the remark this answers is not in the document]
vlad-perevezentsev: @ndgrigorian it is a great suggestion. 
I have added support for `--target-hip` and I am going to add `--target-cuda` instead of `--target` in the next PR.
Thanks

dpnp/backend/extensions/indexing/CMakeLists.txt
vlad-perevezentsev: @antonwolfy mkl extensions require slightly different flags than our own extensions. 
I have added new variables `_dpnp_sycl_mkl_target_compile_options` and `_dpnp_sycl_mkl_target_link_options` for them.

doc/quick_start_guide.rst
```diff
@@ -126,18 +126,37 @@ Building ``dpnp`` for these targets requires that these CodePlay plugins be inst
...
 
 .. code-block:: bash
 
     python scripts/build_locally.py --target=cuda
 
+And for AMD devices:
+
+.. code-block:: bash
+
+    python scripts/build_locally.py --target-hip=gfx90a
```
antonwolfy, line 147: In general it might be unclear what `gfx90a` means here. It'd be great to clarify.

scripts/build_locally.py
```diff
@@ -104,6 +105,15 @@ def run(
         # Always builds using oneMKL interfaces for the cuda target
         onemkl_interfaces = True
 
+    if target_hip is not None:
+        if target_hip == "default":
```
antonwolfy, line 109: We need a special handling for `python scripts/build_locally.py --target-hip=`.
Now it is equal to `python scripts/build_locally.py`, which was not intended, I guess.

antonwolfy: The same comment is applicable to `python scripts/build_locally.py --target=`.

antonwolfy, line 109: It is a bit unclear what the use case assumed here? Is it about `python scripts/build_locally.py --target-hip="default"` only?
Then I believe the error message below needs to be rephrase a bit to something like `No default HIP architecture is supported. It must be specified explicitly`.

CMakeLists.txt
```diff
@@ -75,27 +75,64 @@ option(DPNP_USE_ONEMKL_INTERFACES
     "Build DPNP with oneMKL Interfaces"
     OFF
 )
+set(HIP_TARGETS "" CACHE STRING "HIP architecture for target")
```
antonwolfy, line 78: I assume there is no support for multiple values:
```suggestion
set(HIP_TARGET "" CACHE STRING "HIP architecture for target")
```

CMakeLists.txt
```diff
@@ -75,27 +75,64 @@ option(DPNP_USE_ONEMKL_INTERFACES
...
+            set(_dpnp_sycl_targets "amdgcn-amd-amdhsa,spir64-unknown-unknown")
+        endif()
+    endif()
 else()
-   set(_dpnp_sycl_targets ${DPNP_SYCL_TARGETS})
+    set(_dpnp_sycl_targets ${DPNP_SYCL_TARGETS})
+
+    if (NOT "x${HIP_TARGETS}" STREQUAL "x")
+        set(_dpnp_amd_targets ${HIP_TARGETS})
+        set(_use_onemkl_interfaces_hip ON)
```
antonwolfy, line 112: Do we need here something similar to above?
```python
        set(_dpnp_sycl_targets "amdgcn-amd-amdhsa,${_dpnp_sycl_targets}")
```

CMakeLists.txt
ndgrigorian: At some point, it was clear in docs that only one architecture was supported at a time, but now it isn't as clear and should be tested

Also, there is new information in the [extension guide](https://developer.codeplay.com/products/oneapi/amd/2025.1.0/guides/get-started-guide-amd#compile-for-multiple-targets)

```
The compiler driver also offers alias targets for each target+architecture pair to make the command line shorter and easier to understand for humans. Thanks to the aliases, the -Xsycl-target-backend flags no longer need to be specified.
```

It shows that the command
```
icpx -fsycl -fsycl-targets=spir64_gen,amdgcn-amd-amdhsa,nvptx64-nvidia-cuda \
        -Xsycl-target-backend=spir64_gen '-device pvc' \
        -Xsycl-target-backend=amdgcn-amd-amdhsa --offload-arch=gfx1030 \
        -Xsycl-target-backend=nvptx64-nvidia-cuda --offload-arch=sm_80 \
        -o sycl-app sycl-app.cpp
```
is equivalent to
```
icpx -fsycl -fsycl-targets=intel_gpu_pvc,amd_gpu_gfx1030,nvidia_gpu_sm_80 \
        -o sycl-app sycl-app.cpp
```
so maybe both dpctl and dpnp can simplify by removing the need for `-Xsycl-target-backend=amdgcn-amd-amdhsa --offload-arch=[X]` completely

list of aliases:
https://intel.github.io/llvm/UsersManual.html

ndgrigorian: Aliases list seems to claim only one alias is supported at a time. So probably only one architecture at once is possible? That would be my guess

scripts/build_locally.py
vlad-perevezentsev: I have update this logic
Logic for `--target=` will be handled by `--target="intel"` by default 
` --target-hip=` will raise a ValueError("--target-hip requires an architecture (e.g., gfx90a)")

CMakeLists.txt
vlad-perevezentsev: > I assume there is no support for multiple values:

I am using `HIP_TARGETS` instead of `HIP_TARGET` because oneMath requires `HIP_TARGETS` to be defined
https://github.com/uxlfoundation/oneMath/blob/4ad4dfb5db834117248ad5f8fbded5cfc1097005/CMakeLists.txt#L73

antonwolfy: Is it only declaration of cmake variable which doesn't impact oneMath, isn't that?

vlad-perevezentsev: > Is it only declaration of cmake variable which doesn't impact oneMath, isn't that?

According to [OneMath documentation](https://github.com/uxlfoundation/oneMath/blob/4ad4dfb5db834117248ad5f8fbded5cfc1097005/docs/building_the_project_with_dpcpp.rst#building-for-rocm) HIP_TARGETS must be set for ROCm builds

vlad-perevezentsev: > At some point, it was clear in docs that only one architecture was supported at a time, but now it isn't as clear and should be tested
> 
> Also, there is new information in the [extension guide](https://developer.codeplay.com/products/oneapi/amd/2025.1.0/guides/get-started-guide-amd#compile-for-multiple-targets)
> 
> > The compiler driver also offers alias targets for each target+architecture pair to make the command line shorter and easier to understand for humans. Thanks to the aliases, the -Xsycl-target-backend flags no longer need to be specified.
> 
> It shows that the command
> 
> ```
> icpx -fsycl -fsycl-targets=spir64_gen,amdgcn-amd-amdhsa,nvptx64-nvidia-cuda \
>         -Xsycl-target-backend=spir64_gen '-device pvc' \
>         -Xsycl-target-backend=amdgcn-amd-amdhsa --offload-arch=gfx1030 \
>         -Xsycl-target-backend=nvptx64-nvidia-cuda --offload-arch=sm_80 \
>         -o sycl-app sycl-app.cpp
> ```
> 
> is equivalent to
> 
> ```
> icpx -fsycl -fsycl-targets=intel_gpu_pvc,amd_gpu_gfx1030,nvidia_gpu_sm_80 \
>         -o sycl-app sycl-app.cpp
> ```
> 
> so maybe both dpctl and dpnp can simplify by removing the need for `-Xsycl-target-backend=amdgcn-amd-amdhsa --offload-arch=[X]` completely
> 
> list of aliases: https://intel.github.io/llvm/UsersManual.html

This is a great suggestion.
1. The compiler supports more than one target
2. Using aliases greatly simplifies the logic

In this PR I will implement using aliases for AMD,
In the next PR I will suggest an update for CUDA

CMakeLists.txt
```diff
@@ -75,27 +75,62 @@ option(DPNP_USE_ONEMKL_INTERFACES
     "Build DPNP with oneMKL Interfaces"
     OFF
 )
+set(HIP_TARGETS "" CACHE STRING "HIP architecture for target")
+
 set(_dpnp_sycl_targets)
+set(_use_onemkl_interfaces OFF)
```
antonwolfy, line 81: Please populate the changelog

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This release achieves 100% compliance with Python Array API specification (revis
 * Added implementation of `dpnp.bartlett` [#2366](https://github.com/IntelPython/dpnp/pull/2366)
 * Added implementation of `dpnp.convolve` [#2205](https://github.com/IntelPython/dpnp/pull/2205)
 * Added implementation of `dpnp.kaiser` [#2387](https://github.com/IntelPython/dpnp/pull/2387)
+* Support for compiling `dpnp` for specified AMD GPU architecture with use of [CodePlay oneAPI plug-in](https://developer.codeplay.com/products/oneapi/amd/home/) [#2302](https://github.com/IntelPython/dpnp/pull/2302)
```
antonwolfy, line 19:
```suggestion
* Added support to build `dpnp` for specified AMD GPU architecture using [CodePlay oneAPI plug-in](https://developer.codeplay.com/products/oneapi/amd/home/) [#2302](https://github.com/IntelPython/dpnp/pull/2302)
```

antonwolfy (approved):
Thank you @vlad-perevezentsev, in overall LGTM

CMakeLists.txt
```diff
@@ -75,27 +75,62 @@ option(DPNP_USE_ONEMKL_INTERFACES
...
+        else()
+            set(_dpnp_sycl_targets "amd_gpu_${HIP_TARGETS},${_dpnp_sycl_targets}")
+        endif()
+    endif()
 else()
-   set(_dpnp_sycl_targets ${DPNP_SYCL_TARGETS})
+    set(_dpnp_sycl_targets ${DPNP_SYCL_TARGETS})
+
+    if ("${DPNP_SYCL_TARGETS}" MATCHES "nvptx64-nvidia-cuda")
+        set(_use_onemkl_interfaces ON)
```
vlad-perevezentsev, line 107: Check if user has set `DPNP_USE_ONEMKL_INTERFACES` instead of enabling it

CMakeLists.txt
```diff
@@ -75,27 +75,62 @@ option(DPNP_USE_ONEMKL_INTERFACES
...
+        set(_use_onemkl_interfaces_cuda ON)
+    endif()
+
+    if ("${DPNP_SYCL_TARGETS}" MATCHES "amd_gpu_")
+        set(_use_onemkl_interfaces ON)
```
vlad-perevezentsev, line 112: same

CMakeLists.txt
```diff
@@ -107,13 +142,20 @@ endif()
...
         # set(ENABLE_CURAND_BACKEND True)
-        set(ENABLE_MKLGPU_BACKEND True)
-        set(ENABLE_MKLCPU_BACKEND True)
+    endif()
+    if(_use_onemkl_interfaces_hip)
+        set(ENABLE_ROCBLAS_BACKEND True)
+        set(ENABLE_ROCSOLVER_BACKEND True)
+        set(ENABLE_ROCFFT_BACKEND True)
+        # set(ENABLE_ROCRAND_BACKEND True)
     endif()
```
vlad-perevezentsev, line 160: add 'else' branch and check if `_use_onemkl_interfaces_cuda || _use_onemkl_interfaces_hip` then throw A FATAL_ERROR

antonwolfy (approved):

PR #2310, opened 2025-02-12
Disallow `minlength=None` in `dpnp.bincount`

antonwolfy:
`minlength=None` was deprecated for `numpy.bincount` since numpy `1.14` and numpy is going to raise `TypeError` exception in `2.3.0` release.

The PR propose to update implementation of `dpnp.bincount` and to disallow `minlength=None` in the same way as it will be implemented in NumPy 2.3.0.

Additionally there is explicit type check of input arrays added through `dpnp.check_supported_arrays_type` calls at validation step.

### DIFF
--- a/dpnp/tests/test_histogram.py
+++ b/dpnp/tests/test_histogram.py
@@ -587,18 +587,30 @@ class TestBincount:
     @pytest.mark.parametrize(
         "array",
         [[1, 2, 3], [1, 2, 2, 1, 2, 4], [2, 2, 2, 2]],
-        ids=["[1, 2, 3]", "[1, 2, 2, 1, 2, 4]", "[2, 2, 2, 2]"],
+        ids=["size=3", "size=6", "size=4"],
     )
-    @pytest.mark.parametrize(
-        "minlength", [0, 1, 3, 5], ids=["0", "1", "3", "5"]
-    )
-    def test_bincount_minlength(self, array, minlength):
+    @pytest.mark.parametrize("minlength", [0, 1, 3, 5])
+    def test_minlength(self, array, minlength):
         np_a = numpy.array(array)
         dpnp_a = dpnp.array(array)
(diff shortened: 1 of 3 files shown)


antonwolfy:
> Approved however there are failed tests in Array API that are not related to this PR (relates to `dpnp.linalg.qr`).

I filed #2311 to investigate that.

antonwolfy:
@AlexanderKalistratov, could you please review also?

dpnp/tests/test_histogram.py
```diff
@@ -587,19 +587,31 @@ def test_different_bins_amount(self, bins_count):
...
         dpnp_a = dpnp.array(array)
 
         expected = numpy.bincount(np_a, minlength=minlength)
         result = dpnp.bincount(dpnp_a, minlength=minlength)
         assert_allclose(expected, result)
 
+    # TODO: uncomment once numpy 2.3.0 is released
+    # @testing.with_requires("numpy>=2.3")
+    # @pytest.mark.parametrize("xp", [dpnp, numpy])
+    @pytest.mark.parametrize("xp", [dpnp])
```
AlexanderKalistratov, line 604: Probably it is better to try something like:
```
@pytest.mark.parametrize("xp", [dpnp, pytest.param(numpy, marks=pytest.mark.xfail(condition=..., reason=..., strict=True)]
```
or
```
```
@pytest.mark.parametrize("xp", [dpnp, pytest.param(numpy, marks=pytest.mark.skipif(condition=..., reason=...)]
```
```

AlexanderKalistratov (approved):

dpnp/tests/test_histogram.py
antonwolfy: Thank you for the hint, the explicit `pytest.mark.xfail` is added instead of TODO comment.

Issue #2311, opened 2025-02-12
`dpnp.linalg.qr` raises `ValueError` exception due to non-contagious input array passed to the LAPACK extension


antonwolfy:
The bellow examples caused unexpected `ValueError` exception raised:
```python
import dpnp

dpnp.__version__
# Out: '0.17.0dev6+8.g0bda96adf39'

x = dpnp.array([[[0.], [0.]]])
dpnp.linalg.qr(x, mode='complete')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 dpnp.linalg.qr(x, mode='complete')

File ~/dpnp/dpnp/linalg/dpnp_iface_linalg.py:1633, in qr(a, mode)
   1630 if mode not in ("reduced", "complete", "r", "raw"):
   1631     raise ValueError(f"Unrecognized mode {mode}")
-> 1633 return dpnp_qr(a, mode)

File ~/dpnp/dpnp/linalg/dpnp_utils_linalg.py:2463, in dpnp_qr(a, mode)
   2455 """
   2456 dpnp_qr(a, mode="reduced")
   2457
   2458 Return the qr factorization of `a` matrix.
   2459
   2460 """
   2462 if a.ndim > 2:
-> 2463     return _batched_qr(a, mode=mode)
   2465 a_usm_arr = dpnp.get_usm_ndarray(a)
   2466 a_sycl_queue = a.sycl_queue

File ~/dpnp/dpnp/linalg/dpnp_utils_linalg.py:485, in _batched_qr(a, mode)
    476 lapack_func = (
    477     "_ungqr_batch"
    478     if dpnp.issubdtype(res_type, dpnp.complexfloating)
    479     else "_orgqr_batch"
    480 )
    482 # Call the LAPACK extension function _orgqr_batch/ to generate the real
    483 # orthogonal/complex unitary matrices `Qi` of the QR factorization
    484 # for a batch of general matrices.
--> 485 ht_ev, lapack_ev = getattr(li, lapack_func)(
    486     a_sycl_queue,
    487     q.get_array(),
    488     tau_h.get_array(),
    489     m,
    490     mc,
    491     k,
    492     q_stride,
    493     tau_stride,
    494     batch_size,
    495     depends=[copy_ev],
    496 )
    497 _manager.add_event_pair(ht_ev, lapack_ev)
    499 q = q[..., :mc, :].swapaxes(-2, -1)

[...]

PR #2314, opened 2025-02-14
Fix issue gh-2311

vlad-perevezentsev:
This PR suggests a solving #2311

When the input array is `(n, m)` or `(batch_size, n, m)` where m=1 `getrf` call returns F-contig result array and `dpnp.empty_like()` is used to allocate `q`  allocates F-contig array and this raises `ValueError` exception in `orgqr` and `ungqr` functions.

Added `q` allocation always C-contig and updated tests


dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -2532,13 +2534,15 @@ def dpnp_qr(a, mode="reduced"):
             a_t,
             shape=(m, m),
             dtype=res_type,
+            order="C",
```
antonwolfy, line 2537: I guess the PR description is not fully correct, because there is `_geqrf` and `_geqrf_batch` functions but not `getrf`.

vlad-perevezentsev: Yeah, I made a mistake.
Updated the PR description

antonwolfy (approved):
Thank  you @vlad-perevezentsev, LGTM!

PR #2332, opened 2025-02-21
Update `dpnp.fft` to run on CUDA

vlad-perevezentsev:
This PR suggests fixing the current issues with `dpnp.fft.fftn()` and `dpnp.fft.rfftn()` on CUDA and removing the skip tests for them

The `incorrect result` issue for `dpnp.fft.fftn()` on cuda was because preparing the input array when `batch_fft=True` could change it to `F contiguous` array. cuFFT for correct execution requires `C contiguous` array as input.

The issue with raising `Invalid strides` error is a bug in oneMath ([631](https://github.com/uxlfoundation/oneMath/issues/631)).
As a workaround until this issue is solved it is suggested to use swap of the last two axes if the last dimension is 1 and there are multiple axes.
In this case the strides inside cuFFT are calculated correctly.

Additionally updated the arguments for `test_erf` in `skipped_tests_cuda.tbl` to skip it on cuda.


dpnp/fft/dpnp_utils_fft.py
```diff
@@ -395,13 +395,31 @@ def _fft(a, norm, out, forward, in_place, c2c, axes, batch_fft=True):
         a = dpnp.reshape(a, local_shape)
         index = 1
 
+        # cuFFT requires input arrays to be C-contiguous (row-major)
+        # for correct execution
+        if dpnp.is_cuda_backend(a) and not a.flags.c_contiguous:
+            a = dpnp.ascontiguousarray(a)
```
antonwolfy, line 401: The above call of `reshape` has `order` keyword which defaults to `"C"`.
Why is it required to have explicit separate call of `ascontiguousarray`? And why it is not granted by `reshape` call? Is it a bug or expected `reshape` behavior?

vlad-perevezentsev: In some cases we can have `a.shape=local shape` then `dpnp.reshape()` will return the array view (without copying) regardless of `order` parameter
```
a = dpnp.ones((1,2,3),order="F")
dpnp.reshape(a,(-1,2,3),order="C").flags
Out[46]: 
  C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  WRITABLE : True

```

This is the expected behavior. Similar to NumPy

PR #2339, opened 2025-02-25
Reuse `dpnp.nan_to_num` in `dpnp.nansum` and `dpnp.nanprod`

ndgrigorian:
This PR proposes the use of `nan_to_num` over `_replace_nan` in `nansum`, `nanprod`, `nancumsum`, and `nancumprod` using new internal function `_replace_nan_no_mask`.

### DIFF
--- a/dpnp/dpnp_iface_nanfunctions.py
+++ b/dpnp/dpnp_iface_nanfunctions.py
@@ -60,6 +60,31 @@ __all__ = [
 ]
 
 
+def _replace_nan_no_mask(a, val):
+    """
+    Replace NaNs in array `a` with `val`.
+
+    If `a` is of inexact type, make a copy of `a`, replace NaNs with
+    the `val` value, and return the copy together. If `a` is not of
+    inexact type, do nothing and return `a`.
+
+    Parameters
+    ----------
+    a : {dpnp.ndarray, usm_ndarray}
+        Input array.
+    val : float
+        NaN values are set to `val` before doing the operation.
+
+    Returns
+    -------
+    out : {dpnp.ndarray}
+        If `a` is of inexact type, return a copy of `a` with the NaNs
+        replaced by the fill value, otherwise return `a`.
+    """
+
+    return dpnp.nan_to_num(a, nan=val, posinf=dpnp.inf, neginf=-dpnp.inf)
+
+
 def _replace_nan(a, val):
     """
     Replace NaNs in array `a` with `val`.
@@ -315,7 +340,7 @@ def nancumprod(a, axis=None, dtype=None, out=None):
 
     """
 
-    a, _ = _replace_nan(a, 1)
+    a = _replace_nan_no_mask(a, 1.0)
     return dpnp.cumprod(a, axis=axis, dtype=dtype, out=out)
 
 
@@ -385,7 +410,7 @@ def nancumsum(a, axis=None, dtype=None, out=None):
 
     """
 
-    a, _ = _replace_nan(a, 0)
+    a = _replace_nan_no_mask(a, 0.0)
     return dpnp.cumsum(a, axis=axis, dtype=dtype, out=out)
 
 
@@ -884,7 +909,7 @@ def nanprod(
 
     """
 
-    a = dpnp.nan_to_num(a, nan=1.0, posinf=dpnp.inf, neginf=-dpnp.inf)
+    a = _replace_nan_no_mask(a, 1.0)
     return dpnp.prod(
         a,
         axis=axis,
@@ -988,7 +1013,7 @@ def nansum(
 
     """
 
-    a = dpnp.nan_to_num(a, nan=0.0, posinf=dpnp.inf, neginf=-dpnp.inf)
+    a = _replace_nan_no_mask(a, 0.0)
     return dpnp.sum(
         a,
         axis=axis,
--- a/dpnp/dpnp_iface_nanfunctions.py
+++ b/dpnp/dpnp_iface_nanfunctions.py
@@ -144,6 +144,11 @@ def nanargmax(a, axis=None, out=None, *, keepdims=False):
     the user is recommended to filter NaNs themselves and use `dpnp.argmax`
     on the filtered array.
 
+    Warning
+    -------
+    The results cannot be trusted if a slice contains only NaNs
+    and -Infs.
+
     Parameters
     ----------
     a : {dpnp.ndarray, usm_ndarray}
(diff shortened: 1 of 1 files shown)


ndgrigorian:
This relatively simple and non-invasive change improves performance significantly. On Max GPU

before:
```bash
In [1]: import dpnp

In [2]: x = dpnp.ones(3*10**8, dtype="f4")

In [3]: q = x.sycl_queue

In [4]: %timeit r = dpnp.nansum(x); q.wait()
9.37 ms ± 33.8 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit r = dpnp.nansum(x); q.wait()
9.42 ms ± 18.8 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [6]: x = dpnp.ones(10**8, dtype="f4")

In [7]: %timeit r = dpnp.nansum(x); q.wait()
4.5 ms ± 8.8 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [8]: %timeit r = dpnp.nansum(x); q.wait()
4.51 ms ± 11 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

after:
```bash
In [1]: import dpnp

In [2]: x = dpnp.ones(3*10**8, dtype="f4")

In [3]: q = x.sycl_queue

In [4]: %timeit r = dpnp.nansum(x); q.wait()
6.5 ms ± 24.5 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit r = dpnp.nansum(x); q.wait()
6.47 ms ± 35.7 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [6]: x = dpnp.ones(10**8, dtype="f4")

In [7]: %timeit r = dpnp.nansum(x); q.wait()
2.78 ms ± 14.3 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [8]: %timeit r = dpnp.nansum(x); q.wait()
2.78 ms ± 14 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

dpnp/dpnp_iface_nanfunctions.py
```diff
@@ -884,7 +884,7 @@ def nanprod(
 
     """
 
-    a, _ = _replace_nan(a, 1)
+    a = dpnp.nan_to_num(a, nan=1.0, posinf=dpnp.inf, neginf=-dpnp.inf)
```
vtavana, line 887: could we use it instead of `_replace_nan` inside `dpnp.nancumsum` and `dpnp.nancumprod` as well?

ndgrigorian: Yes, it can be.

`nanargmax` and `nanargmin` also currently raise for all-nan slices (like NumPy) at the expense of performance. CuPy handles things differently:
```python
import cupy

x = cupy.ones((2, 3, 4))
x[:, 0] = cupy.nan
# array([[[nan, nan, nan, nan],
#            [ 1.,  1.,  1.,  1.],
#            [ 1.,  1.,  1.,  1.]],

#           [[nan, nan, nan, nan],
#            [ 1.,  1.,  1.,  1.],
#            [ 1.,  1.,  1.,  1.]]])
cupy.nanargmax(x, axis=2)
# does not throw, gives
# array([[-1, 0, 0],
#        [-1, 0, 0]])
```

I think CuPy's approach is more sensible for performance.

ndgrigorian: For now, I've added this change, and kernels for `nanargmax` and `nanargmin` can be considered at another time. I will test to see if refactoring `_replace_nan` and using `nan_to_num` in these functions out-performs current implementation.

dpnp/dpnp_iface_nanfunctions.py
```diff
@@ -60,6 +60,31 @@
...
+    Parameters
+    ----------
+    a : {dpnp.ndarray, usm_ndarray}
+        Input array.
+    val : float
+        NaN values are set to `val` before doing the operation.
+
+    Returns
+    -------
+    out : {dpnp.ndarray}
```
antonwolfy, line 80: No need to brackets here
```suggestion
    out : dpnp.ndarray
```

dpnp/dpnp_iface_nanfunctions.py
```diff
@@ -60,6 +60,31 @@
...
+        If `a` is of inexact type, return a copy of `a` with the NaNs
+        replaced by the fill value, otherwise return `a`.
+    """
+
+    return dpnp.nan_to_num(a, nan=val, posinf=dpnp.inf, neginf=-dpnp.inf)
```
antonwolfy, line 85: It seems there will be a copy returned for an inexact dtype, because `copy=True` is a default value in `dpnp.nan_to_num`.

dpnp/dpnp_iface_nanfunctions.py
```diff
@@ -60,6 +60,31 @@
...
+        Input array.
+    val : float
+        NaN values are set to `val` before doing the operation.
+
+    Returns
+    -------
+    out : {dpnp.ndarray}
+        If `a` is of inexact type, return a copy of `a` with the NaNs
+        replaced by the fill value, otherwise return `a`.
+    """
```
antonwolfy, line 83: a blank line is required at the end to keep consistency with other docstrings:
```suggestion
        replaced by the fill value, otherwise return `a`.

    """
```

ndgrigorian:
Screenshot 2025-02-28 193645

The warning rendered quite nicely. I recommend we implement such a warning everywhere that a function synchronizes in some future PR.

`dpctl` may need to do this as well for, i.e., boolean indexing, repeat

dpnp/dpnp_iface_nanfunctions.py
```diff
@@ -137,6 +137,13 @@ def nanargmax(a, axis=None, out=None, *, keepdims=False):
 
     For full documentation refer to :obj:`numpy.nanargmax`.
 
+    Warning
+    -------
+    This function synchronizes in order to test for all-NaN slices in the array.
+    This may harm performance in some applications. To avoid synchronization,
+    the user is recommended to filter NaNs themselves and use `dpnp.argmax`
+    on the filtered array.
```
antonwolfy, line 145: Here is the warning in the `Returns` ([ref](https://github.com/IntelPython/dpnp/pull/2339/commits/f69ef2821d66d77f9ec74d11cff25a4b2e787232#diff-22fe858223c2cd5661c90febbce48f902c4cb8a2df0d9c0cbd8b0ecc6d23bce2R176)), should it be move here also?

antonwolfy (approved):
Thank you @ndgrigorian, LGTM!

PR #2361, opened 2025-03-05
Add support of `uint64` dtype in `dpnp.bincount`

antonwolfy:
Previously NumPy prohibits `uint64` dtype of input array in `numpy.bincount`. But that is going to be changed since `2.2.4` release (fixed in numpy#28355 and backported to 2.2.x branch in numpy#28420).

This PR proposes to align with upcoming changes and to allow `uint64` dtype of input array in `dpnp.bincount`.
Additionally, it extends the docstring of histogram-like functions to clearly notify with a warning about possible synchronization impact.


AlexanderKalistratov (approved):
LGTM

PR #2383, opened 2025-03-24
Install dpctl via pip manager in coverage and docs workflows

antonwolfy:
This PR resolves issue with `coverage` and `docs` GitHub workflows observing since 2025.1 release is published.

The reason of the proposed changes is that it isn't possible to install dpctl there as a conda package when the environment is created through installing of Intel OneAPI packages because the dpctl conda package has a runtime dependency on DPC++ RT one. Whereas the DPC++ RT package has been already installed by the apt command above and its version has been matched with the DPC++ compiler.
In case where we install the DPC++ compiler with the apt (including DPC++ RT) and install the DPC++ RT conda package while resolving dependencies, this can lead to a versioning error, i.e. compatibility issue as the DPC++ compiler only guarantees backwards compatibility, not forward compatibility (DPC++ RT may not run a binary built with a newer version of the DPC++ compiler).

Installing dpctl via the pip manager has no such limitation, as the package has no run dependency on the DPC++ RT pip package, so this is why the new step is proposed to be added to the workflows.

### DIFF
--- a/.github/workflows/build-sphinx.yml
+++ b/.github/workflows/build-sphinx.yml
@@ -135,11 +135,26 @@ jobs:
           environment-file: ${{ env.environment-file }}
           activate-environment: 'docs'
 
-      - name: Conda info
-        run: mamba info
+      # We can't install dpctl as a conda package when the environment is created through
+      # installing of Intel OneAPI packages because the dpctl conda package has a runtime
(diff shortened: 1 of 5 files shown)


vtavana (approved):
Thank you @antonwolfy, for fixing it!

PR #2386, opened 2025-03-25
Enable clip test from `array-api-tests` scope

antonwolfy:
There were few fixes implemented in `array-api-tests` around clip tests.
The PR proposes to enable `test_clip` test from `array-api-tests` scope.

### DIFF
--- a/.github/workflows/array-api-skips.txt
+++ b/.github/workflows/array-api-skips.txt
@@ -1,8 +1,5 @@
 # array API tests to be skipped
 
-# hypothesis found failures
-array_api_tests/test_operators_and_elementwise_functions.py::test_clip
-
 # unexpected result is returned - unmute when dpctl-1986 is resolved
 array_api_tests/test_operators_and_elementwise_functions.py::test_asin
 array_api_tests/test_operators_and_elementwise_functions.py::test_asinh


antonwolfy:
`test_clip` is still failing due to [dpctl-1742](https://github.com/IntelPython/dpctl/issues/1742):

```python
File "/home/runner/miniconda3/envs/array-api-conformity/lib/python3.12/site-packages/dpctl/tensor/_clip.py", line 463, in clip
    raise ValueError(
ValueError: function '<function clip at 0x7f52e1a619e0>' does not support input types (uint8, uint8, uint16), and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''.
Falsifying example: test_clip(
    x=array(0, dtype=uint8),
    data=data(...),
...
```

ndgrigorian:
> The issue is still present and caused by [dpctl-1742](https://github.com/IntelPython/dpctl/issues/1742):
> 
> ```python
> File "/home/runner/miniconda3/envs/array-api-conformity/lib/python3.12/site-packages/dpctl/tensor/_clip.py", line 463, in clip
>     raise ValueError(
> ValueError: function '<function clip at 0x7f52e1a619e0>' does not support input types (uint8, uint8, uint16), and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''.
> Falsifying example: test_clip(
>     x=array(0, dtype=uint8),
>     data=data(...),
> ...
> ```

I still think this may be a mistake in the array API tests

Per spec,
"If either min or max is an array having a different data type than x, behavior is unspecified and thus implementation-dependent."

Also,
"Clarified that behavior is only defined when x, min, and max resolve to arrays having the same data type."

ndgrigorian:
Regardless of our intended solution, we can file this with the array API tests as an incorrectly designed test.

It may be most convenient to just enable type promotion in clip after all... but perhaps we should wait for NumPy to make a strong decision on the matter.

ndgrigorian:
Issue open with array API
https://github.com/data-apis/array-api-tests/issues/359

ndgrigorian:
PR is open that will resolve the issue in array API tests
https://github.com/data-apis/array-api-tests/pull/360

PR #2389, opened 2025-03-26
Extend a number of supported types to all numeric dtypes in umath tests

antonwolfy:
Previously the umath functions were filtered out to to test only int32, int64, float32, float64 (if supported) in `test_umaths` test.
But currently all numeric dtypes in supported by dpnp. Thus it is proposed to enable all numeric dtypes there.


vtavana (approved):
Thank you, @antonwolfy!

PR #2405, opened 2025-04-02
Disable w/a for dpctl-2030 on CPU device

antonwolfy:
The PR proposes to partly merge changes from #2403 to unblock coverage GitHub action.
The workaround for dpctl-2030 was kept only for testing with GPU device (the use case from internal CI, where the changes from dpctl-2030 is not available).


vtavana (approved):
Thank you, @antonwolfy!

PR #2407, opened 2025-04-03
Skip test_digitize_empty on LNL Windows

vlad-perevezentsev:
This PR suggests skipping `test_digitize_empty - empty array case`  on LunarLake Windows as a temporary workaround to avoid crashes in the internal CI

### DIFF
--- a/dpnp/tests/test_histogram.py
+++ b/dpnp/tests/test_histogram.py
@@ -1,4 +1,5 @@
 import dpctl
+import dpctl.utils as du
 import numpy
 import pytest
 from numpy.testing import (
@@ -21,6 +22,7 @@ from .helper import (
     get_integer_dtypes,
     get_integer_float_dtypes,
     has_support_aspect64,
+    is_win_platform,
     numpy_version,
 )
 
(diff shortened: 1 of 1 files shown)


dpnp/tests/test_histogram.py
```diff
@@ -101,6 +103,17 @@ def test_digitize_empty(self, x, bins, dtype):
...
 
+        # SAT-7822: w/a to avoid crash on Windows (observing on LNL)
+        # TODO: remove the w/a once resolved
+        if x_dp.size == 0 and is_win_platform():
+            device_mask = (
+                du.intel_device_info(x_dp.sycl_device).get("device_id", 0)
+                & 0xFF00
+            )
+            # upper byte of device_id for LNL architecture
+            if device_mask in [0x6400]:
```
antonwolfy, line 114: It'd be good to have a helper function which obtains a mask for the default or input device and checks if it matches with  specified device architecture.
While for LNL arch value it'd be better to expose it as a enum value.

antonwolfy: Meanwhile it might be helpful to add a [link](https://github.com/intel/compute-runtime/blob/master/shared/source/dll/devices/devices_base.inl) to the enum description on GitHub page where to find the values of device architectures.

antonwolfy:
@vlad-perevezentsev, it seems it's not required anymore. Feel free to close the PR.

PR #2417, opened 2025-04-11
Implement `dpnp.interp()`

vlad-perevezentsev:
This PR suggests adding an implementation of `dpnp.interp()` with sycl_kernel including support for all function arguments.


dpnp/backend/kernels/elementwise_functions/interpolate.hpp
```diff
@@ -0,0 +1,67 @@
...
+                             const TValue *left,
+                             const TValue *right,
+                             TValue *out,
+                             const std::size_t n,
+                             const std::size_t xp_size,
+                             const std::vector<sycl::event> &depends)
+{
+    return q.submit([&](sycl::handler &h) {
+        h.depends_on(depends);
+        h.parallel_for(sycl::range<1>(n), [=](sycl::id<1> i) {
```
antonwolfy, line 30: It would be helpful to leave a comment why it was decided to use `sycl::id<1>`, but not `sycl::nd_item<1>` as we used in many other kernels.

dpnp/tests/test_sycl_queue.py
```diff
@@ -1453,6 +1453,24 @@ def test_choose(device):
     assert_sycl_queue_equal(result.sycl_queue, chc.sycl_queue)
 
 
+@pytest.mark.parametrize("device", valid_dev, ids=dev_ids)
+@pytest.mark.parametrize("left", [None, dpnp.array(-1.0)])
+@pytest.mark.parametrize("right", [None, dpnp.array(99.0)])
+@pytest.mark.parametrize("period", [None, dpnp.array(180.0)])
```
antonwolfy, line 1459: no need to have that as array here, because there will be `dpnp.array(..., sycl_queue=x.sycl_queue)` anyway
```suggestion
@pytest.mark.parametrize("left", [None, -1.0])
@pytest.mark.parametrize("right", [None, 99.0])
@pytest.mark.parametrize("period", [None, 180.0])
```

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
```diff
@@ -0,0 +1,300 @@
...
+                   const dpctl::tensor::usm_ndarray &idx,
+                   const dpctl::tensor::usm_ndarray &xp,
+                   const dpctl::tensor::usm_ndarray &fp,
+                   std::optional<const dpctl::tensor::usm_ndarray> &left,
+                   std::optional<const dpctl::tensor::usm_ndarray> &right,
+                   dpctl::tensor::usm_ndarray &out,
+                   sycl::queue &exec_q,
+                   const std::vector<sycl::event> &depends)
+{
+    if (x.get_size() == 0) {
```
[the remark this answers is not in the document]
antonwolfy: based on python code:
```python
    if xp.size == 0:
        raise ValueError("array of sample points is empty")
```
it seems as unsupported use case

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2742,6 +2777,180 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+        raise TypeError(
+            "Cannot cast array data from"
+            f" {x_dtype} to {x_float_type} "
+            "according to the rule 'safe'"
+        )
+
+    x = dpnp.asarray(x, dtype=x_float_type, order="C")
+    xp = dpnp.asarray(xp, dtype=x_float_type, order="C")
+
+    out_dtype = dpnp.common_type(x, xp, fp)
```
antonwolfy, line 2899: I guess the result has at least default real floating data type, isn't it?
```suggestion
    out_dtype = dpnp.common_type(x_float_type, fp)
```

dpnp/dpnp_iface_mathematical.py
```diff
@@ -348,6 +349,40 @@ def _process_ediff1d_args(arg, arg_name, ary_dtype, ary_sycl_queue, usm_type):
...
+
+    if dpnp.is_supported_array_type(param):
+        if param.ndim != 0:
+            raise ValueError(
+                f"a {name} value must be 0-dimensional, "
+                f"but got {param.ndim}-dim"
+            )
+        if dpu.get_execution_queue([exec_q, param.sycl_queue]) is None:
+            raise ValueError(
+                "input arrays and {name} must be on the same SYCL queue"
```
antonwolfy, line 369:
```suggestion
                "input arrays and {name} must be allocated on the same SYCL queue"
```

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2742,6 +2777,180 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+    x = dpnp.asarray(x, dtype=x_float_type, order="C")
+    xp = dpnp.asarray(xp, dtype=x_float_type, order="C")
+
+    out_dtype = dpnp.common_type(x, xp, fp)
+
+    fp = dpnp.asarray(fp, dtype=out_dtype, order="C")
+
+    if period is not None:
+        period = _validate_interp_param(period, "period", exec_q, usm_type)
+        if period == 0:
```
antonwolfy, line 2905: That check leads to SYCL queue synchronization. Is that expected?

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2742,6 +2777,180 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+        # normalizing periodic boundaries
+        x %= period
+        xp %= period
+        asort_xp = dpnp.argsort(xp)
+        xp = xp[asort_xp]
+        fp = fp[asort_xp]
+        xp = dpnp.concatenate((xp[-1:] - period, xp, xp[0:1] + period))
+        fp = dpnp.concatenate((fp[-1:], fp, fp[0:1]))
+        assert xp.flags.c_contiguous
+        assert fp.flags.c_contiguous
```
antonwolfy, line 2922: What is the intention of that assertions?

dpnp/dpnp_iface_mathematical.py
vlad-perevezentsev: `dpnp.common_type` takes *arrays as input argument and I can not pass `x_float_type`

dpnp/dpnp_iface_mathematical.py
vlad-perevezentsev: To make sure that `xp` and `fp` are contiguous after normalizing periodic boundaries

antonwolfy: But isn't `dpnp.concatenate` return a new array always? We never used the assertation to ensure contiguous of the data before passing to the extension and I wonder why it's needed here.

dpnp/backend/kernels/elementwise_functions/interpolate.hpp
vlad-perevezentsev: This version of kernel implementation is selected over the work-group version because it shows slightly better performance likely due to simpler execution without additional group and loop overhead

Comparison of two versions with different WorkPI on PVC

image

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
```diff
@@ -0,0 +1,296 @@
...
+    array_names names = {{&x, "x"}, {&xp, "xp"}, {&fp, "fp"}, {&out, "out"}};
+
+    auto array_types = td_ns::usm_ndarray_types();
+    int x_type_id = array_types.typenum_to_lookup_id(x.get_typenum());
+    int idx_type_id = array_types.typenum_to_lookup_id(idx.get_typenum());
+    int xp_type_id = array_types.typenum_to_lookup_id(xp.get_typenum());
+    int fp_type_id = array_types.typenum_to_lookup_id(fp.get_typenum());
+    int out_type_id = array_types.typenum_to_lookup_id(out.get_typenum());
+
+    if (x_type_id != xp_type_id) {
```
AlexanderKalistratov, line 123: I'd suggest to add additional functions to `validation_utils`:
```C++
void check_same_dtype(const array_ptr &arr1,
        const array_ptr &arr2,
        const array_names &names)
{
    if (arr1 == nullptr || arr2 == nullptr) {
        return;
    }
    auto array_types = td_ns::usm_ndarray_types();

    auto first_type_id = array_types.typenum_to_lookup_id(arr1->get_typenum());
    auto second_type_id = array_types.typenum_to_lookup_id(arr2->get_typenum());

    if (first_type_id != second_type_id) {
        py::dtype first_dtype = dtype_from_typenum(first_type_id);
        py::dtype second_dtype = dtype_from_typenum(second_type_id);

        auto first_dtype_str = py::str(first_dtype);
        auto second_dtype_str = py::str(second_dtype);

        throw py::value_error(py::str("Arrays " + name_of(arr1, names) + " and " +
                              name_of(arr2, names) + " must have the same "
                              "dtype, but got ") +
                              first_dtype_str + py::str(" and ") +
                              second_dtype_str);
    }
}

void check_same_dtype(const std::vector<array_ptr> &arrays,
    const array_names &names)
{
    if (arrays.empty()) {
        return;
    }

    auto first = arrays[0];

    for (size_t i = 1; i < arrays.size(); ++i) {
        auto second = arrays[i];
        check_same_dtype(first, second, names);
    }
}
```

And call it here

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
```diff
@@ -0,0 +1,296 @@
...
+        throw py::value_error("x and xp must have the same dtype");
+    }
+    if (fp_type_id != out_type_id) {
+        throw py::value_error("fp and out must have the same dtype");
+    }
+    if (idx_type_id != static_cast<int>(td_ns::typenum_t::INT64)) {
```
AlexanderKalistratov, line 129: You probably could add function to `validation utils`. Something like
```C++
void check_has_dtype(const array_ptr &arr,
    const typenum_t dtype,
    const array_names &names);
```

And call it here

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
```diff
@@ -0,0 +1,296 @@
...
+        throw py::value_error("The type of idx must be int64");
+    }
+
+    auto left_v = left ? &left.value() : nullptr;
+    if (left_v) {
+        names.insert({left_v, "left"});
+        if (left_v->get_ndim() != 0) {
```
AlexanderKalistratov, line 136: You can do instead:
```C++
check_num_dims(left_v, 0, names);
```

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
```diff
@@ -0,0 +1,296 @@
...
+            array_types.typenum_to_lookup_id(right_v->get_typenum());
+        if (right_type_id != fp_type_id) {
+            throw py::value_error(
+                "right must have the same dtype as fp and out");
+        }
+    }
+
+    common_checks({&x, &xp, &fp, left_v, right_v}, {&out}, names);
+
+    if (x.get_ndim() != 1 || xp.get_ndim() != 1 || fp.get_ndim() != 1 ||
```
AlexanderKalistratov, line 165: There is a single array version for `check_num_dims` but you could add vector version of it:
```C++
inline void check_num_dims(const std::vector<array_ptr> &arrays,
                           const size_t ndim,
                           const array_names &names)
{
    for (const auto &arr : arrays) {
        check_num_dims(arr, ndim, names);
    }
}
```

And then call it here:
```
check_num_dims({x, xp, fp, idx, out}, 1, names);
```

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
```diff
@@ -0,0 +1,296 @@
...
+        idx.get_ndim() != 1 || out.get_ndim() != 1)
+    {
+        throw py::value_error("All arrays must be one-dimensional");
+    }
+
+    if (xp.get_size() != fp.get_size()) {
```
AlexanderKalistratov, line 171: Add function `check_same_size` and/or `check_same_shape`?

dpnp/backend/extensions/common/ext/common.hpp
```diff
@@ -106,6 +106,18 @@ struct IsNan
...
+struct value_type_of
+{
+    using type = T;
+};
+
+template <typename T>
+struct value_type_of<std::complex<T>>
+{
+    using type = T;
+};
```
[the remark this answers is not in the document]
vlad-perevezentsev: That is good point but I  would prefer `using type = typename std::remove_cv_t<T>::value_type`  to ensure it works correctly with `const` types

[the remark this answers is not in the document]
antonwolfy: Do you mean something like?
```cpp
struct value_type_of<std::complex<T>>
{
    using type = typename std::remove_cv_t<T>::value_type;
};
```
Then, I believe, it will not work.

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
vlad-perevezentsev: done, thank you

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
vlad-perevezentsev: done, thank you

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
vlad-perevezentsev: done, thank you

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
vlad-perevezentsev: done, thank you

dpnp/backend/extensions/common/ext/common.hpp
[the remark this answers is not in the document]
vlad-perevezentsev: I actually meant removing `const` in `value_type_of_impl<T, true>`

```cpp
template<typename T>
struct value_type_of_impl<T, true>
{
    using type = typename std::remove_cv_t<T>::value_type;
};
```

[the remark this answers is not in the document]
vlad-perevezentsev: Done in https://github.com/IntelPython/dpnp/pull/2417/commits/9e06cc35968823ce9fc5eecfd63f3b146c204ea6

dpnp/dpnp_iface_mathematical.py
vlad-perevezentsev: That is good catch!
I have updated this logic to avoid synchronization
Now `period` argument supports only `scalar` and `None`

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2742,6 +2778,181 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+    period : {None, scalar}, optional
+        A period for the x-coordinates. This parameter allows the proper
+        interpolation of angular x-coordinates. Parameters `left` and `right`
+        are ignored if `period` is specified.
+
+        Default: ``None``.
+
+    Returns
+    -------
+    y : {dpnp.ndarray, usm_ndarray}
```
antonwolfy, line 2825: Do we really return `usm_ndarray`?

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2742,6 +2778,181 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+    x = dpnp.asarray(x, dtype=x_float_type, order="C")
+    xp = dpnp.asarray(xp, dtype=x_float_type, order="C")
+
+    out_dtype = dpnp.common_type(x, xp, fp)
+
+    fp = dpnp.asarray(fp, dtype=out_dtype, order="C")
+
+    if period is not None:
+        if not dpnp.isscalar(period):
+            raise TypeError(f"period must be a scalar, but got {type(period)}")
```
antonwolfy, line 2906:
```suggestion
            raise TypeError(f"period must be a scalar or None, but got {type(period)}")
```

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2742,6 +2778,181 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+        if period == 0:
+            raise ValueError("period must be a non-zero value")
+        period = _validate_interp_param(period, "period", exec_q, usm_type)
+
+        period = dpnp.abs(period)
```
antonwolfy, line 2911:
```suggestion
        period = _validate_interp_param(builtins.abs(period), "period", exec_q, usm_type)
```

dpnp/dpnp_iface_mathematical.py
```diff
@@ -348,6 +349,41 @@ def _process_ediff1d_args(arg, arg_name, ary_dtype, ary_sycl_queue, usm_type):
     return arg, usm_type
 
 
+def _validate_interp_param(param, name, exec_q, usm_type, dtype=None):
+    """
+    Validate and convert optional parameters for interpolation.
+
+    Returns a USM array or None if the input is None.
```
antonwolfy, line 356: What is the intention to return usm_ndarray instead of dpnp.ndarray?

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2742,6 +2778,181 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+
+    if not dpnp.can_cast(x_dtype, x_float_type):
+        raise TypeError(
+            "Cannot cast array data from"
+            f" {x_dtype} to {x_float_type} "
+            "according to the rule 'safe'"
+        )
+
+    x = dpnp.asarray(x, dtype=x_float_type, order="C")
+    xp = dpnp.asarray(xp, dtype=x_float_type, order="C")
```
antonwolfy, line 2898: dtype check is implemented as part of `astype`, so:
```suggestion
    x_float_type = dpnp.default_float_type(exec_q)
    x = dpnp.astype(x, x_float_type, order="C", casting="safe", copy=False)
    xp = dpnp.astype(xp, x_float_type, order="C", casting="safe", copy=False)
```

dpnp/dpnp_iface_mathematical.py
```diff
@@ -2742,6 +2778,181 @@ def gradient(f, *varargs, axis=None, edge_order=1):
...
+    One-dimensional linear interpolation.
+
+    Returns the one-dimensional piecewise linear interpolant to a function
+    with given discrete data points (`xp`, `fp`), evaluated at `x`.
+
+    For full documentation refer to :obj:`numpy.interp`.
+
+    Parameters
+    ----------
+    x : {dpnp.ndarray, usm_ndarray}
```
antonwolfy, line 2792: Do we need to state that both `x` and `xp` are expected to have a real-valued floating-point data type (see other ufuncs for a reference)?

dpnp/tests/test_mathematical.py
```diff
@@ -1143,6 +1143,152 @@ def test_complex(self, xp):
...
+        xp = dpnp.array([0, 1])
+        fp = dpnp.array([[1, 2]])
+        assert_raises(ValueError, dpnp.interp, x, xp, fp)
+
+        # xp and fp are empty
+        xp = dpnp.array([])
+        fp = dpnp.array([])
+        assert_raises(ValueError, dpnp.interp, x, xp, fp)
+
+        # x complex
```
antonwolfy, line 1261: Is there any test that `xp` can't be complex?

dpnp/tests/test_mathematical.py
```diff
@@ -1143,6 +1143,152 @@ def test_complex(self, xp):
         assert_raises((ValueError, TypeError), xp.i0, a)
 
 
+class TestInterp:
+    @pytest.mark.parametrize(
+        "dtype_x", get_all_dtypes(no_bool=True, no_complex=True)
```
antonwolfy, line 1148: Do we have tests with bool inputs?

dpnp/tests/test_mathematical.py
```diff
@@ -1143,6 +1143,152 @@ def test_complex(self, xp):
...
+    )
+    @pytest.mark.parametrize("dtype_y", get_all_dtypes(no_bool=True))
+    def test_all_dtypes(self, dtype_x, dtype_y):
+        x = numpy.linspace(0.1, 9.9, 20).astype(dtype_x)
```
antonwolfy, line 1152: Do we have any test where `x` and `xp` have different dtypes?

dpnp/tests/test_mathematical.py
antonwolfy, line 1148:
```suggestion
        "dtype_x", get_all_dtypes(no_bool=True, no_complex=True, no_none=True)
```

antonwolfy: No need to test `None` dtype here and after.

dpnp/tests/test_mathematical.py
```diff
@@ -1143,6 +1143,152 @@ def test_complex(self, xp):
         assert_raises((ValueError, TypeError), xp.i0, a)
 
 
+class TestInterp:
+    @pytest.mark.parametrize(
+        "dtype_x", get_all_dtypes(no_bool=True, no_complex=True)
+    )
+    @pytest.mark.parametrize("dtype_y", get_all_dtypes(no_bool=True))
```
antonwolfy, line 1150: Do we have a test with bool `fp` array?

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
```diff
@@ -0,0 +1,263 @@
...
+                   std::optional<const dpctl::tensor::usm_ndarray> &left,
+                   std::optional<const dpctl::tensor::usm_ndarray> &right,
+                   dpctl::tensor::usm_ndarray &out,
+                   sycl::queue &exec_q,
+                   const std::vector<sycl::event> &depends)
+{
+    common_interpolate_checks(x, idx, xp, fp, out, left, right);
+
+    if (x.get_size() == 0) {
+        return {sycl::event(), sycl::event()};
```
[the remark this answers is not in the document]
vlad-perevezentsev: I will remove it.
Thanks

dpnp/tests/test_mathematical.py
vlad-perevezentsev: Done

dpnp/dpnp_iface_mathematical.py
vlad-perevezentsev: done

dpnp/tests/test_mathematical.py
vlad-perevezentsev: done

dpnp/tests/test_mathematical.py
vlad-perevezentsev: done

dpnp/tests/test_mathematical.py
```diff
@@ -1143,6 +1143,159 @@ def test_complex(self, xp):
...
 
 
+class TestInterp:
+    @pytest.mark.parametrize(
+        "dtype_x", get_all_dtypes(no_complex=True, no_none=True)
+    )
+    @pytest.mark.parametrize(
+        "dtype_xp", get_all_dtypes(no_complex=True, no_none=True)
+    )
+    @pytest.mark.parametrize("dtype_y", get_all_dtypes())
```
antonwolfy, line 1153:
```suggestion
    @pytest.mark.parametrize("dtype_y", get_all_dtypes(no_none=True))
```

dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
```diff
@@ -0,0 +1,256 @@
...
+ * @brief A factory to define pairs of supported types for which
+ * interpolate function is available.
+ *
+ * @tparam T Type of input vector `a` and of result vector `y`.
+ */
+template <typename T>
+struct InterpolateOutputType
+{
+    using value_type = typename std::disjunction<
+        td_ns::TypeMapResultEntry<T, sycl::half>,
```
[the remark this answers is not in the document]
vlad-perevezentsev: You are right! It would be better to just delete it

dpnp/dpnp_iface_mathematical.py
```diff
@@ -348,6 +350,41 @@ def _process_ediff1d_args(arg, arg_name, ary_dtype, ary_sycl_queue, usm_type):
     return arg, usm_type
 
 
+def _validate_interp_param(param, name, exec_q, usm_type, dtype=None):
```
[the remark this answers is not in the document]
vlad-perevezentsev: done

dpnp/tests/test_mathematical.py
```diff
@@ -1197,6 +1197,23 @@ def test_left_right_args(self, dtype):
         result = dpnp.interp(ix, ixp, ifp, left=-40, right=40)
         assert_dtype_allclose(result, expected)
 
+        # left and right as ndarray
```
antonwolfy, line 1200: It can be done through `parametrize("left, right", [[-40, 40], [dpnp.array(-40), dpnp.array(40)]])` and then:
```python
expected = numpy.interp(x, xp, fp, left=get_array(numpy, left), right=get_array(numpy, right))
result = dpnp.interp(ix, ixp, ifp, left=left, right=right)
```
or something similar

dpnp/dpnp_iface_mathematical.py
```diff
@@ -348,6 +350,41 @@ def _process_ediff1d_args(arg, arg_name, ary_dtype, ary_sycl_queue, usm_type):
...
+                f"but got {param.ndim}-dim"
+            )
+        if dpu.get_execution_queue([exec_q, param.sycl_queue]) is None:
+            raise ValueError(
+                f"input arrays and {name} must be allocated "
+                "on the same SYCL queue"
+            )
+        param = param.astype(dtype)
+
+        return param
```
antonwolfy, line 375:
```suggestion
        return param.astype(dtype)
```

antonwolfy (approved):
Thank you @vlad-perevezentsev, LGTM with only a minor comment from me.

PR #2453, opened 2025-05-15
Apply w/a for expected result in spacing test only on non-Intel NumPy

antonwolfy:
The stock NumPy returns incorrect result for `numpy.spacing(-0.0)`, while Intel NumPy returns the expected one (the same as dpnp also).
It might be the case when dpnp test is running with the stock NumPy. Thus the PR add a conditional check to update expected result in `TestSpacing::test_zeros` in case when non-Intel NumPy is used.

Previously the same logic was based on NumPy version, assuming Intel NumPy has always `version < "2.0.0"`, but it is not the case anymore.


vtavana (approved):
Thank you, @antonwolfy!

PR #2458, opened 2025-05-19
Update `mkl_lapack::batch_error` handling in `getrf_batch` / `getri_batch`

vlad-perevezentsev:
This PR suggests improvements to the handling of `oneapi::mkl::lapack::batch_error` in `oneapi::mkl::lapack::getrf_batch` and `oneapi::mkl::lapack::getri_batch`

OneMKL batched functions throw a single `oneapi::mkl::lapack::batch_error` interpreted as computation_error for all failed matrices (ids()). Now set dev_info[...] = 1 for each to determinate singular matrix.

This change allows a consistent raising of `LinAlgError` in `dpnp.linalg.inv()` for both non-batched and batched singular matrices and unskipping tests with singular matrices for `dpnp.linalg.inv()` , `dpnp.linalg.det`, `dpnp.linalg.slogdet`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ This release achieves 100% compliance with Python Array API specification (revis
 * Updated `conda create` commands build and install instructions of `Quick start guide` to avoid a compilation error [#2395](https://github.com/IntelPython/dpnp/pull/2395)
 * Added handling of empty string passed to a test env variable defining data type scope as a `False` value [#2415](https://github.com/IntelPython/dpnp/pull/2415)
 * Resolved build issues on non-Intel targets in `dpnp.i0` and `dpnp.kaiser` [#2439](https://github.com/IntelPython/dpnp/pull/2439)
-* Ensure consistent `dpnp.linalg.LinAlgError` on singular matrices for both non-batched and batched cases in `dpnp.linalg.inv` [#2458] (https://github.com/IntelPython/dpnp/pull/2458)
+* Ensure consistency in the `dpnp.linalg.LinAlgError` exception raised on singular input matrices for both non-batched and batched cases in `dpnp.linalg.inv` [#2458] (https://github.com/IntelPython/dpnp/pull/2458)
 
 
 ## [0.17.0] - 02/26/2025
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ This release achieves 100% compliance with Python Array API specification (revis
 * Updated `conda create` commands build and install instructions of `Quick start guide` to avoid a compilation error [#2395](https://github.com/IntelPython/dpnp/pull/2395)
 * Added handling of empty string passed to a test env variable defining data type scope as a `False` value [#2415](https://github.com/IntelPython/dpnp/pull/2415)
 * Resolved build issues on non-Intel targets in `dpnp.i0` and `dpnp.kaiser` [#2439](https://github.com/IntelPython/dpnp/pull/2439)
+* Ensure consistent `dpnp.linalg.LinAlgError` on singular matrices for both non-batched and batched cases in `dpnp.linalg.inv` [#2458] (https://github.com/IntelPython/dpnp/pull/2458)
 
 
 ## [0.17.0] - 02/26/2025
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ This release achieves 100% compliance with Python Array API specification (revis
 * Updated `conda create` commands build and install instructions of `Quick start guide` to avoid a compilation error [#2395](https://github.com/IntelPython/dpnp/pull/2395)
 * Added handling of empty string passed to a test env variable defining data type scope as a `False` value [#2415](https://github.com/IntelPython/dpnp/pull/2415)
 * Resolved build issues on non-Intel targets in `dpnp.i0` and `dpnp.kaiser` [#2439](https://github.com/IntelPython/dpnp/pull/2439)
+* Ensure consistent `dpnp.linalg.LinAlgError` on singular matrices for both non-batched and batched cases in `dpnp.linalg.inv` [#2458] (https://github.com/IntelPython/dpnp/pull/2458)
```
antonwolfy, line 50:
```suggestion
* Ensure consistency in the `dpnp.linalg.LinAlgError` exception raised on singular input matrices for both non-batched and batched cases in `dpnp.linalg.inv` [#2458] (https://github.com/IntelPython/dpnp/pull/2458)
```

dpnp/tests/test_linalg.py
```diff
@@ -334,11 +333,7 @@ def test_nan(self, p):
         # while OneMKL returns nans
         if is_cuda_device() and p in [-dpnp.inf, -1, 1, dpnp.inf, "fro"]:
             pytest.skip("Different behavior on CUDA")
-        elif (
-            is_gpu_device()
-            and is_win_platform()
-            and p in [-dpnp.inf, -1, 1, dpnp.inf, "fro"]
-        ):
+        elif p in [-dpnp.inf, -1, 1, dpnp.inf, "fro"]:
```
antonwolfy, line 336: Do we need to check `get_intel_mkl_version` here? Since there is no need for the w/a with 2025.1 MKL.

dpnp/tests/test_linalg.py
```diff
@@ -2785,6 +2775,12 @@ def test_slogdet_strides(self):
         assert_allclose(sign_result, sign_expected)
         assert_allclose(logdet_result, logdet_expected)
 
+    # TODO: remove skipif when Intel MKL 2025.2 is released
+    # Skip running on cpu because dpnp uses _getrf_batch only on cpu.
```
antonwolfy, line 2779:
```suggestion
    # Skip running on CPU because dpnp uses _getrf_batch only on CPU.
```
The comment is quite unclear why we are skipping on CPU only, because `_getrf_batch` is used by `_batched_inv` and `_lu_factor` and there is no check that it is only when CPU device. OR do I miss something?

dpnp/tests/helper.py
```diff
@@ -372,6 +372,24 @@ def not_excluded(dtype):
...
+
+    The check is based on MKL backend name stored in Build Dependencies
+    and only applies if Intel NumPy is detected.
+    The version is extracted from the BLAS section of NumPy's build
+    information.
+
+    Return None if Intel MKL is not used.
+    """
+    if not is_intel_numpy():
+        return None
```
antonwolfy, line 387: Are use sure it is fine to compare None with strings?

antonwolfy: Would it be better to rework the helper function to something like:
```python
def requires_intel_mkl_version(version):
    if not is_intel_numpy():
        return False

    build_deps = numpy.show_config(mode="dicts")["Build Dependencies"]
    return build_deps["blas"]["version"] < version

dpnp/tests/test_linalg.py
vlad-perevezentsev: `dpnp.linalg.slogdet/det` call `getf_batch` through `_lu_factor()` which uses `getf_batch` only on CPU because of slow performance on GPU. That is why this condition was added here

`_batched_inv` calls `getf_batch` directly and `getf_batch` is used on GPU and CPU.

I have updated the comment and I will investigate the performance of `gerf_batch` on GPU again to consistent logic( In the next PR).

antonwolfy (approved):
LGTM! Thank you @vlad-perevezentsev

PR #2460, opened 2025-05-20
Clarify `dpnp.linalg.cond()` behavior with singular matrices

vlad-perevezentsev:
This PR suggests clarifying the behavior and updating `dpnp.linalg.cond()` tests for singular matrices.
It adds `Notes` section to the docs explaining the difference with NumPy when handling singular input and updates tests to reflect behavior specific to `CUDA backend` and different versions of `OneMKL`

### DIFF
--- a/dpnp/linalg/dpnp_iface_linalg.py
+++ b/dpnp/linalg/dpnp_iface_linalg.py
@@ -186,6 +186,14 @@ def cond(x, p=None):
 
         Default: ``None``.
 
+    Notes
+    -----
+    This function raises :class:`dpnp.linalg.LinAlgError` for singular input
+    matrices when using norm orders:
+    ``1``, ``-1``, ``inf``, ``-inf``, or ``'fro'``.
+    In contrast, :obj:`numpy.linalg.cond` returns ``inf``for each singular
+    matrix in the input regardless of the norm order.
+
     Returns
     -------
     out : dpnp.ndarray
--- a/dpnp/tests/test_linalg.py
+++ b/dpnp/tests/test_linalg.py
@@ -326,29 +326,62 @@ class TestCond:
     @pytest.mark.parametrize(
         "p", [None, -dpnp.inf, -2, -1, 1, 2, dpnp.inf, "fro"]
     )
-    def test_nan(self, p):
+    def test_singular_2D(self, p):
+        a = numpy.ones((2, 2))
+        ia = dpnp.array(a)
+
+        # Unlike NumPy which returns `inf` for all norm orders,
+        # DPNP raises LinAlgError for 1, -1, inf, -inf, and 'fro'
+        # due to use of gesv in 2D case.
+        # For None, 2, and -2 DPNP matches NumPy behavior.
+        if p in [None, 2, -2]:
+            result = dpnp.linalg.cond(ia, p=p)
+            expected = numpy.linalg.cond(a, p=p)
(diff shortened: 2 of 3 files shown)


dpnp/linalg/dpnp_iface_linalg.py
```diff
@@ -186,6 +186,14 @@ def cond(x, p=None):
 
         Default: ``None``.
 
+    Notes
```
antonwolfy, line 189: The section is not rendered

dpnp/linalg/dpnp_iface_linalg.py
```diff
@@ -186,6 +186,14 @@ def cond(x, p=None):
...
+    -----
+    This function raises :class:`dpnp.linalg.LinAlgError` for singular input
+    matrices when using norm orders:
+    ``1``, ``-1``, ``inf``, ``-inf``, or ``'fro'``.
+    In contrast, :obj:`numpy.linalg.cond` returns ``inf``for each singular
+    matrix in the input regardless of the norm order.
```
antonwolfy, line 195:
```suggestion
    This function will raise :class:`dpnp.linalg.LinAlgError` on singular input when using any of norm: ``1``, ``-1``, ``inf``, ``-inf``, or ``'fro'``.
    In contrast, :obj:`numpy.linalg.cond` will fill the result array with ``inf`` values on every 2D batch in input array which is singular regardless of the norm value.
```

dpnp/tests/test_linalg.py
```diff
@@ -326,29 +326,62 @@ def test_bool(self, p):
     @pytest.mark.parametrize(
         "p", [None, -dpnp.inf, -2, -1, 1, 2, dpnp.inf, "fro"]
     )
-    def test_nan(self, p):
+    def test_singular_2D(self, p):
```
antonwolfy, line 329: The coverage seems impacted

vlad-perevezentsev: It is a false positive

dpnp/tests/test_linalg.py
```diff
@@ -326,29 +326,22 @@ def test_bool(self, p):
...
         ia = dpnp.array(a)
 
-        result = dpnp.linalg.cond(ia, p=p)
-        expected = numpy.linalg.cond(a, p=p)
-        assert_dtype_allclose(result, expected)
+        # NumPy does not raise LinAlgError on singular matrices.
+        # It returns `inf`, `0`, or large/small finite values
+        # depending on the norm and the matrix content.
+        # DPNP raises LinAlgError for 1, -1, inf, -inf, and 'fro'
+        # due to use of gesv in the 2D case.
```
antonwolfy, line 337: Are you planning to add a test for > 2D case?

vlad-perevezentsev: No, numpy behavior with singular matrices varies depending on the norm value and the matrix content so it is better not to rely on it in tests. This test was added to cover case nan to inf conversion to improve code coverage

antonwolfy: What is about testing only that the exception raised?

vlad-perevezentsev: In the next PR I am going to remove  synchronization with host and device while converting nan to inf which will cause changes in this test so I will add test for ND there

antonwolfy (approved):
Thank you @vlad-perevezentsev

PR #2478, opened 2025-06-10
Add `--target-cuda` argument for selecting CUDA architecture

vlad-perevezentsev:
This PR suggests adding `--target-cuda` argument to `scripts/build_locally.py` allowing to enable CUDA support and optionally specify the target architecture (e.g. `sm_80`).
If no architecture is specified, `sm_50` is used by default.
```bash
$ python scripts/build_locally.py --target-cuda
# or
$ python scripts/build_locally.py --target-cuda=<arch>
```

The specified architecture is used to construct a SYCL alias target (e.g. `nvidia_gpu_sm_80`) and passed via `-fsycl-targets` option, following [OneAPI for NVIDIA GPUs](https://developer.codeplay.com/products/oneapi/nvidia/2025.0.0/guides/get-started-guide-nvidia#use-dpc-to-target-nvidia-gpus) 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-* Added support for selecting CUDA architecture via the `--target-cuda` option using [CodePlay oneAPI plug-in](https://developer.codeplay.com/products/oneapi/nvidia/home/) [#2478](https://github.com/IntelPython/dpnp/pull/2478)
+* Added `--target-cuda[=ARCH]` option to replace the deprecated `--target=cuda`, allowing users to build for CUDA devices with optional architecture selection using [CodePlay oneAPI plug-in](https://developer.codeplay.com/products/oneapi/nvidia/home/) [#2478](https://github.com/IntelPython/dpnp/pull/2478)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Added support for selecting CUDA architecture via the `--target-cuda` option using [CodePlay oneAPI plug-in](https://developer.codeplay.com/products/oneapi/nvidia/home/) [#2478](https://github.com/IntelPython/dpnp/pull/2478)
+
 ### Changed
 
 ### Fixed
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,17 +68,21 @@ find_package(Dpctl REQUIRED)
 message(STATUS "Dpctl_INCLUDE_DIR=" ${Dpctl_INCLUDE_DIR})
 message(STATUS "Dpctl_TENSOR_INCLUDE_DIR=" ${Dpctl_TENSOR_INCLUDE_DIR})
 
-option(DPNP_TARGET_CUDA
-    "Build DPNP to target CUDA devices"
-    OFF
-)
 option(DPNP_USE_ONEMKL_INTERFACES
     "Build DPNP with oneMKL Interfaces"
     OFF
 )
+set(DPNP_TARGET_CUDA
+    ""
+    CACHE STRING
+    "Build DPNP to target CUDA device. \
+Set to a truthy value (e.g., ON, TRUE) to use default architecture (sm_50), \
+or to a specific architecture like sm_80."
+)
 set(HIP_TARGETS "" CACHE STRING "HIP architecture for target")
 
 set(_dpnp_sycl_targets)
+set(_dpnp_cuda_arch)
 set(_use_onemkl_interfaces OFF)
 set(_use_onemkl_interfaces_cuda OFF)
 set(_use_onemkl_interfaces_hip OFF)
@@ -87,8 +91,18 @@ set(_dpnp_sycl_target_compile_options)
 set(_dpnp_sycl_target_link_options)
 
 if ("x${DPNP_SYCL_TARGETS}" STREQUAL "x")
-    if(DPNP_TARGET_CUDA)
-        set(_dpnp_sycl_targets "nvptx64-nvidia-cuda,spir64-unknown-unknown")
+    if (DPNP_TARGET_CUDA)
+        if(DPNP_TARGET_CUDA MATCHES "^sm_")
+            set(_dpnp_cuda_arch ${DPNP_TARGET_CUDA})
+        elseif(DPNP_TARGET_CUDA MATCHES "^(ON|TRUE|YES|Y|1)$")
+            set(_dpnp_cuda_arch "sm_50")
+        else()
+            message(FATAL_ERROR
+                "Invalid value for DPNP_TARGET_CUDA: \"${DPNP_TARGET_CUDA}\". "
+                "Expected 'ON', 'TRUE', 'YES', 'Y', '1', or a CUDA architecture like 'sm_80'."
+            )
+        endif()
+        set(_dpnp_sycl_targets "nvidia_gpu_${_dpnp_cuda_arch},spir64-unknown-unknown")
         set(_use_onemkl_interfaces_cuda ON)
     endif()
 
@@ -104,7 +118,7 @@ if ("x${DPNP_SYCL_TARGETS}" STREQUAL "x")
 else()
     set(_dpnp_sycl_targets ${DPNP_SYCL_TARGETS})
 
-    if ("${DPNP_SYCL_TARGETS}" MATCHES "nvptx64-nvidia-cuda")
+    if("${DPNP_SYCL_TARGETS}" MATCHES "(nvidia_gpu_sm_|nvptx64-nvidia-cuda)")
         set(_use_onemkl_interfaces_cuda ON)
     endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,7 +82,6 @@ or to a specific architecture like sm_80."
 set(HIP_TARGETS "" CACHE STRING "HIP architecture for target")
 
 set(_dpnp_sycl_targets)
-set(_dpnp_cuda_arch)
 set(_use_onemkl_interfaces OFF)
 set(_use_onemkl_interfaces_cuda OFF)
 set(_use_onemkl_interfaces_hip OFF)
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -68,17 +68,21 @@ find_package(Dpctl REQUIRED)
...
+    ""
+    CACHE STRING
+    "Build DPNP to target CUDA device. \
+Set to a truthy value (e.g., ON, TRUE) to use default architecture (sm_50), \
+or to a specific architecture like sm_80."
+)
 set(HIP_TARGETS "" CACHE STRING "HIP architecture for target")
 
 set(_dpnp_sycl_targets)
+set(_dpnp_cuda_arch)
```
antonwolfy: It seems only used inside the if-statement below at lines 95-106.
Can we moved declaration there?

CMakeLists.txt
```diff
@@ -87,8 +91,18 @@ set(_dpnp_sycl_target_compile_options)
 set(_dpnp_sycl_target_link_options)
 
 if ("x${DPNP_SYCL_TARGETS}" STREQUAL "x")
-    if(DPNP_TARGET_CUDA)
-        set(_dpnp_sycl_targets "nvptx64-nvidia-cuda,spir64-unknown-unknown")
+    if (DPNP_TARGET_CUDA)
```
antonwolfy, line 93: It is not `OFF` by default now. Should this be updated?
```suggestion
    if (NOT "x${DPNP_TARGET_CUDA}" STREQUAL "x")
```

doc/quick_start_guide.rst
```diff
@@ -144,13 +144,33 @@ installation layout of compatible version. The following plugins from CodePlay a
...
+Note that kernels are built for ``sm_50`` by default, allowing them to work on a wider
+range of architectures, but limiting the usage of more recent CUDA features.
+
+For reference, compute architecture strings like ``sm_80`` correspond to specific
+CUDA Compute Capabilities (e.g., Compute Capability 8.0 corresponds to ``sm_80``).
+A complete mapping between NVIDIA GPU models and their respective
+Compute Capabilities can be found in the official
+`CUDA GPU Compute Capability <https://developer.nvidia.com/cuda-gpus>`_ documentation.
+
+A full list of available SYCL alias targets is available in the
```
antonwolfy: It is not clear what we are saying by that sentence and how it is connected with `--target-cuda` option.
Should that be moved just after line 145?

doc/quick_start_guide.rst
```diff
@@ -144,13 +144,33 @@ installation layout of compatible version. The following plugins from CodePlay a
...
-``dpnp`` can be built for CUDA devices as follows:
+``dpnp`` can be built for CUDA devices using the ``--target-cuda`` argument.
+
+To target a specific architecture (e.g., ``sm_80``):
+
+.. code-block:: bash
+
+    python scripts/build_locally.py --target-cuda=sm_80
+
+To use the default architecture (``sm_50``), run:
```
antonwolfy, line 162: Why do we have default arch for CUDA but not for AMD?

doc/quick_start_guide.rst
```diff
@@ -144,13 +144,33 @@ installation layout of compatible version. The following plugins from CodePlay a
 Building ``dpnp`` also requires `building Data Parallel Control Library for custom SYCL targets.
 <https://intelpython.github.io/dpctl/latest/beginners_guides/installation.html#building-for-custom-sycl-targets>`_
 
-``dpnp`` can be built for CUDA devices as follows:
+``dpnp`` can be built for CUDA devices using the ``--target-cuda`` argument.
```
antonwolfy: Is it possible to add subchapters: one for CUDA build and another for AMD?

CMakeLists.txt
```diff
@@ -68,17 +68,21 @@ find_package(Dpctl REQUIRED)
...
 
-option(DPNP_TARGET_CUDA
-    "Build DPNP to target CUDA devices"
-    OFF
-)
 option(DPNP_USE_ONEMKL_INTERFACES
     "Build DPNP with oneMKL Interfaces"
     OFF
 )
+set(DPNP_TARGET_CUDA
```
antonwolfy, line 75: Please update the changelog

doc/quick_start_guide.rst
ndgrigorian: The reason is that there is a default architecture supported by the plugin for CUDA, but not for AMD

by using the default architecture, it grandfathers in the old behavior

antonwolfy: The new implementation does not fallback on the default architecture supported by the plugin for CUDA.
Previously we passed `nvptx64-nvidia-cuda`, while now it is set to `nvidia_gpu_sm_50` be default.

Or is there explicit mapping from `nvptx64-nvidia-cuda` to `nvidia_gpu_sm_50` value inside the plugin which will be kept in the future plugin's implementation?

ndgrigorian: Per the [plugin's documentation](https://developer.codeplay.com/products/oneapi/nvidia/2025.1.1/guides/get-started-guide-nvidia#compile-for-nvidia-gpus) it should be explicitly mapped to `sm_50`.

>Note that kernels are built for sm_50 by default, allowing them to work on a wider range of architectures, but limiting the usage of more recent CUDA features.

CMakeLists.txt
vlad-perevezentsev: The empty string is False for this check `if (DPNP_TARGET_CUDA)` 
I added this check in case when `DPNP_TARGET_CUDA` is passed as `0, OFF, NO, FALSE, N` via `cmake-opts` argument

doc/quick_start_guide.rst
vlad-perevezentsev: done

doc/quick_start_guide.rst
vlad-perevezentsev: Done

CMakeLists.txt
vlad-perevezentsev: done

doc/quick_start_guide.rst
```diff
@@ -144,13 +144,40 @@ installation layout of compatible version. The following plugins from CodePlay a
...
+    python scripts/build_locally.py --target-cuda=sm_80
+
+To use the default architecture (``sm_50``), run:
 
 .. code-block:: bash
 
-    python scripts/build_locally.py --target=cuda
+    python scripts/build_locally.py --target-cuda
+
+Note that kernels are built for ``sm_50`` by default, allowing them to work on a wider
```
antonwolfy:
```suggestion
Note that kernels are built for the default architecture (``sm_50``), allowing them to work on a wider
```

doc/quick_start_guide.rst
```diff
@@ -173,13 +200,15 @@ For example:
 .. code-block:: bash
     python scripts/build_locally.py --target-hip=gfx90a
 
+Multi-target build
+~~~~~~~~~~~~~~~~~~
 
 It is, however, possible to build for Intel devices, CUDA devices, and an AMD device
```
antonwolfy: I believe it would be good to clarify, that the default build from the source enables support of Intel devices. But it is possible to extend that with support of both CUDA and AMD devices.

CMakeLists.txt
antonwolfy: > in case when `DPNP_TARGET_CUDA` is passed as `0, OFF, NO, FALSE, N` via `cmake-opts` argument

That is not the case when `DPNP_TARGET_CUDA` passed as an empty string. So it's still unclear for me.
Per my understanding the string can't be empty due to the [check](https://github.com/IntelPython/dpnp/pull/2478/files#diff-2602a9a708bebc6aeac410c2b67fe353dd1ef87f895569336478772989b8bfffR102).

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Added support for selecting CUDA architecture via the `--target-cuda` option using [CodePlay oneAPI plug-in](https://developer.codeplay.com/products/oneapi/nvidia/home/) [#2478](https://github.com/IntelPython/dpnp/pull/2478)
```
antonwolfy: We have to point that `--target=cuda` was replaced with `--target-cuda`. It might be critical from the impact on the user.

CMakeLists.txt
antonwolfy: > in case when `DPNP_TARGET_CUDA` is passed as `0, OFF, NO, FALSE, N` via `cmake-opts` argument

That is not the case when `DPNP_TARGET_CUDA` passed as an empty string. So it's still unclear for me.
Per my understanding the string can't be empty due to the [check](https://github.com/IntelPython/dpnp/pull/2478/files#diff-2602a9a708bebc6aeac410c2b67fe353dd1ef87f895569336478772989b8bfffR102).

vlad-perevezentsev: You are right that `--target-cuda= ` is checked in `build_locally.py`. 
But if someone bypasses it via `--cmake-opts="-DDPNP_TARGET_CUDA="`  the empty string is still evaluated as `FALSE` in `if(DPNP_TARGET_CUDA)`. Thus this condition safely handles both cases.
Using `if (NOT "x${DPNP_TARGET_CUDA}" STREQUAL "x")` would only check for non-empty strings but still treat values like `OFF` or `0` as `TRUE`

antonwolfy: Yes, it sounds reasonable.
But what is about similar flag for AMD build? Why don't we check the same there then?

doc/quick_start_guide.rst
```diff
@@ -173,13 +200,15 @@ For example:
 .. code-block:: bash
     python scripts/build_locally.py --target-hip=gfx90a
 
+Multi-target build
+~~~~~~~~~~~~~~~~~~
 
-It is, however, possible to build for Intel devices, CUDA devices, and an AMD device
-architecture all at once:
+By default, building from source enables support for Intel devices.
+However, the build can be extended to support both CUDA and AMD devices simultaneously:
```
antonwolfy:
```suggestion
The default ``dpnp`` build from the source enables support of Intel devices only.
Extending the build with a custom SYCL target additionally enables support of CUDA or AMD device in ``dpnp``. Besides, the support can be also extended to enable both CUDA and AMD devices at the same time:
```

CMakeLists.txt
vlad-perevezentsev: @antonwolfy it will be updated in the next PR.
Thank you

antonwolfy (approved):
No more comments from me.
Thank you @vlad-perevezentsev

PR #2511, opened 2025-07-03
Fix license deprecation warning

antonwolfy:
Table-based license keys of the form `license = {text = "<license-string>"}` are deprecated since `setuptools >= 77.0.0` and produces warning:
> SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated

This PR aligns with new format of license expression ([PEP 639](https://peps.python.org/pep-0639/#add-license-expression-field)).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,8 @@ keywords = [
   "gpu",
   "dpcpp"
 ]
-license = {text = "Apache 2.0"}
+license = "Apache 2.0"
+license-files = ["LICENSE.txt"]
 maintainers = [{name = "Intel Corporation"}]
 name = "dpnp"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -66,7 +66,8 @@ keywords = [
   "gpu",
   "dpcpp"
 ]
-license = {text = "Apache 2.0"}
+license = "Apache-2.0"
```
vtavana: The license used by `dpnp` is `BSD-2-Clause` which is different than `dpctl` that uses `Apache-2.0`.
```suggestion
license = "BSD-2-Clause"
```

vtavana: image

vtavana: image

PR #2540, opened 2025-07-25
Update `dpnp.isclose` with scalar-specific SYCL kernels

vlad-perevezentsev:
This PR suggests updating `dpnp.isclose()` function adding a scalar-specific SYCL kernels for both contiguous and stride cases to improve performance when `rtol` and `atol` are scalars.
Also extends and updates tests for `dpnp.isclose()`

The new kernel **improves performance** by **up to 10x** compared to the previous implementation when `rtol` and `atol` are scalars (tested on PVC).

CPU results:
image

GPU results:
image

### DIFF
--- a/.github/workflows/openssf-scorecard.yml
+++ b/.github/workflows/openssf-scorecard.yml
@@ -7,13 +7,16 @@ on:
   # For Branch-Protection check. Only the default branch is supported. See
   # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
   branch_protection_rule:
+  # To be able to be triggered manually
+  workflow_dispatch:
   # To guarantee Maintained check is occasionally updated. See
   # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
   schedule:
     - cron: '28 2 * * 1'
     - cron: '28 2 * * 4'
   push:
-    branches: [ "master" ]
+    branches:
+      - master
 
 # Declare default permissions as read only.
 permissions: read-all
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added `timeout-minutes` property to GitHub jobs [#2526](https://github.com/IntelPython/dpnp/pull/2526)
 * Added implementation of `dpnp.ndarray.data` and `dpnp.ndarray.data.ptr` attributes [#2521](https://github.com/IntelPython/dpnp/pull/2521)
 * Added `dpnp.ndarray.__contains__` method [#2534](https://github.com/IntelPython/dpnp/pull/2534)
+* Added implementation of `dpnp.linalg.lu_factor` (SciPy-compatible) [#2557](https://github.com/IntelPython/dpnp/pull/2557), [#2565](https://github.com/IntelPython/dpnp/pull/2565)
 
 ### Changed
 
@@ -35,6 +36,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * FFT module is updated to perform in-place FFT in intermediate steps of ND FFT [#2543](https://github.com/IntelPython/dpnp/pull/2543)
 * Reused dpctl tensor include to enable experimental SYCL namespace for complex types [#2546](https://github.com/IntelPython/dpnp/pull/2546)
 * Changed Windows-specific logic in dpnp initialization [#2553](https://github.com/IntelPython/dpnp/pull/2553)
+* Added missing includes to files in ufunc and VM pybind11 extensions [#2571](https://github.com/IntelPython/dpnp/pull/2571)
+* Refactored backend implementation of `dpnp.linalg.solve` to use oneMKL LAPACK `gesv` directly [#2558](https://github.com/IntelPython/dpnp/pull/2558)
 * Improved performance of `dpnp.isclose` function by implementing a dedicated kernel for scalar `rtol` and `atol` arguments [#2540](https://github.com/IntelPython/dpnp/pull/2540)
 
 ### Deprecated
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/bitwise_count.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/bitwise_count.cpp
@@ -23,6 +23,10 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <cstdint>
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/degrees.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/degrees.cpp
@@ -23,6 +23,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/fabs.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/fabs.cpp
@@ -23,6 +23,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/fix.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/fix.cpp
@@ -23,6 +23,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/float_power.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/float_power.cpp
@@ -23,6 +23,10 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <complex>
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/fmax.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/fmax.cpp
@@ -23,6 +23,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/fmin.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/fmin.cpp
@@ -23,6 +23,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/fmod.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/fmod.cpp
@@ -23,6 +23,10 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <cstdint>
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/gcd.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/gcd.cpp
@@ -23,6 +23,10 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <cstdint>
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/heaviside.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/heaviside.cpp
@@ -23,6 +23,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/i0.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/i0.cpp
@@ -23,6 +23,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/interpolate.cpp
@@ -24,7 +24,15 @@
 //*****************************************************************************
 
 #include <complex>
+#include <cstddef>
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <type_traits>
+#include <utility>
 #include <vector>
+
+#include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
 #include <pybind11/pybind11.h>
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/lcm.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/lcm.cpp
@@ -23,6 +23,10 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <cstdint>
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/ldexp.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/ldexp.cpp
@@ -23,6 +23,10 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <cstdint>
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/logaddexp2.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/logaddexp2.cpp
@@ -23,6 +23,8 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
--- a/dpnp/backend/extensions/ufunc/elementwise_functions/radians.cpp
+++ b/dpnp/backend/extensions/ufunc/elementwise_functions/radians.cpp
@@ -23,6 +23,9 @@
 // THE POSSIBILITY OF SUCH DAMAGE.
 //*****************************************************************************
 
+#include <type_traits>
+#include <vector>
+
 #include <sycl/sycl.hpp>
 
 #include "dpctl4pybind11.hpp"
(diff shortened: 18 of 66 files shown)


dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
```diff
@@ -0,0 +1,418 @@
...
+    const py::object &, // atol
+    const py::object &, // equal_nan
+    const char *,       // a
+    const char *,       // b
+    char *,             // out
+    const std::vector<sycl::event> &);
+
+template <typename T>
+sycl::event isclose_contig_scalar_call(sycl::queue &q,
+                                       std::size_t nelems,
```
antonwolfy:
```suggestion
                                       const std::size_t nelems,
```

dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
```diff
@@ -0,0 +1,418 @@
...
+using ext::validation::check_same_size;
+
+namespace dpnp::extensions::ufunc
+{
+
+namespace impl
+{
+
+template <typename T>
+using value_type_of_t = typename value_type_of<T>::type;
```
antonwolfy: `value_type_of_t` is also part of `ext::common` namespace

dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
```diff
@@ -0,0 +1,418 @@
...
+#include <pybind11/numpy.h>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include "kernels/elementwise_functions/isclose.hpp"
+
+#include "../../elementwise_functions/simplify_iteration_space.hpp"
+
+// dpctl tensor headers
+#include "utils/memory_overlap.hpp"
```
antonwolfy: It seems not needed, part of `ext/validation_utils.hpp`
```suggestion
```

dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
```diff
@@ -0,0 +1,418 @@
...
+    check_same_dtype(&a, &b, names);
+    check_has_dtype(&res, td_ns::typenum_t::BOOL, names);
+
+    check_same_size({&a, &b, &res}, names);
+    int res_nd = res.get_ndim();
+    check_num_dims({&a, &b}, res_nd, names);
+
+    check_queue({&a, &b, &res}, names, exec_q);
+    check_no_overlap({&a, &b}, {&res}, names);
+    dpctl::tensor::validation::CheckWritable::throw_if_not_writable(res);
```
antonwolfy: there is `check_writable` available

dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
```diff
@@ -0,0 +1,418 @@
...
+
+    check_same_dtype(&a, &b, names);
+    check_has_dtype(&res, td_ns::typenum_t::BOOL, names);
+
+    check_same_size({&a, &b, &res}, names);
+    int res_nd = res.get_ndim();
+    check_num_dims({&a, &b}, res_nd, names);
+
+    check_queue({&a, &b, &res}, names, exec_q);
+    check_no_overlap({&a, &b}, {&res}, names);
```
antonwolfy, line 173: it seems there is missing call of `dpctl::tensor::overlap::SameLogicalTensors()` in `check_no_overlap`

dpnp/backend/kernels/elementwise_functions/isclose.hpp
```diff
@@ -0,0 +1,333 @@
...
+            const std::uint16_t elems_per_sg = sgSize * elems_per_wi;
+
+            const std::size_t start =
+                (gid / sgSize) * (elems_per_sg - sgSize) + gid;
+            const std::size_t end = std::min(nelems_, start + elems_per_sg);
+            for (std::size_t offset = start; offset < end; offset += sgSize) {
+                if constexpr (is_complex_v<T>) {
+                    T z_a = a_[offset];
+                    T z_b = b_[offset];
+                    bool x = isclose(z_a.real(), z_b.real(), rtol_, atol_,
```
antonwolfy: Is that mathematically correct? Since abs of complex number isn't the same as abs of both real and imag parts.

antonwolfy: There is `exprm_ns::abs` available in dpctl headers which works with complex type.

dpnp/dpnp_iface_logic.py
```diff
@@ -870,6 +872,75 @@ def isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
         rtol, atol, scalar_type=True, all_scalars=True
     )
 
+    # Use own SYCL kernel for scalar rtol/atol
+    if dpnp.isscalar(rtol) and dpnp.isscalar(atol):
+        dt = dpnp.result_type(a, b, 1.0)
```
antonwolfy: Can we move that handling to a separate function to improve code readability?

dpnp/dpnp_iface_logic.py
```diff
@@ -870,6 +872,75 @@ def isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
...
+                usm_type=usm_type,
+                sycl_queue=exec_q,
+            )
+        else:
+            usm_type, exec_q = get_usm_allocations([a, b])
+
+        a = dpnp.astype(a, dt, casting="same_kind", copy=False)
+        b = dpnp.astype(b, dt, casting="same_kind", copy=False)
+
+        # Convert complex rtol/atol to real values
```
antonwolfy: I wonder if that true, like in the example below:
```python
import numpy

numpy.__version__
# Out: '2.3.1'

a = numpy.ones(10, dtype='f8')
b = a.copy()
b[2] = 1.00000000001

s = complex(0, 10e-3)
numpy.isclose(a, b, s, s)
# Out:
# array([ True,  True, False,  True,  True,  True,  True,  True,  True, True])

s = abs(s)
numpy.isclose(a, b, s, s)
# Out:
# array([ True,  True,  True,  True,  True,  True,  True,  True,  True, True])

```

dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
vlad-perevezentsev: Good catch!
I fixed it. Thank you

dpnp/dpnp_iface_logic.py
vlad-perevezentsev: @antonwolfy based on your example most likely NumPy is using the real part of a complex number.
I fixed this and extended the tests for this case.
Thank you

dpnp/backend/kernels/elementwise_functions/isclose.hpp
vlad-perevezentsev: Good catch!
I updated the kernel isclose logic.
Thank you

dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
```diff
@@ -0,0 +1,415 @@
...
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//*****************************************************************************
+
+#include <algorithm>
```
antonwolfy: For what do we need that include?

dpnp/backend/kernels/elementwise_functions/isclose.hpp
```diff
@@ -0,0 +1,345 @@
...
+
+#pragma once
+
+#define SYCL_EXT_ONEAPI_COMPLEX
+#if __has_include(<sycl/ext/oneapi/experimental/sycl_complex.hpp>)
+#include <sycl/ext/oneapi/experimental/sycl_complex.hpp>
+#else
+#include <sycl/ext/oneapi/experimental/complex/complex.hpp>
+#endif
+
```
antonwolfy: need for `std::min`:
```suggestion

#include <algorithm>
```

dpnp/backend/kernels/elementwise_functions/isclose.hpp
```diff
@@ -0,0 +1,345 @@
...
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//*****************************************************************************
+
+#pragma once
+
+#define SYCL_EXT_ONEAPI_COMPLEX
```
antonwolfy: add similar change as in #2546

dpnp/backend/kernels/elementwise_functions/isclose.hpp
```diff
@@ -0,0 +1,345 @@
...
+{
+namespace exprm_ns = sycl::ext::oneapi::experimental;
+
+template <typename T>
+inline bool isclose(const T a,
+                    const T b,
+                    const T rtol,
+                    const T atol,
+                    const bool equal_nan)
+{
```
antonwolfy, line 51: Add a static assertion, because we do not expect an integer or bool type here
```cpp
static_assert(std::is_floating_point_v<T> || std::is_same_v<T, sycl::half>);
```

dpnp/backend/kernels/elementwise_functions/isclose.hpp
```diff
@@ -0,0 +1,345 @@
...
+    const std::size_t lws = (n_work_items_needed <= empirical_threshold)
+                                ? std::size_t(128)
+                                : std::size_t(256);
+
+    const std::size_t n_groups =
+        ((nelems + lws * elems_per_wi - 1) / (lws * elems_per_wi));
+    const auto gws_range = sycl::range<1>(n_groups * lws);
+    const auto lws_range = sycl::range<1>(lws);
+
+    // ? + offset
```
antonwolfy: What does the comment mean?

dpnp/backend/kernels/elementwise_functions/isclose.hpp
```diff
@@ -0,0 +1,345 @@
...
+          std::uint8_t vec_sz = 4u,
+          std::uint8_t n_vecs = 2u>
+sycl::event
+    isclose_contig_scalar_impl(sycl::queue &exec_q,
+                               std::size_t nelems,
+                               const scT rtol,
+                               const scT atol,
+                               const bool equal_nan,
+                               const char *a_cp,
+                               ssize_t a_offset,
```
antonwolfy: It seems we are always passing offsets as 0. Is that intentional? If yes, then it looks we don't need it. Or we missed to pass them properly?

dpnp/dpnp_iface_logic.py
```diff
@@ -82,6 +84,79 @@
...
+    # Convert complex rtol/atol to to their real parts
+    # to avoid pybind11 cast errors and match NumPy behavior
+    if isinstance(rtol, complex):
+        rtol = rtol.real
+    if isinstance(atol, complex):
+        atol = atol.real
+
+    # pylint: disable=W0707
+    try:
+        res_shape = dpnp.broadcast_shapes(a.shape, b.shape)
```
antonwolfy: Would it be helpful to use `broadcast_arrays` instead?

dpnp/dpnp_iface_logic.py
```diff
@@ -870,6 +945,10 @@ def isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
         rtol, atol, scalar_type=True, all_scalars=True
     )
 
+    # Use own SYCL kernel for scalar rtol/atol
+    if dpnp.isscalar(rtol) and dpnp.isscalar(atol):
+        return _isclose_scalar_tol(a, b, rtol, atol, equal_nan)
+
```
antonwolfy, line 943: some of the legacy code below is not properly covered by tests now

dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
```diff
@@ -0,0 +1,415 @@
...
+        shapes_equal = shapes_equal && (a_shape[i] == res_shape[i] &&
+                                        b_shape[i] == res_shape[i]);
+    }
+    if (!shapes_equal) {
+        throw py::value_error("Array shapes are not the same");
+    }
+
+    // if nelems is zero, return
+    if (nelems == 0) {
+        return std::make_pair(sycl::event(), sycl::event());
```
antonwolfy, line 197: Please cover that use case by a test

dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
```diff
@@ -0,0 +1,415 @@
...
+    py_internal::simplify_iteration_space_3(
+        nd, shape, a_strides, b_strides, res_strides,
+        // output
+        simplified_shape, simplified_a_strides, simplified_b_strides,
+        simplified_res_strides, a_offset, b_offset, res_offset);
+
+    if (nd == 1 && simplified_a_strides[0] == 1 &&
+        simplified_b_strides[0] == 1 && simplified_res_strides[0] == 1)
+    {
+        // Special case of contiguous data
```
antonwolfy: seems no test is covering that

dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
vlad-perevezentsev: done
Thank you!

dpnp/dpnp_iface_logic.py
vlad-perevezentsev: That is a good suggestion!
I have changed it

dpnp/backend/kernels/elementwise_functions/isclose.hpp
vlad-perevezentsev: Offsets are unnecessary for contig implementation,
I removed their passing and changed the function signature.
Thank you

dpnp/backend/kernels/elementwise_functions/isclose.hpp
vlad-perevezentsev: done
Thank you

dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
vlad-perevezentsev: I have removed it
Thank you

dpnp/backend/kernels/elementwise_functions/isclose.hpp
vlad-perevezentsev: done

dpnp/backend/extensions/ufunc/elementwise_functions/isclose.cpp
vlad-perevezentsev: This case is used when we keep the order of the output array as the input if `order=K` using `_empty_like_pair_orderK()` to initialize the output array. This allows us to make the strides of the output array the same as the input and use this special case.
In accordance with NumPy behavior the output array is `always C-contig` regardless of the input data and this part of the code is impossible to use
I have removed it

dpnp/dpnp_iface_logic.py
```diff
@@ -82,6 +84,74 @@
...
+    if isinstance(rtol, complex):
+        rtol = rtol.real
+    if isinstance(atol, complex):
+        atol = atol.real
+
+    # pylint: disable=W0707
+    try:
+        a, b = dpnp.broadcast_arrays(a, b)
+    except ValueError:
+        raise ValueError(
```
antonwolfy: What is the reason reissuing the same exception ?
Can we just call `a, b = dpnp.broadcast_arrays(a, b)` without any try/except?

dpnp/dpnp_iface_logic.py
```diff
@@ -82,6 +84,74 @@
...
+            "operands could not be broadcast together with shapes "
+            f"{a.shape} and {b.shape}"
+        )
+
+    out_dtype = dpnp.bool
+    output = dpnp.empty(
+        a.shape, dtype=out_dtype, sycl_queue=exec_q, usm_type=usm_type
```
antonwolfy:
```suggestion
    output = dpnp.empty(
        a.shape, dtype=dpnp.bool, sycl_queue=exec_q, usm_type=usm_type
```

dpnp/dpnp_iface_logic.py
```diff
@@ -82,6 +84,74 @@
...
+    )
+
+    _manager = dpu.SequentialOrderManager[exec_q]
+    mem_ev, ht_ev = ufi._isclose_scalar(
+        a.get_array(),
+        b.get_array(),
+        rtol,
+        atol,
+        equal_nan,
```
antonwolfy, line 142: It seems there is no check above that `equal_nan` is bool scalar. Should we add that?

dpnp/dpnp_iface_logic.py
```diff
@@ -870,6 +940,10 @@ def isclose(a, b, rtol=1e-05, atol=1e-08, equal_nan=False):
         rtol, atol, scalar_type=True, all_scalars=True
     )
 
+    # Use own SYCL kernel for scalar rtol/atol
+    if dpnp.isscalar(rtol) and dpnp.isscalar(atol):
```
antonwolfy, line 941: in above docstrings there is missing mentioning that `equal_nan` is optional

dpnp/tests/test_logic.py
```diff
@@ -515,23 +518,177 @@ def test_infinity_sign_errors(func):
...
+class TestIsClose:
+    @pytest.mark.parametrize("dtype", get_all_dtypes(no_bool=True))
+    @pytest.mark.parametrize(
+        "rtol", [1e-5, dpnp.array(1e-5), dpnp.full((10,), 1e-5)]
+    )
+    @pytest.mark.parametrize(
+        "atol", [1e-8, dpnp.array(1e-8), dpnp.full((10,), 1e-8)]
+    )
+    def test_isclose(self, dtype, rtol, atol):
+        a = numpy.random.rand(10)
```
antonwolfy: Would it be better to use `generate_random_numpy_array` which has seeding?

dpnp/tests/test_logic.py
```diff
@@ -515,23 +518,177 @@ def test_infinity_sign_errors(func):
...
-    "rtol", [1e-05, dpnp.array(1e-05), dpnp.full(10, 1e-05)]
-)
-@pytest.mark.parametrize(
-    "atol", [1e-08, dpnp.array(1e-08), dpnp.full(10, 1e-08)]
-)
-def test_isclose(dtype, rtol, atol):
-    a = numpy.random.rand(10)
-    b = a + numpy.random.rand(10) * 1e-8
+class TestIsClose:
+    @pytest.mark.parametrize("dtype", get_all_dtypes(no_bool=True))
```
antonwolfy: Do we need to exclude None from list of dtypes?

dpnp/tests/test_logic.py
```diff
@@ -515,23 +518,177 @@ def test_infinity_sign_errors(func):
...
-    dpnp_b = dpnp.array(b, dtype=dtype)
+        dpnp_a = dpnp.array(a, dtype=dtype)
+        dpnp_b = dpnp.array(b, dtype=dtype)
 
-    np_res = numpy.isclose(a, b, 1e-05, 1e-08)
-    dpnp_res = dpnp.isclose(dpnp_a, dpnp_b, rtol, atol)
-    assert_allclose(dpnp_res, np_res)
+        np_res = numpy.isclose(a, b, rtol=1e-5, atol=1e-8)
+        dpnp_res = dpnp.isclose(dpnp_a, dpnp_b, rtol=rtol, atol=atol)
+        assert_allclose(dpnp_res, np_res)
```
antonwolfy: the result is always a bool array:
```suggestion
        assert_equal(dpnp_res, np_res)
```

antonwolfy: the same below

dpnp/dpnp_iface_logic.py
vlad-perevezentsev: done
thank you

dpnp/dpnp_iface_logic.py
vlad-perevezentsev: I did this to make the error test more clear and to match numpy/cupy
```
numpy.isclose(dpnp.asnumpy(a),dpnp.asnumpy(b))
# ValueError: operands could not be broadcast together with shapes (2,2) (2,2,3) 

numpy.broadcast_arrays(dpnp.asnumpy(a),dpnp.asnumpy(b))
# ValueError: shape mismatch: objects cannot be broadcast to a single shape.  Mismatch is between arg 0 with shape (2, # 2) and arg 1 with shape (2, 2, 3).


dpnp.isclose(a,b)
# ValueError: operands could not be broadcast together with shapes (2, 2) and (2, 2, 3)

dpnp.broadcast_arrays(a,b)
# ValueError: Shape mismatch: two or more arrays have incompatible dimensions on axis (2,)
```

dpnp/dpnp_iface_logic.py
vlad-perevezentsev: I added conversion of `equal_nan` to `bool` to match NumPy and the logic for rtol/atol is not scalar

dpnp/tests/test_logic.py
vlad-perevezentsev: Done, thank you!

dpnp/tests/test_logic.py
vlad-perevezentsev: Done
Thank you!

dpnp/tests/test_logic.py
vlad-perevezentsev: Done

dpnp/dpnp_iface_logic.py
antonwolfy: We haven't match the numpy's exception text. The code would more clear without reraising `ValueError` but with numpy equivalent text.

antonwolfy (approved):
Thank you @vlad-perevezentsev for implementing the significant performance improvement. No more comments from me.

PR #2551, opened 2025-08-04
Implement `dpnp.erf` through pybind11 extension

antonwolfy:
The PR changes the implementation of `dpnp.erf` function. It now invokes `oneapi::mkl::vm` implementation from pybind11 extension of OneMKL VM if possible or uses dedicated kernel from pybind11 extension of `ufunc`.

### DIFF
--- a/dpnp/tests/test_special.py
+++ b/dpnp/tests/test_special.py
@@ -1,6 +1,6 @@
 import numpy
 import pytest
-from numpy.testing import assert_allclose
+from numpy.testing import assert_allclose, assert_equal
 
 import dpnp
 
@@ -43,6 +43,17 @@ class TestErf:
         result = dpnp.special.erf(ia)
         expected = scipy.special.erf(a)
         assert_allclose(result, expected)
+
+    def test_zeros(self):
+        import scipy.special
+
+        a = numpy.array([0.0, -0.0])
+        ia = dpnp.array(a)
+
+        result = dpnp.special.erf(ia)
+        expected = scipy.special.erf(a)
+        assert_allclose(result, expected)
+        assert_equal(dpnp.signbit(result), numpy.signbit(expected))
 
     @pytest.mark.parametrize("dt", get_complex_dtypes())
     def test_complex(self, dt):


dpnp/backend/extensions/vm/erf.cpp
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * @tparam T Type of input vector `a` and of result vector `y`.
+ */
+template <typename T>
+struct OutputType
+{
+    using value_type =
+        typename std::disjunction<td_ns::TypeMapResultEntry<T, float>,
+                                  td_ns::TypeMapResultEntry<T, double>,
+                                  td_ns::DefaultResultEntry<void>>::result_type;
```
ndgrigorian, line 73: SYCL spec defines `sycl::half` to `sycl::half` overload, is it worth adding?

dpnp/special/_erf.py
```diff
@@ -0,0 +1,123 @@
...
+            mkl_fn_to_call=mkl_fn_to_call,
+            mkl_impl_fn=mkl_impl_fn,
+        )
+
+    def __call__(self, x, out=None):  # pylint: disable=signature-differs
+        return super().__call__(x, out=out)
+
+
+_ERF_DOCSTRING = r"""
+Returns the error function of complex argument.
```
ndgrigorian: this should be rewritten: complex arguments aren't accepted by this implementation. We could at some point attempt a GPU-friendly version based on the Faddeeva function (SciPy does something similar internally), but that will take some research, and CuPy doesn't bother

dpnp/backend/extensions/vm/erf.cpp
antonwolfy: It's a bit tricky. SciPy doesn't have a dedicated path for that:
```python
scipy.special.erf.types
# Out: ['d->d', 'f->f', 'D->D', 'F->F']

scipy.special.erf(numpy.ones(10, dtype='e')).dtype
# Out: dtype('float64')
```
While in CuPy supported paths: `'e->d', 'f->f', 'd->d'`.

But it might not be the case for dpnp, because the device can't support fp64. It would mean we have to add 2 kernels to cover both paths: `e->f` and `e->d`. But I assume that sounds excess, at least for me, since it'd impact the binary size. I'd propose to keep the current implementation.

vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy for the implementation of dpnp.erf and significant cleanup of cython code!!!

dpnp/backend/extensions/vm/erf.cpp
antonwolfy: I added explicit kernel `td_ns::TypeMapResultEntry<T, sycl::half, float>` to support matrix in above `dpnp/backend/extensions/ufunc/elementwise_functions/erf.cpp` source.
While that source assumes for MKL VM extension where is no sycl::half dtype support.

dpnp/backend/kernels/elementwise_functions/erf.hpp
```diff
@@ -0,0 +1,58 @@
...
+#pragma once
+
+#include <type_traits>
+
+#include <sycl/sycl.hpp>
+
+namespace dpnp::kernels::erf
+{
+template <typename argT, typename Tp>
+struct ErfFunctor
```
ndgrigorian, line 35: to this day, I'm not sure why we didn't use inheritance with these structs from a common parent with sensible defaults. That may be the smart thing to do someday. For sure not in this PR though. Leaving this comment to remember.

dpnp/tests/test_special.py
```diff
@@ -1,32 +1,51 @@
...
+        )
+
+    def test_nan_inf(self):
+        import scipy.special
+
+        a = numpy.array([numpy.nan, -numpy.inf, numpy.inf])
+        ia = dpnp.array(a)
+
+        result = dpnp.special.erf(ia)
+        expected = scipy.special.erf(a)
```
ndgrigorian, line 44: it might be more readable and reliable to simply test against the expected values, i.e, that `erf(nan)` gives `nan`, `erf(inf)` gives 1, and `erf(-inf)` gives -1

this way bugs or changes in behavior in scipy don't impact this test

ndgrigorian: also wouldn't mind tests for 0 and -0

antonwolfy: It's the current dpnp approach to test against numpy or scipy to ensure alignment in the behavior.
Btw, I've added a dedicated test for signed zero.

PR #2553, opened 2025-08-06
Adjust Windows-only logic in `__init__.py`

ndgrigorian:
This PR suggests small changes to Windows-specific logic in `__init__.py`

`hasattr(os, "add_dll_directory")` is unnecessary after platform check and is removed, and `sys.platform == "win32"` is used which is more commonly seen in the ecosystem and avoids importing `platform` module


antonwolfy (approved):
Thank you @ndgrigorian

PR #2559, opened 2025-08-14
Unskip some linalg tests 

vlad-perevezentsev:
This PR suggests unskipping batch tests in linalg since oneMKL 2025.2 has been released and version 2025.1 is no longer relevant.
It also excludes `requires_intel_mkl_version` from coverage until it is needed.

### DIFF
--- a/dpnp/tests/helper.py
+++ b/dpnp/tests/helper.py
@@ -483,7 +483,7 @@ def numpy_version():
     return numpy.lib.NumpyVersion(numpy.__version__)
 
 
-def requires_intel_mkl_version(version):
+def requires_intel_mkl_version(version):  # pragma: no cover
     """
     Check if Intel MKL is used and its version is greater than or
     equal to the specified one.


dpnp/tests/test_linalg.py
```diff
@@ -2766,13 +2761,6 @@ def test_slogdet_strides(self):
         assert_allclose(sign_result, sign_expected)
         assert_allclose(logdet_result, logdet_expected)
 
-    # TODO: remove skipif when Intel MKL 2025.2 is released
-    # Skip running on CPU because dpnp uses _getrf_batch only on CPU
-    # for dpnp.linalg.det/slogdet.
-    @pytest.mark.skipif(
-        is_cpu_device() and not requires_intel_mkl_version("2025.2"),
-        reason="mkl<2025.2",
-    )
```
ndgrigorian, line 2775: seems that skipping on CPU was removed in this change

dpnp/tests/test_linalg.py
```diff
@@ -2803,13 +2791,6 @@ def test_slogdet_singular_matrix(self, matrix):
...
```
ndgrigorian, line 2812: same here

dpnp/tests/test_linalg.py
ndgrigorian: disregard this, I understand now that the CPU skipping was introduced with MKL version skipping

dpnp/tests/test_linalg.py
ndgrigorian: same as above, disregard

ndgrigorian (approved):
3.10 CI failure seems unrelated and unskipped tests are passing

LGTM!

PR #2563, opened 2025-08-18
Extend `getrs` with `trans_code` argument 

vlad-perevezentsev:
This PR suggests expanding `getrs` function by adding `trans_code` argument and handling its values to match the behavior of `oneapi::mkl::lapack::getrs`

These changes are the first step to implement `lu_solve(lu_and_piv, b, trans=0, overwrite_b=False, check_finite=True)` 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added implementation of `dpnp.ndarray.data` and `dpnp.ndarray.data.ptr` attributes [#2521](https://github.com/IntelPython/dpnp/pull/2521)
 * Added `dpnp.ndarray.__contains__` method [#2534](https://github.com/IntelPython/dpnp/pull/2534)
 * Added implementation of `dpnp.linalg.lu_factor` (SciPy-compatible) [#2557](https://github.com/IntelPython/dpnp/pull/2557), [#2565](https://github.com/IntelPython/dpnp/pull/2565)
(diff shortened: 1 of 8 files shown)


dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -2688,6 +2694,7 @@ def dpnp_solve(a, b):
         a_h.get_array(),
         ipiv_h.get_array(),
         b_h.get_array(),
+        trans_code,
```
antonwolfy: Here is another way to implement that.
In `lapack_py.cpp`:
```cpp
PYBIND11_MODULE(_lapack_impl, m)
{
    // Expose enum to Python
    py::enum_<oneapi::mkl::transpose>(m, "Transpose")
        .value("N", oneapi::mkl::transpose::N)
        .value("T", oneapi::mkl::transpose::T)
        .value("C", oneapi::mkl::transpose::C)
        .export_values();  // Optional, allows access like `Transpose.N`
```
in `getrs.cpp`:
```cpp
std::pair<sycl::event, sycl::event>
    getrs(sycl::queue &exec_q,
          const dpctl::tensor::usm_ndarray &a_array,
          const dpctl::tensor::usm_ndarray &ipiv_array,
          const dpctl::tensor::usm_ndarray &b_array,
          const oneapi::mkl::transpose code,
{
```
and then here:
```suggestion
        li.Transpose.T,
```

vlad-perevezentsev: Done
Thank you!

PR #2557, opened 2025-08-13
Implement `dpnp.linalg.lu_factor` 2D inputs

vlad-perevezentsev:
This PR suggests adding `dpnp.linalg.lu_factor()` for 2D arrays similar to `scipy.linalg.lu_factor` 
Support for ND inputs will be added in the next phase.

In addition, this PR includes:

1. An updated implementation of `getrf` to support non-square matrices.
2. Refactoring of `_lu_factor()` by splitting the logic into separate functions to improve readability and maintainability.

- [X] Have you provided a meaningful PR description?
- [X] Have you added a test, reproducer or referred to an issue with a reproducer?
- [X] Have you tested your changes locally for CPU and GPU devices?
- [X] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] Have you added documentation for your changes, if necessary?
- [ ] Have you added your changes to the changelog?

PR #2565, opened 2025-08-21
Implement dpnp.linalg.lu_factor batch inputs

vlad-perevezentsev:
This PR suggests extending `dpnp.linalg.lu_factor()`  #2557  for batch arrays

In addition, this PR includes:

1. An updated implementation of getrf_batch to support non-square matrices.


dpnp/tests/test_linalg.py
```diff
@@ -2013,9 +2013,118 @@ def test_check_finite_raises(self, bad):
...
+    @staticmethod
+    def _apply_pivots_rows(A_dp, piv_dp):
+        m = A_dp.shape[0]
+        rows = dpnp.arange(m)
+        for i in range(int(piv_dp.shape[0])):
+            r = int(piv_dp[i].item())
+            if i != r:
+                tmp = rows[i].copy()
+                rows[i] = rows[r]
+                rows[r] = tmp
```
ndgrigorian: is it possible to make `rows` on the host and then swap `rows[i]` and `rows[r]` without all of the additional copies on the device?

since `rows` is just being used as an index after all, maybe something like
```
rows = list(range(m))
...

rows[i], rows[r] = rows[r], rows[i]
```
then `A_dp` can be indexed with `rows`

dpnp/tests/test_linalg.py
```diff
@@ -2013,9 +2013,118 @@ def test_check_finite_raises(self, bad):
...
+        lu, piv = dpnp.linalg.lu_factor(a_stride, check_finite=False)
+        for i in range(a_stride.shape[0]):
+            L, U = self._split_lu(lu[i], 3, 3)
+            PA = self._apply_pivots_rows(
+                a_stride[i].astype(L.dtype, copy=False), piv[i]
+            )
+            assert_allclose(L @ U, PA, rtol=1e-6, atol=1e-6)
+
+    def test_singular_matrix(self):
+        a = dpnp.zeros((3, 2, 2), dtype=dpnp.float64)
```
ndgrigorian:
```suggestion
        a = dpnp.zeros((3, 2, 2), dtype=dpnp.float32)
```

dpnp/tests/test_linalg.py
```diff
@@ -2013,9 +2013,118 @@ def test_check_finite_raises(self, bad):
...
+        a[0] = dpnp.array([[1.0, 2.0], [2.0, 4.0]])
+        a[1] = dpnp.eye(2)
+        a[2] = dpnp.array([[1.0, 1.0], [1.0, 1.0]])
+        with pytest.warns(RuntimeWarning, match="Singular matrix"):
+            dpnp.linalg.lu_factor(a, check_finite=False)
+
+    def test_check_finite_raises(self):
+        a = dpnp.ones((2, 3, 3), dtype=dpnp.float64, order="F")
```
ndgrigorian:
```suggestion
        a = dpnp.ones((2, 3, 3), dtype=dpnp.float32, order="F")
```

dpnp/tests/test_linalg.py
ndgrigorian: otherwise fails if double-precision is not supporter

dpnp/tests/test_linalg.py
vlad-perevezentsev: Good suggestion. Thank you!
I applied it

dpnp/tests/test_linalg.py
vlad-perevezentsev: done, thank you!

dpnp/tests/test_linalg.py
vlad-perevezentsev: done, thank you!

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -396,6 +398,172 @@ def _batched_lu_factor(a, res_type):
...
+        )
+        _manager.add_event_pair(ht_ev, copy_ev)
+
+        ipiv_vecs[i] = dpnp.empty(
+            (k,),
+            dtype=dpnp.int64,
+            order="C",
+            usm_type=a_usm_type,
+            sycl_queue=a_sycl_queue,
+        )
```
ndgrigorian: is there any reason not to perform single copies and allocations, then just make views along the batch axis?

that is, copy all of `src` into one new array, and allocate all `ipiv_vecs` arrays as one allocation, then split both along batch axis in the loop

dpnp/linalg/dpnp_utils_linalg.py
```diff
@@ -396,6 +398,172 @@ def _batched_lu_factor(a, res_type):
...
+    a_usm_arr = dpnp.get_usm_ndarray(a)
+
+    # Initialize lists for storing arrays and events for each batch
+    a_vecs = [None] * batch_size
+    ipiv_vecs = [None] * batch_size
+    dev_info_vecs = [None] * batch_size
+    dep_evs = _manager.submitted_events
+
+    # Process each batch
+    for i in range(batch_size):
```
ndgrigorian: it may not make sense in this PR, but it might be worth looking at moving this entire loop into C++, a custom pybind11 function that takes lists of arrays instead and loops, calling `_getrf` on each

dpnp/linalg/dpnp_utils_linalg.py
vlad-perevezentsev: Good point!
I have updated `_batched_lu_factor_scipy()` and see a performance improvement
```
=== dtype: f4 ===
DPNP (GPU, New):
156 ms ± 1.26 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
DPNP (GPU, Old):
193 ms ± 836 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)

=== dtype: f8 ===
DPNP (GPU, New):
183 ms ± 626 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
DPNP (GPU, Old):
233 ms ± 212 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)

=== dtype: c8 ===
DPNP (GPU, New):
209 ms ± 9.63 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
DPNP (GPU, Old):
250 ms ± 3.22 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

=== dtype: c16 ===
DPNP (GPU, New):
247 ms ± 552 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)
DPNP (GPU, Old):
300 ms ± 7.73 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

dpnp/linalg/dpnp_utils_linalg.py
vlad-perevezentsev: Makes sense to try to implement this in another PR

PR #2574, opened 2025-09-03
Remove workflow trigger on change in branch protection rules

antonwolfy:
There is no need to trigger workflows with `Autoupdate pre-commit` and `Run tests suite` on any change in branch protection rules. That trigger was intended only to `Scorecard supply-chain security` workflow.
Thus the PR removes that triggers.


vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy

PR #2588, opened 2025-09-23
Added implementation of `dpnp.special.erfc`

antonwolfy:
The PR add the implementation of `dpnp.special.erfc` function, including tests coverage.

Also it reworks the previous implementation of erf function to reduce future code duplication. because most of erf-like function has the same types supported and similar signature.


dpnp/backend/kernels/elementwise_functions/erf.hpp
```diff
@@ -31,28 +31,44 @@
 
 namespace dpnp::kernels::erf
```
vlad-perevezentsev: It would be clearer to use for example `erfs`  as the namespace name for all erf functions

antonwolfy: Renamed

PR #2623, opened 2025-10-15
Exclude headers with SYCL kernels from the coverage report

antonwolfy:
The PR excludes the headers files under `backend/kernels/elementwise_functions` from the coverage report, because there is no way to cover structures used inside the SYCL kernels.

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -29,7 +29,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python: ['3.10', '3.11', '3.12', '3.13']
         os: [ubuntu-22.04, windows-2022]
 
     permissions:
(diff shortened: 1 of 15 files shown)


scripts/gen_coverage.py
```diff
@@ -146,6 +143,7 @@ def find_objects():
                 "export",
                 "-format=lcov",
                 "-ignore-filename-regex=/tmp/icpx*",
+                r"-ignore-filename-regex=.*/backend/kernels/elementwise_functions/.*\.hpp$",
```
ndgrigorian, line 146: there are also kernels in `extensions` which will have to be ignored

antonwolfy: I'll do that in the follow up PR, because they have to be separated to dedicated headers first. Otherwise it's impossible to ignore.

vlad-perevezentsev (approved):
LGTM!
Thank you @antonwolfy 
A small suggestion for the next PR: 
Could you also exclude `dpnp/to_numba` folder from the coverage report?

antonwolfy:
> Could you also exclude `dpnp/to_numba` folder from the coverage report?

I think the folder has to be removed at all. It can be done in the follow up PR.

PR #2642, opened 2025-10-28
Increase swap size up to 8GB in GitHub runners

antonwolfy:
There are out-of-memory issues observing when running dpnp tests in parallel with `pytest-xdist` and `-n auto`.
The default swap size was set to 4 GB for Linux runners and it seems not enough to have the parallel tests run, which caused:
> node down: Not properly terminated

The PR proposes to increase the swap size for Linux runners up to 8 GB. That makes the tests run more stable.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2643, opened 2025-10-29
Explicitly set python version in GitHub workflow for testing with `oneMath`

antonwolfy:
The PR updates `conda-incubator/setup-miniconda` step in the GH workflow for testing with `oneMath` to explicitly set python version the same as defined in testing matrix, otherwise the latest available python will be taken.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2644, opened 2025-10-30
Use GitHub action to upload the coverage data to `coveralls.io`

antonwolfy:
The PR replaces the old implementation through manual submitting `.json` file to `coveralls.io` with `coverallsapp/github-action` step which does everything automatically.

### DIFF
--- a/.github/workflows/openssf-scorecard.yml
+++ b/.github/workflows/openssf-scorecard.yml
@@ -72,6 +72,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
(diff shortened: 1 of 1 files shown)


scripts/gen_coverage.py
```diff
@@ -96,8 +96,7 @@ def run(
             "pyproject.toml",
             "--cov",
             "dpnp",
-            "--cov-report",
-            "term-missing",
+            "--cov-report=lcov:coverage-python.lcov",
```
vlad-perevezentsev, line 100: `pytest-cov` supports multiple `--cov-report` arguments so it would be better to keep `--cov-report=term-missing` as well
Then we can still see uncovered lines directly in CI logs if coveralls reports are unavailable

vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2648, opened 2025-11-04
Update cmake configs for dependent oneAPI components

antonwolfy:
The PR updates cmake files for DPC++, oneMKL, oneDPL and TBB components with 2025.3 release changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Also, that release drops support for Python 3.9, making Python 3.10 the minimum 
 * Added implementation of `dpnp.scipy.special.erfcx` [#2596](https://github.com/IntelPython/dpnp/pull/2596)
 * Added implementation of `dpnp.scipy.special.erfinv` and `dpnp.scipy.special.erfcinv` [#2624](https://github.com/IntelPython/dpnp/pull/2624)
 * Enabled support of Python 3.14 [#2631](https://github.com/IntelPython/dpnp/pull/2631)
+* Added implementation of `dpnp.ndarray.tolist` method [#2652](https://github.com/IntelPython/dpnp/pull/2652)
 
 ### Changed
 
--- a/doc/reference/io.rst
+++ b/doc/reference/io.rst
@@ -1,0 +1,76 @@
+.. currentmodule:: dpnp
+
+Input and output
+================
+
+.. hint:: `NumPy API Reference: Input and output <https://numpy.org/doc/stable/reference/routines.io.html>`_
+
+.. NumPy binary files (npy, npz)
+.. -----------------------------
+.. .. autosummary::
+..    :toctree: generated/
+..    :nosignatures:
+
+..    load
+..    save
+..    savez
+..    savez_compressed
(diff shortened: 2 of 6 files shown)


CMakeLists.txt
```diff
@@ -45,36 +45,28 @@ set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE BOTH)
...
-    find_package(TBB REQUIRED PATHS ${CMAKE_SOURCE_DIR}/dpnp/backend/cmake/Modules NO_DEFAULT_PATH)
-endif()
+# find_package() search order:
+# 1. User-specified overrides (-DMKL_DIR=..., cache variables, package root variables like MKLROOT)
+# 2. Environment variables (ENV{MKLROOT}, CMAKE_PREFIX_PATH which is automatically set by conda)
+# 3. PATHS (if specified, but without NO_DEFAULT_PATH)
+# 4. HINTS
+# 5. System prefixes (/usr, /usr/local, etc.)
+# 6. CMake built-in modules
+find_package(IntelSYCL REQUIRED HINTS ${CMAKE_SOURCE_DIR}/dpnp/backend/cmake/Module)
```
vlad-perevezentsev: Probably a typo in the path, should be `.../Modules` here and below

CMakeLists.txt
```diff
@@ -45,36 +45,28 @@ set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE BOTH)
...
-find_package(TBB QUIET)
-if(TBB_FOUND)
-    find_package(TBB REQUIRED)
-else()
-    find_package(TBB REQUIRED PATHS ${CMAKE_SOURCE_DIR}/dpnp/backend/cmake/Modules NO_DEFAULT_PATH)
-endif()
+# find_package() search order:
+# 1. User-specified overrides (-DMKL_DIR=..., cache variables, package root variables like MKLROOT)
+# 2. Environment variables (ENV{MKLROOT}, CMAKE_PREFIX_PATH which is automatically set by conda)
+# 3. PATHS (if specified, but without NO_DEFAULT_PATH)
```
vlad-perevezentsev: It is not quite right.
It should be `CMake-specific environment variables` like `CMAKE_PREFIX_PATH` , not `PATHS`
According to the [official documentation,](https://cmake.org/cmake/help/latest/command/find_package.html#id21) the PATHS argument belongs to step 9

vlad-perevezentsev: Or remove this because 2. says the same thing

CMakeLists.txt
vlad-perevezentsev: You can define a dedicated variable for `Modules` directory like for example `DPNP_CMAKE_MODULES_DIR = "${CMAKE_SOURCE_DIR}/dpnp/backend/cmake/Modules")` and use it here and below

CMakeLists.txt
antonwolfy: Reworked

vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2649, opened 2025-11-04
Remove wildcard imports in `dpnp.fft` and `dpnp.random` submodules

vlad-perevezentsev:
This PR proposes to remove wildcard imports from `dpnp.fft` and `dpnp.random` submodules and register them as explicit submodules in `dpnp` namespace.

It also removes wildcard import of `dpnp.memory` from `dpnp_iface.py`


dpnp/fft/__init__.py
```diff
@@ -153,7 +153,26 @@
 
 """
 
-from dpnp.fft.dpnp_iface_fft import *
-from dpnp.fft.dpnp_iface_fft import __all__ as __all__fft
+from .dpnp_iface_fft import __all__ as __all__fft
```
antonwolfy: Why do we need `__all__` in `dpnp/fft/dpnp_iface_fft.py` then?
Can we move it here?

vlad-perevezentsev: Done, thank you for the suggestion

PR #2650, opened 2025-11-04
Deprecate `dpnp.asfarray()` and raise a DeprecationWarning

vlad-perevezentsev:
This PR marks `dpnp.asfarray()` as deprecated to align with NumPy>=2.0 where this function was removed ([Release Notes Numpy 2.0](https://numpy.org/doc/2.0/release/2.0.0-notes.html))

A `DeprecationWarning` is now raised informing users that the function will be removed in a future release and suggesting to use `dpnp.asarray(dtype=...)` instead.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -90,7 +90,7 @@ repos:
     -   id: clang-format
         args: ["-i"]
 -   repo: https://github.com/gitleaks/gitleaks
-    rev: v8.28.0
+    rev: v8.29.0
     hooks:
     -   id: gitleaks
 -   repo: https://github.com/jumanjihouse/pre-commit-hooks


dpnp/dpnp_iface_manipulation.py
```diff
@@ -906,6 +906,14 @@ def asfarray(a, dtype=None, *, device=None, usm_type=None, sycl_queue=None):
```
antonwolfy, line 911: We need to add a warning to the docstring that:
> This function is deprecated in favor of :obj:`dpnp.asarray` and will be removed in a future release.

dpnp/dpnp_iface_manipulation.py
```diff
@@ -906,6 +906,14 @@ def asfarray(a, dtype=None, *, device=None, usm_type=None, sycl_queue=None):
 
     """
 
+    warnings.warn(
```
antonwolfy, line 914: I think it might be helpful to create a lingering doc page like CuPy has: `docs/source/reference/_deprecated.rst`. It would list all the active deprecations.
So the functions from that page will keep its docstring rendered, but not indexed.

antonwolfy: `asfarray` has to be removed from `doc/reference/array-manipulation.rst`, since deprecated

antonwolfy: All of that can be done in the follow up PR

dpnp/dpnp_iface_manipulation.py
```diff
@@ -906,6 +906,14 @@ def asfarray(a, dtype=None, *, device=None, usm_type=None, sycl_queue=None):
...
+        "`dpnp.asfarray` is deprecated, "
+        "and will be removed in a future release. "
+        "Please use `dpnp.asarray` with an appropriate dtype instead.",
+        DeprecationWarning,
+        stacklevel=2,
```
antonwolfy, line 919: `dpnp/tests/test_arraymanipulation.py::TestAsfarray::test_asfarray_coverage` is now raising a deprecation warning, please update the test to catch and check that.

dpnp/dpnp_iface_manipulation.py
antonwolfy, line 911: We have to remove links on :obj:`dpnp.asfarray` from the docstrings of other functions

dpnp/dpnp_iface_manipulation.py
vlad-perevezentsev: I introduced  a new `_deprecated.rst` and moved `dpnp.asfarray` there

dpnp/dpnp_iface_manipulation.py
```diff
@@ -894,6 +894,11 @@ def asfarray(a, dtype=None, *, device=None, usm_type=None, sycl_queue=None):
     out : dpnp.ndarray
         The input `a` as a float ndarray.
 
+    Warning
+    -------
+    This function is deprecated in favor of :obj:dpnp.asarray and
```
antonwolfy:
```suggestion
    This function is deprecated in favor of :obj:`dpnp.asarray` and
```

dpnp/tests/test_arraymanipulation.py
```diff
@@ -32,6 +32,7 @@ def test_asfarray2(self, dtype, data, data_dtype):
         assert_array_equal(result, expected)
 
     # This is only for coverage with NumPy 2.0 and above
+    @pytest.mark.filterwarnings("ignore::DeprecationWarning")
```
antonwolfy: We need a test which checks the warning raised

vlad-perevezentsev: Done!

PR #2656, opened 2025-11-07
Add implementation of `dpnp.ndarray.tobytes` method

antonwolfy:
The PR adds implementation of `dpnp.ndarray.tobytes` method converting the array to `numpy.ndarray` and calling `.tobytes(...)` for the result.

### DIFF
--- a/dpnp/dpnp_iface_indexing.py
+++ b/dpnp/dpnp_iface_indexing.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # *****************************************************************************
 # Copyright (c) 2016, Intel Corporation
 # All rights reserved.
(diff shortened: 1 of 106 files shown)


dpnp/dpnp_iface_indexing.py
```diff
@@ -396,7 +396,7 @@ def compress(condition, a, axis=None, out=None):
 
     Examples
     --------
-    >>> import numpy as np
+    >>> import dpnp as np
```
vlad-perevezentsev, line 398: Please update the same line in `def poisson` and `def rayleigh` as well

antonwolfy: It was intended to keep that as it is now. The functions needs to be fully reworked and the whole docstrings have to be updated there, not only examples.

PR #2665, opened 2025-11-13
Unify dpnp public API exports by refactoring `dpnp_iface.py`

vlad-perevezentsev:
This PR proposes to refactor the export structure of `dpnp_iface.py` by moving all python function exports into `dpnp/__init__.py`
Redundant `__all__` declarations and wildcard imports inside `dpnp_iface.py` have been removed.

This is the first step toward unifying dpnp public API exports and eliminating wildcard imports.
A next PR will remove the remaining wildcard imports in `__init__.py` and finalize the unified function export

- [X] Have you provided a meaningful PR description?
- [X] Have you added a test, reproducer or referred to an issue with a reproducer?
- [X] Have you tested your changes locally for CPU and GPU devices?
- [ ] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] Have you added documentation for your changes, if necessary?
- [ ] Have you added your changes to the changelog?

PR #2666, opened 2025-11-14
Unify dpnp public API exports by refactoring `__init__.py`

vlad-perevezentsev:
This PR removes the remaining wildcard imports and `__all__` declarations from internal modules and consolidates the corresponding exports in `dpnp/__init__.py` as part of the ongoing public API unification.

This work follows #2665.

### DIFF
--- a/dpnp/__init__.py
+++ b/dpnp/__init__.py
@@ -68,7 +68,6 @@ with warnings.catch_warnings():
 
 from .dpnp_array import dpnp_array as ndarray
 from .dpnp_array_api_info import __array_namespace_info__
-from .dpnp_flatiter import flatiter as flatiter
 from ._version import get_versions
 from . import exceptions as exceptions
 from . import fft as fft
@@ -308,6 +307,7 @@ from .dpnp_iface_indexing import (
     triu_indices_from,
     unravel_index,
 )
+from .dpnp_flatiter import flatiter
 
 # -----------------------------------------------------------------------------
 # Linear algebra


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ Also, that release drops support for Python 3.9, making Python 3.10 the minimum
 * Redesigned `dpnp.modf` function to be a part of `ufunc` and `vm` pybind11 extensions [#2654](https://github.com/IntelPython/dpnp/pull/2654)
 * Refactored `dpnp.fft` and `dpnp.random` submodules by removing wildcard imports and defining explicit public exports [#2649](https://github.com/IntelPython/dpnp/pull/2649)
 * Added support for the `out` keyword to accept a tuple, bringing ufunc signatures into alignment with those in NumPy [#2664](https://github.com/IntelPython/dpnp/pull/2664)
+* Unified `dpnp` public API exports by consolidating function exports in `__init__.py` and removing wildcard imports [#2665](https://github.com/IntelPython/dpnp/pull/2665) [#2666](https://github.com/IntelPython/dpnp/pull/2666)
```
antonwolfy, line 40: Can we remove `"--disable=c-extension-no-member"` from `.pre-commit-config.yaml` now?

antonwolfy: And E501 skipping in `.flake8`

dpnp/__init__.py
```diff
@@ -77,6 +74,81 @@
...
+# Constants (borrowed from NumPy)
+# -----------------------------------------------------------------------------
+from .dpnp_iface_types import (
+    e,
+    euler_gamma,
+    inf,
+    nan,
+    newaxis,
+    pi,
+)
```
antonwolfy, line 138: It has to be a part of the below `Routines`

dpnp/__init__.py
```diff
@@ -77,6 +74,81 @@
...
+    uint64,
+    uintc,
+    uintp,
+    unsignedinteger,
+    ushort,
+    ulonglong,
+)
+
+# -----------------------------------------------------------------------------
+# Constants (borrowed from NumPy)
```
antonwolfy:
```suggestion
# Constants
```

dpnp/__init__.py
```diff
@@ -77,6 +74,81 @@
...
 
+# =============================================================================
+# Data types, constants and type-related helpers
+# =============================================================================
+
+# -----------------------------------------------------------------------------
+# Data types (borrowed from NumPy)
+#
+# The order of these declarations are borrowed from the NumPy document:
+# https://numpy.org/doc/stable/reference/arrays.scalars.html
```
antonwolfy: It does not really make any sense, because there is a single `from .dpnp_iface_types import` below which is connected with that.

antonwolfy: The order is defined by the python sort linter and not by the pointing numpy's page.

dpnp/__init__.py
```diff
@@ -77,6 +74,81 @@
...
+    inf,
+    nan,
+    newaxis,
+    pi,
+)
+
+# -----------------------------------------------------------------------------
+# Type-related helper functions
+# -----------------------------------------------------------------------------
+from .dpnp_iface_types import (
```
antonwolfy: That is a part of `Data type routines` from https://numpy.org/doc/stable/reference/routines.dtype.html.
Why isn't it moved below to the routines?

dpnp/__init__.py
```diff
@@ -77,6 +74,81 @@
...
+    common_type,
+    finfo,
+    iinfo,
+    isdtype,
+    issubdtype,
+    is_type_supported,
```
antonwolfy: That needs to be moved away. It's a kind of deprecated API which we should not expose.
Probably it would be better to move that to `dpnp/random/dpnp_iface_random.py` (where it is only used) and to make it an internal function.

dpnp/__init__.py
```diff
@@ -77,6 +74,81 @@
...
+    newaxis,
+    pi,
+)
+
+# -----------------------------------------------------------------------------
+# Type-related helper functions
+# -----------------------------------------------------------------------------
+from .dpnp_iface_types import (
+    common_type,
+    finfo,
```
antonwolfy: Do we need to move `dtype` here instead of the above line#97 ?

dpnp/__init__.py
```diff
@@ -77,6 +74,81 @@
...
+    iinfo,
+    isdtype,
+    issubdtype,
+    is_type_supported,
+)
```
antonwolfy: It seems `result_type` has be a part of that block, if we are following the routines page.

antonwolfy: And `can_cast` also

dpnp/__init__.py
vlad-perevezentsev: Done
Thank you!

CHANGELOG.md
vlad-perevezentsev: Probably you meant `F403` : ‘from module import *’ used
Since some files that are to be removed in the future still contain wildcard imports I would prefer not to change flake8 config
```
dpnp/dpnp_algo/__init__.py:29:1: F403 'from dpnp.dpnp_algo.dpnp_algo import *' used; unable to detect undefined names
from dpnp.dpnp_algo.dpnp_algo import *
^
dpnp/dpnp_utils/__init__.py:29:1: F403 'from dpnp.dpnp_utils.dpnp_algo_utils import *' used; unable to detect undefined names
from dpnp.dpnp_utils.dpnp_algo_utils import *
^
dpnp/random/dpnp_iface_random.py:44:1: F403 'from dpnp.dpnp_algo import *' used; unable to detect undefined names
from dpnp.dpnp_algo import *
^
dpnp/random/dpnp_iface_random.py:45:1: F403 'from dpnp.dpnp_utils import *' used; unable to detect undefined names
from dpnp.dpnp_utils import *
^
dpnp/random/dpnp_iface_random.py:47:1: F403 'from .dpnp_algo_random import *' used; unable to detect undefined names
from .dpnp_algo_random import *
^
5     F403 'from dpnp.dpnp_algo.dpnp_algo import *' used; unable to detect undefined names
5
```

antonwolfy: I'm fine to remove that in the follow up PR

PR #2673, opened 2025-11-17
Remove `shape=None` default in `dpnp.reshape`

vlad-perevezentsev:
This PR updates the signature of  `dpnp.reshape` by removing `None`  default value for `shape` argument
and requiring it to be passed explicitly according to [Python array API ](https://data-apis.org/array-api/draft/API_specification/generated/array_api.reshape.html#reshape)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Also, that release drops support for Python 3.9, making Python 3.10 the minimum 
 * Extended `pre-commit` configuration with `pyupgrade`, `actionlint`, and `gersemi` hooks [#2658](https://github.com/IntelPython/dpnp/pull/2658)
 * Added implementation of `dpnp.ndarray.tobytes` method [#2656](https://github.com/IntelPython/dpnp/pull/2656)
 * Added implementation of `dpnp.ndarray.__format__` method [#2662](https://github.com/IntelPython/dpnp/pull/2662)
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Also, that release drops support for Python 3.9, making Python 3.10 the minimum
 * Refactored `dpnp.fft` and `dpnp.random` submodules by removing wildcard imports and defining explicit public exports [#2649](https://github.com/IntelPython/dpnp/pull/2649)
 * Added support for the `out` keyword to accept a tuple, bringing ufunc signatures into alignment with those in NumPy [#2664](https://github.com/IntelPython/dpnp/pull/2664)
 * Unified public API definitions in `dpnp.linalg` and `dpnp.scipy` submodules [#2663](https://github.com/IntelPython/dpnp/pull/2663)
+* Aligned the signature of `dpnp.reshape` function with Python array API by requiring an explicit `shape` argument [#2673](https://github.com/IntelPython/dpnp/pull/2673)
```
antonwolfy: What do you mean by "an explicit `shape` argument"?
Acc to Python array API the parameter is not optional and can be passed as either positional or keyword argument.
While the current implementation the `shape` was optional, since it had a default value.

antonwolfy: In general it has been implemented that way only to deprecate the legacy `newshape` argument.

antonwolfy (approved):
Thank you @vlad-perevezentsev

PR #2716, opened 2026-01-09
Remove use of `python setup.py develop/install`

antonwolfy:
A warning that `python setup.py develop` is deprecated and will no longer be supported at the end of October, 2025 has been ongoing for some time when using dpnp build driver scripts, and can be seen in the CI.

The PR proposes instead relying on `pip` for installing dpnp in scripts, and reworks the scripts to maintain use of `scikit-build`.
In the future, this will also simplify a transition to `scikit-build-core`.

This PR also introduces options
* `--clean` and `--skip-editable` to `build_locally` driver
* `--clean` and `--skip-pytest` to `gen_coverage` driver

Also it includes update of the documentation with new build instruction.


CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ Also, that release drops support for Python 3.9, making Python 3.10 the minimum
 * Unified `dpnp` public API exports by consolidating function exports in `__init__.py` and removing wildcard imports [#2665](https://github.com/IntelPython/dpnp/pull/2665) [#2666](https://github.com/IntelPython/dpnp/pull/2666)
 * Updated tests to reflect the new scalar conversion rules for non-0D `usm_ndarray` [#2694](https://github.com/IntelPython/dpnp/pull/2694)
 * Compile indexing extension with `-fno-sycl-id-queries-fit-in-int` to support huge arrays [#2721](https://github.com/IntelPython/dpnp/pull/2721)
+* Changed the build scripts and documentation due to `python setup.py develop` deprecation notice [#2172](https://github.com/IntelPython/dpnp/pull/2172)
```
vlad-perevezentsev:
```suggestion
* Changed the build scripts and documentation due to `python setup.py develop` deprecation notice [#2716](https://github.com/IntelPython/dpnp/pull/2716)
```

CMakeLists.txt
```diff
@@ -107,7 +107,7 @@ set(DPNP_TARGET_CUDA
 Set to a truthy value (e.g., ON, TRUE) to use default architecture (sm_50), \
 or to a specific architecture like sm_80."
 )
-set(HIP_TARGETS "" CACHE STRING "HIP architecture for target")
+set(DPNP_TARGET_HIP "" CACHE STRING "HIP architecture for target")
```
vlad-perevezentsev, line 110: I know that AMD build in dpnp are no longer supported but according to `OneMath` docs the ROCm target actitecture should be set via `HIP_TARGETS`
https://uxlfoundation.github.io/oneMath/building_the_project_with_dpcpp.html#building-for-rocm

scripts/_build_helper.py
```diff
@@ -0,0 +1,243 @@
...
+        c_compiler = "icx"
+    if cxx_compiler is None:
+        cxx_compiler = "icpx" if is_linux else "icx"
+
+    for name, opt_name in (
+        (c_compiler, "--c-compiler"),
+        (cxx_compiler, "--cxx-compiler"),
+    ):
+        if os.path.isabs(name):
+            path = name
```
vlad-perevezentsev, line 94: this logic validates `path` but returns `c_compiler, cxx_compiler` regardlesss. Should it return resolved absolute paths instead?

scripts/_build_helper.py
```diff
@@ -0,0 +1,243 @@
...
+def get_dpctl_cmake_dir():
+    """
+    If dpctl is locally built using `script/build_locally.py`, it is needed
+    to pass the -DDpctl_ROOT=$(python -m dpctl --cmakedir) during the build.
+    If dpctl is conda installed, it is optional to pass this parameter.
+
+    """
+
+    process = subprocess.Popen(
+        ["python", "-m", "dpctl", "--cmakedir"],
```
vlad-perevezentsev: Why does this function use `python` instead of `sys.executable` while `def build_extensions()` and `def install_editable()` already use it?

scripts/build_locally.py
```diff
@@ -26,290 +26,211 @@
 # THE POSSIBILITY OF SUCH DAMAGE.
 # *****************************************************************************
 
+import argparse
 import os
-import subprocess
 import sys
-import warnings
 
-warnings.simplefilter("default", DeprecationWarning)
```
vlad-perevezentsev, line 34: What was the reason for removing it without adding it back in `_build_helper.py`?
`onemkl_interfaces` options are still supported  there and emit `DepricationWarning`

scripts/gen_coverage.py
```diff
@@ -26,67 +26,174 @@
...
+        help="Path to compiler installation root",
+    )
+    p.add_argument(
+        "--oneapi",
+        dest="oneapi",
+        action="store_true",
+        help="Use default oneAPI compiler layout",
+    )
+    p.add_argument(
+        "--bin-llvm",
```
vlad-perevezentsev, line 71: `args.bin_llvm` is never used

scripts/_build_helper.py
antonwolfy: Aligned, thanks

scripts/gen_coverage.py
antonwolfy: Great catch! The handling is added.

scripts/build_locally.py
antonwolfy: We should not emit the deprecation warning. We do want to highlight the options are deprecated to the user.
I have no clue why it was silenced previously.

vlad-perevezentsev: `DeprecationWarning` is usually filtered out by default so users may never see this warning without keeping
`warnings.simplefilter("default", DeprecationWarning)` 

log without
```
$ python scripts/build_locally.py --onemkl_interfaces
[build_locally] Found DPCTL CMake dir: /nfs/site/home/vperevez/work/miniforge3/envs/dpnp_conda_2_4/lib/python3.13/site-packages/dpctl/resources/cmake
[build_locally] Using CMake args:
```
log with it in `_build_helper.py`
```
$ python scripts/build_locally.py --onemkl_interfaces
[build_locally] Found DPCTL CMake dir: /nfs/site/home/vperevez/work/miniforge3/envs/dpnp_conda_2_4/lib/python3.13/site-packages/dpctl/resources/cmake
/localdisk/work/vperevez/work/dpnp/scripts/_build_helper.py:121: DeprecationWarning: Using 'onemkl_interfaces' is deprecated. Please use 'onemath' instead.
  warnings.warn(
[build_locally] Using CMake args:
```

scripts/_build_helper.py
antonwolfy: Right, updated

scripts/build_locally.py
antonwolfy: I'd propose to use `stacklevel=2` instead.

CMakeLists.txt
antonwolfy: Added setting of `HIP_TARGETS` variable

scripts/gen_coverage.py
```diff
@@ -26,67 +26,182 @@
...
+
+    compiler_dir = os.path.join(bin_dir, "compiler")
+    if os.path.exists(compiler_dir):
+        bin_llvm = compiler_dir
+    else:
+        bin_dir = os.path.dirname(bin_dir)
+        bin_llvm = os.path.join(bin_dir, "bin-llvm")
+
+    if not os.path.exists(bin_llvm):
+        raise RuntimeError(f"--bin-llvm value {bin_llvm} not found")
```
vlad-perevezentsev: This looks unclear since this error can happen even the user did not pass `--bin-llvm`

antonwolfy: The exception text is updated.

PR #2722, opened 2026-01-15
Refactor `dpnp.fix()` to reuse `dpnp.trunc` internally

vlad-perevezentsev:
This PR rewrites `dpnp.fix` to reuse `dpnp.trunc` internally removing the backend implementation of `dpnp.fix` 

### DIFF
--- a/.github/workflows/build-sphinx.yml
+++ b/.github/workflows/build-sphinx.yml
@@ -47,7 +47,7 @@ jobs:
 
     steps:
       - name: Cancel Previous Runs
-        uses: styfle/cancel-workflow-action@85880fa0301c86cca9da44039ee3bb12d3bedbfa # 0.12.1
+        uses: styfle/cancel-workflow-action@3155a141048f8f89c06b4cdae32e7853e97536bc # 0.13.0
         with:
           access_token: ${{ github.token }}
 
--- a/.github/workflows/check-onemath.yaml
+++ b/.github/workflows/check-onemath.yaml
@@ -34,7 +34,7 @@ jobs:
 
     steps:
       - name: Cancel Previous Runs
-        uses: styfle/cancel-workflow-action@85880fa0301c86cca9da44039ee3bb12d3bedbfa # 0.12.1
+        uses: styfle/cancel-workflow-action@3155a141048f8f89c06b4cdae32e7853e97536bc # 0.13.0
         with:
           access_token: ${{ github.token }}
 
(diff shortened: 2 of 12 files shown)


dpnp/tests/test_mathematical.py
```diff
@@ -2607,6 +2605,15 @@ def test_out_usm_ndarray(self, func, dt):
         assert result.get_array() is usm_out
         assert_array_equal(result, expected)
 
+    @pytest.mark.parametrize("xp", [numpy, dpnp])
+    def test_out_wrong_tuple_len(self, xp, func):
```
antonwolfy: What is the point to move the test from `TestUfunc`  scope to `TestRoundingFuncs`?

vlad-perevezentsev: `dpnp.fix` no longer exposes `.nin` and `.nout` so I removed it from `TestUfunc` and added `test_out_wrong_tuple_len` to `TestRoundingFuncs` to keep the coverage and extend this case for rounding functions

antonwolfy: But there is no increase in the coverage. It tests exactly the same functionally of `TestUfunc` class, but does that the 2nd duplicating time.
I still doubt we need that test here.

dpnp/dpnp_iface_mathematical.py
```diff
@@ -1813,66 +1812,57 @@ def ediff1d(ary, to_end=None, to_begin=None):
...
+    array([ 2.,  2., -2., -2.])
+    """
 
-fix = DPNPFix(
-    "fix",
-    ufi._fix_result_type,
-    ufi._fix,
-    _FIX_DOCSTRING,
-)
+    return trunc(x, out=out)
```
antonwolfy: Why do we need to limit `fix` implementation by only `out` keyword and don't expose `nin` and `nout` anymore?
Would it be possible and better to have something like:
```python
fix = trunc
fix.__doc__ = ...
fix.name_ = "fix"
```
It will allow to keep all attributes of ufunc function

vlad-perevezentsev: Kept `dpnp.fix` as ufunc-like function and reverted the changes in the tests
Thanks

dpnp/tests/test_mathematical.py
vlad-perevezentsev: Updated. thanks

PR #2732, opened 2026-01-20
Deprecate setting the strides attribute of dpnp array

vlad-perevezentsev:
This PR deprecates setting the strides attribute of dpnp array (following NumPy 2.4)


dpnp/tests/test_ndarray.py
```diff
@@ -75,6 +75,11 @@ def test_attributes(self):
         assert_equal(self.two.nbytes, 20 * num)
         assert_equal(self.two.itemsize, self.two.dtype.itemsize)
 
+    def test_set_strides_deprecated(self):
+        x = dpnp.eye(2)
+        with pytest.warns(DeprecationWarning, match="Setting the strides"):
+            setattr(x, "strides", x.strides)
```
antonwolfy, line 81: We need some validation the strides really changed

PR #2746, opened 2026-01-26
Support tuple of integers for `axis` in `dpnp.trim_zeros`

antonwolfy:
The PR aligns with NumPy 2.4 change and adds support for tuple of integers passed with `axis` argument in `dpnp.trim_zeros`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -127,7 +127,7 @@ repos:
     hooks:
     -   id: actionlint
 -   repo: https://github.com/BlankSpruce/gersemi
-    rev: 0.25.1
+    rev: 0.25.4
     hooks:
     -   id: gersemi
         exclude: "dpnp/backend/cmake/Modules/"


dpnp/dpnp_iface_manipulation.py
```diff
@@ -4038,11 +4038,14 @@ def trim_zeros(filt, trim="fb", axis=None):
         raise ValueError(f"unexpected character(s) in `trim`: {trim!r}")
 
     nd = filt.ndim
-    if axis is not None:
-        axis = normalize_axis_index(axis, nd)
+    if axis is None:
```
vlad-perevezentsev, line 4041: Incorrect shape for empty input with Numpy 2.4
```
a = numpy.empty((0,3))
a_dp = dpnp.empty((0,3))

numpy.trim_zeros(a,axis=None).shape
# (0, 0)

dpnp.trim_zeros(a_dp, axis=None).shape
# (0, 3)

```

antonwolfy: It seems more as a bug in NumPy implementation which existed before 2.4 release as well.
An empty array does not contain any zero elements, so there is nothing to trim. IMHO it is correct to return the input array unchanged then.

vlad-perevezentsev: @antonwolfy could you add a test that shows different behavior with NumPy and a comment for it?

antonwolfy: I added a test to cover dpnp behavior with 0-sized input array.
I don't see a reason to check the NumPy behavior, if it's faulty.

PR #2747, opened 2026-01-27
Align strides with numpy

antonwolfy:
The PR changes implementation of `strides` property in `dpnp.ndarray` to align with NumPy and CuPy and to return bytes displacement in memory (previously and in dpctl it returns elements displacement).

### DIFF
--- a/.github/workflows/openssf-scorecard.yml
+++ b/.github/workflows/openssf-scorecard.yml
@@ -72,6 +72,6 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
+        uses: github/codeql-action/upload-sarif@45cbd0c69e560cd9e7cd7f8c32362050c9b7ded2 # v4.32.2
         with:
           sarif_file: results.sarif


dpnp/dpnp_iface_indexing.py
```diff
@@ -731,7 +731,7 @@ def diagonal(a, offset=0, axis1=0, axis2=1):
     elif 0 < offset < m:
         out_shape = a_shape[:-2] + (min(n, m - offset),)
         out_strides = a_straides[:-2] + (st_n + st_m,)
-        out_offset = st_m * offset
+        out_offset = st_m // a.itemsize * offset
     else:
         out_shape = a_shape[:-2] + (0,)
         out_strides = a_straides[:-2] + (1,)
```
vlad-perevezentsev: I think the logic here also needs to be updated because currently dpnp returns incorrect strides compared to NumPy. Could you also add a test to cover this case?
```
a = dpnp.arange(10).reshape(2,5)
a_np = numpy.arange(10).reshape(2,5)

dpnp.diagonal(a, offset=5).strides
# (0. )

numpy.diagonal(a_np, offset=5).strides
# (48, )
```

antonwolfy: It seems that wasn't aligned on the master as well:
```python
a = np.arange(10).reshape(2,5)
np.diagonal(a, offset=5).strides
# Out: (1,)
```
Btw I will update the code to keep the same behavior as was on the master. But I wonder if there is a strict requirement on strides equality with NumPy for a 0-size array. In any case it'd be better to handle that (if needed) in the seprarte PR.

abagusetty: @antonwolfy Thanks for bringing this distinction. Just curious to know if dpnp/dpctl recommends any particular numpy version dependency for testing. The app I was working with failed too with the `diagonal` method in stacktrace. 

Would the below legacy behavior be restored, only when the offsets is outside the bounds and rest of the functionality to changed to number of bytes from number of items ?

antonwolfy: > Just curious to know if dpnp/dpctl recommends any particular numpy version dependency for testing.

We are targeting to align with the latest published numpy version.

> The app I was working with failed too with the diagonal method in stacktrace.

Could you please share an example of the code with failure? It might be some other issue we have to resolve.

> Would the below legacy behavior be restored, only when the offsets is outside the bounds and rest of the functionality to changed to number of bytes from number of items ?

Everywhere it will be to return the number of bytes.
In case when the offsets is outside the bounds, previously dpnp returns strides end with `...,1)`, and now it will end with `..., 1 * a.itemsize)`.
The "legacy" in that use case means "different from the numpy" for returned 0-sized array.

vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy !

PR #2757, opened 2026-02-05
Add infra pytest warnings plugin

vchamarthi:
This PR introduce the plugin feature to capture and format pytest warnings to process it in internal CI.
These changes will not add any overhead to existing pytest scope. This feature can be fully enabled/disabled by env var - `DPNP_INFRA_WARNINGS_ENABLE`

### DIFF
--- a/dpnp/tests/infra_warning_utils.py
+++ b/dpnp/tests/infra_warning_utils.py
@@ -66,7 +66,7 @@ class DpnpInfraWarningsPlugin:
         except Exception:
             pass
 
-    def pytest_configure(self, _config):
+    def pytest_configure(self):
         if not self.enabled:
             return
 
@@ -177,7 +177,7 @@ class DpnpInfraWarningsPlugin:
 
         self._log_stdout(f"{self.EVENT_PREFIX} {_json_dumps_one_line(event)}")
 
-    def pytest_terminal_summary(self, terminalreporter, exitstatus, _config):
+    def pytest_terminal_summary(self, terminalreporter, exitstatus):
         if not self.enabled:
             return
 
@@ -212,7 +212,7 @@ class DpnpInfraWarningsPlugin:
         terminalreporter.write_line(_json_dumps_one_line(summary))
         terminalreporter.write_line(self.SUMMARY_END)
 
-    def pytest_unconfigure(self, _config):
+    def pytest_unconfigure(self):
         self._close_events_fp()
 
     def _close_events_fp(self):


dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,239 @@
...
+
+        try:
+            numpy_version = numpy.__version__
+            numpy_path = getattr(numpy, "__file__", "unknown")
+            dpnp_version = dpnp.__version__
+            dpnp_path = getattr(dpnp, "__file__", "unknown")
+            dpctl_version = dpctl.__version__
+            dpctl_path = getattr(dpctl, "__file__", "unknown")
+        except Exception:
+            numpy_version = "unknown"
```
antonwolfy: For what use case do we need that exception branch?

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,239 @@
...
+
+    SUMMARY_BEGIN = "DPNP_WARNINGS_SUMMARY_BEGIN"
+    SUMMARY_END = "DPNP_WARNINGS_SUMMARY_END"
+    EVENT_PREFIX = "DPNP_WARNING_EVENT "
+
+    def __init__(self):
+        self.enabled = _env_check("DPNP_INFRA_WARNINGS_ENABLE", default=False)
+        self.directory = os.getenv("DPNP_INFRA_WARNINGS_DIRECTORY", None)
+        self.events_artifact = os.getenv(
+            "DPNP_INFRA_WARNINGS_EVENTS_ARTIFACT", "dpnp_infra_warnings_events.jsonl"
```
antonwolfy: Is that a typo?
```suggestion
            "DPNP_INFRA_WARNINGS_EVENTS_ARTIFACT", "dpnp_infra_warnings_events.json"
```

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,239 @@
...
+            "origin": origin,
+            "message": message,
+            "filename": filename,
+            "lineno": lineno,
+            "function": func,
+        }
+
+        if self._events_fp is not None:
+            try:
+                self._events_fp.write(_json_dumps_one_line(event) + "\n")
```
antonwolfy, line 174: Would it be better to dump all events at once in `pytest_terminal_summary` to reduce pressure on the file system?

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+class DpnpInfraWarningsPlugin:
+    """Pytest custom plugin that records pytest-captured warnings.
+
+    It only records what pytest already captures (via pytest_warning_recorded).
+    Does not change warnings filters.
+
+    Env vars:
+    - DPNP_INFRA_WARNINGS_ENABLE=1 (enables the plugin)
```
antonwolfy, line 37: We have test relating env defined in `dpnp/tests/config.py`. Is that possible to do the same for the new ones?

dpnp/tests/infra_warning_utils.py
vchamarthi: Just for Fail safe scenario. so that this new plugin  could not break the exisiting pytest workflow

dpnp/tests/infra_warning_utils.py
vchamarthi: Not a typo, It is design choice. This plugin is event driven, `jsonl` is a line-buffered and newline-delimited json format bit different than regular json. The regular json format needs to load entire content into memory.

dpnp/tests/infra_warning_utils.py
vchamarthi: This particular line is writing the warnings to the artifact (events.jsonl) file whenever there is a warning.

The summary (summarized based of event.jsonl) will be all dumped at once to terminal and summary.json file at the end.

Summary written to terminal here - https://github.com/IntelPython/dpnp/pull/2757/changes#diff-e9924e662a511d3ac4d9aa5d26f9e9ac8865aad6a67b1dedbd9e24ce0079e590R213 

Summary written to artifact here - https://github.com/IntelPython/dpnp/pull/2757/changes#diff-e9924e662a511d3ac4d9aa5d26f9e9ac8865aad6a67b1dedbd9e24ce0079e590R201

dpnp/tests/infra_warning_utils.py
vchamarthi: I can make use of it. Good pointer.

dpnp/tests/infra_warning_utils.py
antonwolfy: I don't feel it is a real use case. If you want, you can use `getattr(numpy, "__version__", "unknown")` for the versions as well.
But I can't image the use case when and why `getattr` could fail here.

antonwolfy: IMHO, that only adds unnecessary code complexity.

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,239 @@
...
+        self.enabled = _env_check("DPNP_INFRA_WARNINGS_ENABLE", default=False)
+        self.directory = os.getenv("DPNP_INFRA_WARNINGS_DIRECTORY", None)
+        self.events_artifact = os.getenv(
+            "DPNP_INFRA_WARNINGS_EVENTS_ARTIFACT", "dpnp_infra_warnings_events.jsonl"
+        )
+        self.summary_artifact = os.getenv(
+            "DPNP_INFRA_WARNINGS_SUMMARY_ARTIFACT", "dpnp_infra_warnings_summary.json"
+        )
+
+        self.print_events = self.enabled
```
antonwolfy: What is the difference between `self.print_events` and `self.enabled`? Why do we need both of them?

dpnp/tests/infra_warning_utils.py
antonwolfy: > This particular line is writing the warnings to the artifact (events.jsonl) file whenever there is a warning.

But we can do that 2 different ways:
1. Write each warning to `events.jsonl` file once `pytest_warning_recorded` called (current approach as I can see)
2. Add each warning reported by `pytest_warning_recorded` callback to a list (or whatever collection used) and write all of them at once inside `pytest_terminal_summary` callback

vchamarthi: This is already implemented. There are two different things written to terminal
1. Actual warning{s) and 
2. The summary of all warnings.

The warnings are written to terminal in the same callback `pytest_warning_recorded`  here - https://github.com/IntelPython/dpnp/pull/2757/changes#diff-e9924e662a511d3ac4d9aa5d26f9e9ac8865aad6a67b1dedbd9e24ce0079e590R174

In the `pytest_terminal_summary` callback only the summary of the total warnings is prepared and written to terminal at the end of the tests.

dpnp/tests/infra_warning_utils.py
vchamarthi: It is redundant. I removed it.

dpnp/tests/infra_warning_utils.py
vchamarthi: I agree with you, I have captured it much clean and straighforward.

dpnp/tests/config.py
```diff
@@ -4,3 +4,15 @@
 float16_types = bool(os.getenv("DPNP_TEST_FLOAT_16", 0))
 complex_types = bool(os.getenv("DPNP_TEST_COMPLEX_TYPES", 0))
 bool_types = bool(os.getenv("DPNP_TEST_BOOL_TYPES", 0))
+
+
+infra_warnings_enable = os.getenv("DPNP_INFRA_WARNINGS_ENABLE", "0") == "1"
```
antonwolfy:
```suggestion
infra_warnings_enable = bool(os.getenv("DPNP_INFRA_WARNINGS_ENABLE", 0))
```

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,213 @@
...
+        if self.directory:
+            os.makedirs(self.directory, exist_ok=True)
+            self._events_file = os.path.join(self.directory, self.events_artifact)
+            self._events_fp = open(
+                self._events_file,
+                "w",
+                encoding="utf-8",
+                buffering=1,
+                newline="\n",
+            )
```
antonwolfy: It would be safe and clear to use `from pathlib import Path` and then:
```suggestion
            p = Path(self.directory).expanduser().resolve()
            if p.exists() and not p.is_dir():
                raise ValueError(f"{p} exists and is not a directory")

            p.mkdir(parents=True, exist_ok=True)

            if Path(self.self.events_artifact).name != self.events_artifact:
                raise ValueError(f"Invalid filename (must not contain path parts): {self._events_file}")

            self._events_file = p / self.events_artifact
            self._events_fp = self._events_file.open(mode="w", encoding="utf-8", buffering=1, newline="\n")
```

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,213 @@
...
+                "numpy_version": getattr(numpy, "__version__", "unknown"),
+                "numpy_path": getattr(numpy, "__file__", "unknown"),
+                "dpnp_version": getattr(dpnp, "__version__", "unknown"),
+                "dpnp_path": getattr(dpnp, "__file__", "unknown"),
+                "dpctl_version": getattr(dpctl, "__version__", "unknown"),
+                "dpctl_path": getattr(dpctl, "__file__", "unknown"),
+                "job": os.getenv("JOB_NAME", "unknown"),
+                "build_number": os.getenv("BUILD_NUMBER", "unknown"),
+                "git_sha": os.getenv("GIT_COMMIT", "unknown"),
+                "events_file": self._events_file,
```
antonwolfy: Is `events_file` always `None` here? Is that intended?

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,213 @@
...
+            "function": func,
+        }
+
+        if self._events_fp is not None:
+            try:
+                self._events_fp.write(_json_dumps_one_line(event) + "\n")
+            except Exception:
+                pass
+
+        #Write the warnings to terminal
```
antonwolfy:
```suggestion
        # Write the warnings to terminal
```

vchamarthi: Formatted it properly.

dpnp/tests/infra_warning_utils.py
vchamarthi: It is  not need in the env, I removed it. `events_file` will be set during the directory creation.

dpnp/tests/infra_warning_utils.py
vchamarthi: Good suggestion that works safe across platforms. I have adopted you suggestion.
Since this is a public recipe, my agenda here is to not raise error and break the workflow because of this plugin for what ever the reason. This would rather print and artifact wont be created.

dpnp/tests/config.py
vchamarthi: I will also adapt the variable type in the CI pipeline. Updated it.

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,232 @@
...
+                    )
+
+                self._events_file = p / self.events_artifact
+                self._events_fp = self._events_file.open(
+                    mode="w", encoding="utf-8", buffering=1, newline="\n"
+                )
+            except Exception as exc:
+                self._events_fp = None
+                self._events_file = None
+                self.directory = None
```
antonwolfy: It holds the env variable and it was set during the init. Why do we want to reset that here?

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,232 @@
...
+    - DPNP_INFRA_WARNINGS_EVENTS_ARTIFACT (optional filename)
+    - DPNP_INFRA_WARNINGS_SUMMARY_ARTIFACT (optional filename)
+    """
+
+    SUMMARY_BEGIN = "DPNP_WARNINGS_SUMMARY_BEGIN"
+    SUMMARY_END = "DPNP_WARNINGS_SUMMARY_END"
+    EVENT_PREFIX = "DPNP_WARNING_EVENT - "
+
+    def __init__(self):
+        self.enabled = bool(warn_config.infra_warnings_enable)
```
antonwolfy: No need to cast, already holds a boolean value:
```suggestion
        self.enabled = warn_config.infra_warnings_enable
```

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,232 @@
...
+        self._close_events_fp()
+
+    def _close_events_fp(self):
+        if self._events_fp is None:
+            return
+        try:
+            self._events_fp.close()
+        except Exception:
+            pass
+        self._events_fp = None
```
antonwolfy: since we don't have any special exception handling:
```suggestion
        finally:
            self._events_fp = None
```

dpnp/tests/infra_warning_utils.py
vchamarthi: This is the configure hook which tries to configure the artifact file,path etc. In the case of exception(for any reason) without setting these to None, the subsequent hooks ( `pytest_warning_recorded` and `pytest_terminal_summary` ) will safely skip the artifact generation part otherwise they will also fail.

dpnp/tests/infra_warning_utils.py
vchamarthi: you are right!

dpnp/tests/infra_warning_utils.py
antonwolfy: IMHO, I don't think it's a good idea to rely on `self.directory` in that case.
The variable holds a value comes from the env variable. It's not intended for handling of file opening exception.
It'd be much more better to rely of values of `self._events_fp` and `elf._events_file` variables.

Moreover, you are already doing that in `pytest_warning_recorded`, where no event will be written if `self._events_fp` is None. And there is no dependency on `self.directory`. And it'd good to keep that logic.

Also it seems, the tracing in `pytest_terminal_summary ` shouldn't depend on exception when opening event file. You writing to another dedicated summary file where. And if `self.directory` is a valid path, why shouldn't write to summary file then?

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,232 @@
...
+                p.mkdir(parents=True, exist_ok=True)
+
+                if (not self.events_artifact
+                    or Path(self.events_artifact).name != self.events_artifact
+                ):
+                    raise ValueError(
+                        "Invalid events artifact filename (must not contain path parts)"
+                    )
+
+                self._events_file = p / self.events_artifact
```
antonwolfy, line 111: I think we need to check `self.summary_artifact` here also, something like:
```python
if not (self.summary_artifact and Path(self.summary_artifact).name == self.summary_artifact):
  raise ValueError(f"Invalid summary artifact filename (must not contain path parts): {self.summary_artifact}")

self._summary_file = p / self.summary_artifact
```
and then to use validated `self._summary_file` later in `pytest_terminal_summary`

dpnp/tests/infra_warning_utils.py
```diff
@@ -0,0 +1,231 @@
...
+                if p.exists() and not p.is_dir():
+                    raise ValueError(f"{p} exists and is not a directory")
+
+                p.mkdir(parents=True, exist_ok=True)
+
+                if (not self.events_artifact
+                    or Path(self.events_artifact).name != self.events_artifact
+                ):
+                    raise ValueError(
+                        "Invalid events artifact filename (must not contain path parts)"
```
antonwolfy: I think it'd better to be more verbose:
```suggestion
                        "Invalid events artifact filename (must not contain path parts): {self.events_artifact}"
```

dpnp/tests/infra_warning_utils.py
vchamarthi: I see your point, `self.directory` is an env, shouldn't be set to none. Fixed it. Thanks for the suggestion.

dpnp/tests/infra_warning_utils.py
vchamarthi: I have added this check. to configuration.

antonwolfy: What is about adding `self._summary_file = p / self.summary_artifact` and then in `pytest_terminal_summary`?
```python
with self._summary_file.open(...) as f:
  ...
```

vchamarthi: Good catch! I have updated it.

PR #2758, opened 2026-02-06
Extend `._tensor_impl` with remaining functions used by dpnp

vlad-perevezentsev:
This PR extends `_tensor_impl` in `dpctl_ext.tensor` with the remaining functions that are explicitly used in `dpnp` implementations (`_take`, `_full_usm_ndarray`, `_zeros_usm_ndarray`, `_triu`) enabling a complete switch to `dpctl_ext.tensor._tensor_impl` instead of `dpctl.tensor._tensor_impl`

It also adds `take()`, `put()`, `full()`,`tril()` and `triu()` to `dpctl_ext.tensor` and updates the corresponding dpnp functions to use these implementations internally


dpctl_ext/tensor/libtensor/include/kernels/constructors.hpp
```diff
@@ -0,0 +1,303 @@
...
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file defines kernels for tensor constructors.
+//===----------------------------------------------------------------------===//
+
+#pragma once
+#include <array>
+#include <complex>
```
ndgrigorian, line 37:
```suggestion
```
I believe it is unused in this file, leftover from when we had to unpack the scalar manually here

dpctl_ext/tensor/libtensor/include/kernels/integer_advanced_indexing.hpp
```diff
@@ -0,0 +1,418 @@
...
+//*****************************************************************************
+//
+//===----------------------------------------------------------------------===//
+///
+/// \file
+/// This file defines kernels for advanced tensor index operations.
+//===----------------------------------------------------------------------===//
+
+#pragma once
+#include <complex>
```
ndgrigorian, line 36:
```suggestion
```

dpctl_ext/tensor/libtensor/include/kernels/constructors.hpp
ndgrigorian: actually it seems in these cases you added the header in intentionally, so I assume there was a reason

dpctl_ext/tensor/libtensor/include/kernels/integer_advanced_indexing.hpp
ndgrigorian: see above

dpctl_ext/tensor/__init__.py
```diff
@@ -25,3 +25,22 @@
...
+    take,
+)
+
+__all__ = [
+    "full",
+    "put",
+    "take",
+    "tril",
+    "triu",
+]
```
ndgrigorian, line 46: if we're adding these functions in, should we start adding dpctl test suite as well?

Or would you rather start reusing the functions in dpnp and relying on dpnp tests?

I think it's sensible to have the two test suites and merge later if possible, but it's just a choice we have to make

ndgrigorian (approved):
I don't see much else that I would change, looks like you've looked over includes. Only question from my side is the test suite

@antonwolfy if you have any thoughts you should add them but this LGTM

dpctl_ext/tensor/__init__.py
vlad-perevezentsev: It is a good point.
At the moment I am relying only on dpnp tests.
I agree that having both test suites makes sense with a possible merge later on.

I can add dpctl tests but that would require a separate PR since adding dpctl tests will put additional load on our public CI and we will likely need to increase `rerun-tests-timeout`

vlad-perevezentsev: Also there is a separate task for adding tests SAT-8344
So now I plan to continue relying on dpnp tests and once the migration is fully completed I will add tests to ensure full coverage of the migrated module

PR #2766, opened 2026-02-11
Improve `dpnp.partition` implementation

antonwolfy:
The PR propose to improve implementation and to use `dpnp.sort` call when
- input array has number of dimensions > 1
- input array has previously not supported integer dtype
- `axis` keyword is passed (previously not supported)
- sequence of `kth` is passed (previously not supported)
In case of `ndim > 1` previously the implementation from legacy backend was used, which is significantly slow (see performance comparation below). It used a copy of input data into the shared USM memory and included computations on the host.

This PR proposes to reuse `dpnp.sort` for all the above cases.
While in case when the legacy implementation is stable and fast (for 1D input array), it will remain, because it relays on `std::nth_element` from OneDPL.

The benchmark results were collected on PVC with help of the below code:
```python
import dpnp, numpy as np
from dpnp.tests.helper import generate_random_numpy_array

a = generate_random_numpy_array(10**7, dtype=np.float64, seed_value=117)
ia = dpnp.array(a)
%timeit x = dpnp.partition(ia, 513); x.sycl_queue.wait()
```

Below tables contains data in case of 1D input array (shape=(10**7,)), where the implementation path was kept the same, plus adding support of missing integer dtypes using fallback on the sort function:
| Implementation | int32 | uint32 | int64 | uint64 | float32 | float64 | complex64 | complex128 |
|--------|--------|--------|--------|--------|--------|--------|--------|--------|
| old (legacy backend) | 7.46 ms | not supported | 9.46 ms | not supported | 7.39 ms |  8.92 ms | 10.9 ms | 21.2 ms |
| new (backend + sort) | 7.34 ms | 10.8 ms | 9.48 ms | 12.5 ms | 7.37 ms | 8.89 ms | 11 ms | 21.2 ms |

The following code was used for 2D input array with shape=(10000, 10000):
```python
import dpnp, numpy as np
from dpnp.tests.helper import generate_random_numpy_array

a = generate_random_numpy_array((10**4, 10**4), dtype=np.float64, seed_value=117)
ia = dpnp.array(a)
%timeit x = dpnp.partition(ia, 1513); x.sycl_queue.wait()
```

In that case the new implementation is fully based on the sort call:
| Implementation | int32 | int64 | float32 | float64 | complex64 | complex128 |
|--------|--------|--------|--------|--------|--------|--------|
| old (legacy backend) |  6.4 s | 6.89 s | 7.36 s | 7.66 s | 8.61 s | 10 s |
| new (sort) | 57.4 ms | 64.7 ms | 62.2 ms | 68 ms | 77 ms | 151 ms |


dpnp/dpnp_iface_sorting.py
```diff
@@ -201,44 +201,128 @@ def argsort(
...
-    return call_origin(numpy.partition, x1, kth, axis, kind, order)
+    if axis is None:
+        a = dpnp.ravel(a)
+        axis = -1
+
+    nd = a.ndim
+    axis = normalize_axis_index(axis, nd)
+    length = a.shape[axis]
+
+    if isinstance(kth, int):
```
vlad-perevezentsev, line 291: It looks too strict because does not support integer-like scalars as `dpnp.int64`
In other dpnp functions this type are accepted for axes:
```
# axis : {None, int, tuple}, optional
In [22]: dpnp.count_nonzero(a, axis=dpnp.int64(0))
Out[22]: array([1, 1, 2, 1])
```

```
# axis : {None, int, tuple of ints}, optional
np.all(x, axis=np.int64(0))
Out[25]: array([ True, False])
```

dpnp/backend/kernels/dpnp_krnl_sorting.cpp
```diff
@@ -70,90 +70,25 @@ DPCTLSyclEventRef dpnp_partition_c(DPCTLSyclQueueRef q_ref,
...
+    auto policy = oneapi::dpl::execution::make_device_policy<
+        dpnp_partition_c_kernel<_DataType>>(q);
 
-        // fill the result array with data from input one
-        q.memcpy(result, arr, size * sizeof(_DataType)).wait();
+    // fill the result array with data from input one
+    q.memcpy(result, arr, size * sizeof(_DataType)).wait();
 
-        // make a partial sorting such that:
+    for (size_t i = 0; i < size_; i++) {
```
vlad-perevezentsev, line 84: What is the reason for using a loop here?
`dpnp_partition` is now only used for `ndim == 1` and the previous implementation handled this case without a loop

dpnp/dpnp_iface_sorting.py
antonwolfy: But `axes` is validating by special function `normalize_axis_tuple` from NumPy.

dpnp/backend/kernels/dpnp_krnl_sorting.cpp
antonwolfy: This is how it can be implemented through OneDPL for ndim > 1.
I'd prefer to keep that in case it will be needed in case of future rework.

> `dpnp_partition` is now only used for `ndim == 1`

It isn't limited from backend implementation, it's only how we are currently using that.

vlad-perevezentsev: It seems that the current implementation will not work correctly for `ndim>1`
Could you please add a comment that this kernel currently only supports `ndim == 1`

antonwolfy: Could you please share an example where it doesn't work?

dpnp/dpnp_iface_sorting.py
vlad-perevezentsev: I meant `kth` argument 
```
dpnp.partition(a,dpnp.int64(4),axis=None)
TypeError: kth must be int or sequence of ints, but got <class 'numpy.int64'>
```
Should we use `operator.index` for this as implemented in `dpnp.tri` for `k`?

dpnp/backend/kernels/dpnp_krnl_sorting.cpp
vlad-perevezentsev:
```
import numpy
import dpnp

a = numpy.array([[3,1,2],[6,4,5]])
a_dp = dpnp.array(a)

numpy.partition(a, 1, axis=1)
# array([[1, 2, 3],
#           [4, 5, 6]])

dpnp.partition(a_dp, 1, axis=1)
use dpnp_partition kernel
# array([[1, 2, 0],
#            [0, 3, 4]])
```

dpnp/dpnp_iface_sorting.py
antonwolfy: We could, but it's not mandatory, because we explicitly state the support of int and sequence of itns in the documentation.
We can implement a general improvement and to add a helper function, like `normalize_int_sequence`. But I'd prefer to have that done separately as a general improvement.

dpnp/backend/kernels/dpnp_krnl_sorting.cpp
antonwolfy: `axis` keyword isn't supported by `dpnp_partition`, it assumes `axis=-1` (default value) always

vlad-perevezentsev:
```
In [3]: a = numpy.array([[3,1,2],[6,4,5]])

In [4]: a_dp = dpnp.array(a)

In [5]: numpy.partition(a, 1, axis=-1)
Out[5]: 
array([[1, 2, 3],
       [4, 5, 6]])

In [6]: dpnp.partition(a_dp, 1, axis=-1)
use dpnp_partition kernel
Out[6]: 
array([[1, 2, 0],
       [0, 3, 4]])
```

antonwolfy: I can't reproduce that. For me the example works:
```python
import dpnp as np, numpy
from dpnp.dpnp_algo import dpnp_partition

a = numpy.array([[3,1,2],[6,4,5]])
ia = np.array(a)

numpy.partition(a, 1, axis=-1)
# Out:
# array([[1, 2, 3],
#        [4, 5, 6]])

desc = np.get_dpnp_descriptor(a, copy_when_nondefault_queue=False)
dpnp_partition(desc, -1, "introselect", None).get_pyobj()
# Out:
# array([[1, 2, 3],
#        [4, 5, 6]])
```

antonwolfy: Btw, I'll add a comment that ndim>1 is experimental and wasn't tested properly.

PR #2768, opened 2026-02-12
fix: change MKL required version constraints

jharlow-intel:
Straight forward PR


jharlow-intel:
Ugh, fails on the stable builds due to the new constraint

PR #2770, opened 2026-02-16
Suppress the compiler warnings marking them as external

antonwolfy:
On Linux, the DPC++ compiler headers are automatically treated as system headers. But it is not the case on Windows.
Due to that, there is a ton of deprecation warnings generated inside the compiler headers when building dpnp extensions on Windows.

This PR updates CMake files of the extensions to add include of DPC++ compiler headers explicitly and to mark the compiler and dpctl headers as system one to suppress the warning inside them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ Also, that release drops support for Python 3.9, making Python 3.10 the minimum 
 * Aligned `strides` property of `dpnp.ndarray` with NumPy and CuPy implementations [#2747](https://github.com/IntelPython/dpnp/pull/2747)
 * Extended `dpnp.nan_to_num` to support broadcasting of `nan`, `posinf`, and `neginf` keywords [#2754](https://github.com/IntelPython/dpnp/pull/2754)
 * Changed `dpnp.partition` implementation to reuse `dpnp.sort` where it brings the performance benefit [#2766](https://github.com/IntelPython/dpnp/pull/2766)
+* `dpnp` uses pybind11 3.0.2 [#27734](https://github.com/IntelPython/dpnp/pull/2773)
 * Modified CMake files for the extension to explicitly mark DPC++ compiler and dpctl headers as system ones and so to suppress the build warning generated inside them [#2770](https://github.com/IntelPython/dpnp/pull/2770)
 
 ### Deprecated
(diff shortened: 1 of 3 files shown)


dpnp/backend/extensions/ufunc/CMakeLists.txt
```diff
@@ -84,16 +84,16 @@ set_target_properties(
 
 target_include_directories(
     ${python_module_name}
-    PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../../
+    PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../../ ${CMAKE_CURRENT_SOURCE_DIR}/../common
```
vlad-perevezentsev, line 87:
```suggestion
    PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../common
```

dpnp/backend/extensions/window/CMakeLists.txt
```diff
@@ -62,16 +62,16 @@ set_target_properties(
 
 target_include_directories(
     ${python_module_name}
-    PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../../
+    PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../../ ${CMAKE_CURRENT_SOURCE_DIR}/../common
```
vlad-perevezentsev, line 65:
```suggestion
    PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../common
```

dpnp/backend/extensions/ufunc/CMakeLists.txt
antonwolfy: It's automatically formatting by the pre-commit

dpnp/backend/extensions/window/CMakeLists.txt
antonwolfy: The same, pre-commit will fail then

dpnp/backend/extensions/ufunc/CMakeLists.txt
antonwolfy: Misread the comment. Btw that will not work, the compilation failed due to missing kernel include.

dpnp/backend/extensions/window/CMakeLists.txt
antonwolfy: And this one the same, the compilation will fail then.

dpnp/backend/extensions/ufunc/CMakeLists.txt
vlad-perevezentsev: Ok, then this comment is not relevant

PR #2797, opened 2026-03-04
Clean up dpctl.tensor code

antonwolfy:
This PR adds a small clean up to already porting dpctl.tensor code:
* remove unused includes
* add missing includes
* remove redundant namespace qualifications when calling function from the same namespace


ndgrigorian (dismissed):
LGTM, I see no problems in this

with the simplify_iteration_space variants, many of the files use `using dpctl::tensor::py_internal::...` so we would have to remove those lines too, but we can do it in a follow up

PR #2817, opened 2026-03-12
Move tensor tests from dpctl 

vlad-perevezentsev:
This PR moves all tensor-related tests to `dpnp/tests/tensor` as part of the ongoing migration of tensor functionality from `dpctl`  to `dpnp.tensor`

Key changes:

> - Relocated 89 tensor tests (elementwise functions, `usm_ndarray`, and tensor utilities)
> - Updated imports to use `dpnp.tensor`
> - Included tests in packaging configuration
> - Integrated tensor tests into CI
> - Fixed several issues discovered during migration (dtype expectations, boolean reductions, etc.)
> - Fixed a circular import in _usmarray.py
> - Added `SKIP_TENSOR_TESTS` env variable to manage the launch of the test scope

In a follow-up PR:

> - Conditional logic will be added to run dpctl_ext/tests only when changes affect the tensor code.
> - Array API tests for tensor will be introduced and executed as a separate CI job.


ndgrigorian (dismissed):
lgtm

dpnp/tensor/libtensor/include/kernels/elementwise_functions/round.hpp
```diff
@@ -116,6 +116,7 @@ template <typename T>
 struct RoundOutputType
 {
     using value_type = typename std::disjunction<
+        td_ns::TypeMapResultEntry<T, bool, sycl::half>,
```
antonwolfy, line 119: I believe the idea was to have only kernels for numeric data types per Python array API [spec](https://data-apis.org/array-api/latest/API_specification/generated/array_api.round.html#round).
What was the reason to extend?

.github/workflows/conda-package.yml
```diff
@@ -242,11 +242,31 @@ jobs:
 
             if [[ "${{ matrix.python }}" == "${{ env.python-ver-test-all-dtypes }}" ]]; then
               export DPNP_TEST_ALL_INT_TYPES=1
-              python -m pytest -ra --pyargs ${{ env.package-name }}.tests
+              python -m pytest -ra --ignore=dpnp/tests/tensor --pyargs ${{ env.package-name }}.tests
```
antonwolfy: What is the idea to decouple the running of tensor tests?

dpnp/tensor/libtensor/include/kernels/elementwise_functions/round.hpp
vlad-perevezentsev: Otherwise `test_round_order` would fail with an error
```
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=0
E       
E       (dtypes int8, float16 mismatch)
```
dpctl tests did not include type checks so the error was not detected until now

I can also change the dtypes being tested in the test if that is  a better solution 
@ndgrigorian  what do you think?

.github/workflows/conda-package.yml
vlad-perevezentsev: 1. to reduce memory usage in a single step
2. In #2823  conditional running of tensor tests was added and test separation was implemented here as a preparatory step for future changes

dpnp/tensor/libtensor/include/kernels/elementwise_functions/round.hpp
ndgrigorian: If I am understanding correctly, without adding the bool to half loop, the output type is int8?

I think it's not so intrusive to add a bool to half loop to align with numpy, but perfect dtype alignment was never the goal for `tensor`. But if it's useful to dpnp (avoids some work-around logic) I don't see an issue.

antonwolfy: As I remember there is no w/a logic in dpnp. I guess we didn't check the result type exactly match NumPy in case of bool input.

The reason of change here was failing dpctl test `test_round_order` after migration.
But I have no objections to extend with extra kernel for bool dtype.

antonwolfy (dismissed):
In general, LGTM

dpnp/tensor/libtensor/include/kernels/elementwise_functions/round.hpp
ndgrigorian: yes, it's a good point -- I'm not sure why this test would suddenly start failing, that's a very surprising regression. It is passing fine in dpctl as-is.

vlad-perevezentsev: @ndgrigorian this issue exists in dpctl as well
It is just that in dpnp `numpy.testing.assert_allclose` is rebound to `testing.assert_allclose` which performs an additional data type check

Here is a run of `test_round_order` with type prints (dpctl 0.22.0dev1)
```
test_round.py::test_round_order[b1] 
order:  C
dpctl_res_dtype:  int8
numpy_res_dtype:  float16
order:  F
dpctl_res_dtype:  int8
numpy_res_dtype:  float16
order:  A
dpctl_res_dtype:  int8
numpy_res_dtype:  float16
order:  K
dpctl_res_dtype:  int8
numpy_res_dtype:  float16
PASSED
test_round.py::test_round_order[i1] 
order:  C
dpctl_res_dtype:  int8
numpy_res_dtype:  int8
order:  F
dpctl_res_dtype:  int8
numpy_res_dtype:  int8
order:  A
dpctl_res_dtype:  int8
numpy_res_dtype:  int8
order:  K
```

ndgrigorian: that explains it then

dpnp/tests/test_ndarray.py
```diff
@@ -490,6 +490,9 @@ def test_print_dpnp_special_character(character):
     assert result == expected
 
 
+# TODO: repr formatting is inconsistent (scientific vs integer-like output)
+# This is a minor issue that does not depend on compiler flags
+@pytest.mark.skip(reason="Flaky: repr formatting is not stable")
```
antonwolfy: file a JIRA ticket for that and post link here

vlad-perevezentsev: done

antonwolfy (approved):
Thank you @vlad-perevezentsev, no more comments from me

PR #2823, opened 2026-03-19
Enable conditional execution of Array API tests for dpctl_ext in GH Action

vlad-perevezentsev:
This PR proposes introducing conditional execution of Python Array API  tests in CI based on whether files under the `dpctl_ext` directory are modified.

It also enables conditional execution of dpctl_ext tests using [Paths Changes Filter](https://github.com/marketplace/actions/paths-changes-filter) action, adds a dedicated skip list for tensor Array API tests, reports results in PR comments


antonwolfy (commented):
I think we can park that PR currently, and to implement it later when the migration is stabilized.
As for now, I believe it might be helpful to run all tests scope even if no changes in the tensor. It'd help to monitor if any flaky issue happens.

PR #2836, opened 2026-04-03
Add `__main__.py` for CLI options to get `libtensor` include dirs from module

ndgrigorian:
This PR proposes module CLI options that allow users to get a path to the `libtensor` include dir or to get include flags for `libtensor`

This streamlines workflow for users building native extensions

### DIFF
--- a/dpnp/__main__.py
+++ b/dpnp/__main__.py
@@ -1,0 +1,78 @@
+# *****************************************************************************
+# Copyright (c) 2016, Intel Corporation
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+# - Redistributions of source code must retain the above copyright notice,
+#   this list of conditions and the following disclaimer.
+# - Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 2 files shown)


dpnp/__main__.py
```diff
@@ -0,0 +1,78 @@
...
+def print_tensor_include_flags() -> None:
+    """Prints include flags for dpnp tensor library"""
+    libtensor_dir = get_tensor_include_dir()
+    print("-I " + libtensor_dir)
+
+
+def main() -> None:
+    """Main entry-point."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
```
antonwolfy, line 58: Can we add tests to validate the options?

ndgrigorian: we can, though I would want https://github.com/IntelPython/dpnp/pull/2817 merged first so that the tests can be added with the other tensor tests

We can more or less copy the tests from dpctl

vlad-perevezentsev: @ndgrigorian #2817 have been merged

antonwolfy (approved):
Thank you @ndgrigorian, LGTM

PR #2841, opened 2026-04-09
Feature sparse linalg solvers

abagusetty:
Adds support for `from dpnp.scipy.sparse.linalg import LinearOperator, cg, gmres, minres` 
Fixes: https://github.com/IntelPython/dpnp/issues/2831


dpnp/backend/extensions/sparse/gemv.cpp
```diff
@@ -0,0 +1,416 @@
...
+#include "utils/output_validation.hpp"
+#include "utils/type_utils.hpp"
+
+#include "gemv.hpp"
+#include "types_matrix.hpp"
+
+namespace dpnp::extensions::sparse
+{
+
+namespace mkl_sparse = oneapi::mkl::sparse;
```
antonwolfy, line 57: I believe we have to add a run dependency on `onemkl-sycl-sparse` to `conda-recipe/meta.yaml`

abagusetty:
@antonwolfy Apologies for the delay. I got side tracked for a bit. The current workflow is failing with a `sycl-post-link` issue with `2026.0.0` compiler version during the link of `libdpnp_backend_c.so` and I see an other workflows are also affected by this. 

Is this a known limitation and being triaged by compiler team

abagusetty:
Seems like it is: https://github.com/IntelPython/dpnp/pull/2887

antonwolfy:
@abagusetty, yep, it's expected for the coverage due to known compiler bug, we will implement the w/a in #2887
Other workflows should pass now.

dpnp/backend/extensions/sparse/CMakeLists.txt
```diff
@@ -0,0 +1,119 @@
+# -*- coding: utf-8 -*-
+# *****************************************************************************
+# Copyright (c) 2025, Intel Corporation
```
antonwolfy: need to bump copyright year for new added file, should be the current year
```suggestion
# Copyright (c) 2026, Intel Corporation
```

dpnp/backend/extensions/sparse/CMakeLists.txt
```diff
@@ -0,0 +1,119 @@
...
+
+set(python_module_name _sparse_impl)
+set(_module_src
+    ${CMAKE_CURRENT_SOURCE_DIR}/sparse_py.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/gemv.cpp
+)
+
+pybind11_add_module(${python_module_name} MODULE ${_module_src})
+add_sycl_to_target(TARGET ${python_module_name} SOURCES ${_module_src})
+
```
antonwolfy, line 38:
```suggestion

# Ensure Cython modules build first so _usmarray.h exists
add_dependencies(${python_module_name} _usmarray)

```

dpnp/backend/extensions/sparse/gemv.cpp
```diff
@@ -0,0 +1,400 @@
...
+#include <tuple>
+#include <utility>
+#include <vector>
+
+#include <pybind11/pybind11.h>
+
+// utils extension header
+#include "ext/common.hpp"
+
+// dpctl tensor headers
```
antonwolfy:
```suggestion
// dpnp tensor headers
```

dpnp/backend/extensions/sparse/gemv.cpp
```diff
@@ -0,0 +1,400 @@
...
+ */
+typedef std::pair<std::uintptr_t, sycl::event> (*gemv_init_fn_ptr_t)(
+    sycl::queue &,
+    oneapi::mkl::transpose,
+    const char *, // row_ptr (typeless)
+    const char *, // col_ind (typeless)
+    const char *, // values  (typeless)
+    std::int64_t, // num_rows
+    std::int64_t, // num_cols
+    std::int64_t, // nnz
```
antonwolfy:
```suggestion
    const std::int64_t, // num_rows
    const std::int64_t, // num_cols
    const std::int64_t, // nnz
```

dpnp/backend/extensions/sparse/types_matrix.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+#include <complex>
+#include <cstdint>
+#include <type_traits>
+
+// dpctl tensor headers
+#include "utils/type_dispatch.hpp"
+
+// dpctl namespace alias for type dispatch utilities
+namespace dpctl_td_ns = dpnp::tensor::type_dispatch;
```
antonwolfy:
```suggestion
// namespace for operations with types
namespace dpnp_td_ns = dpnp::tensor::type_dispatch;
```

dpnp/backend/extensions/sparse/types_matrix.hpp
```diff
@@ -0,0 +1,122 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+//*****************************************************************************
+
+#pragma once
+
+#include <complex>
+#include <cstdint>
+#include <type_traits>
+
+// dpctl tensor headers
```
antonwolfy:
```suggestion
// dpnp tensor headers
```

dpnp/backend/extensions/sparse/gemv.cpp
```diff
@@ -0,0 +1,400 @@
...
+ * compute_impl: fires sparse::gemv using a pre-built handle.
+ * Returns the gemv event directly -- no host_task wrapping.
+ */
+typedef sycl::event (*gemv_compute_fn_ptr_t)(
+    sycl::queue &,
+    oneapi::mkl::sparse::matrix_handle_t,
+    oneapi::mkl::transpose,
+    double,       // alpha (cast to Tv inside)
+    const char *, // x (typeless)
+    double,       // beta  (cast to Tv inside)
```
antonwolfy:
```suggestion
    const double,       // alpha (cast to Tv inside)
    const char *, // x (typeless)
    const double,       // beta  (cast to Tv inside)
```

CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ Also, that release drops support for Python 3.9, making Python 3.10 the minimum
 * Added implementation of `dpnp.scipy.linalg.lu` (SciPy-compatible) [#2787](https://github.com/IntelPython/dpnp/pull/2787)
 * Added support for ndarray subclassing via `dpnp.ndarray.view` method with `type` parameter [#2815](https://github.com/IntelPython/dpnp/pull/2815)
 * Migrated tensor implementation from `dpctl.tensor` into `dpnp.tensor`, making `dpnp` the primary owner of the Array API-compliant tensor layer [#2856](https://github.com/IntelPython/dpnp/pull/2856)
+* Added implementation of `dpnp.scipy.sparse.linalg import LinearOperator, cg, gmres, minres` [#2841](https://github.com/IntelPython/dpnp/pull/2841)
```
antonwolfy: need to move to 0.21.0 release

dpnp/scipy/sparse/linalg/_interface.py
```diff
@@ -0,0 +1,593 @@
...
+      4. Duck-typed objects with .shape and .matvec / @ support.
+    """
+    if isinstance(A, LinearOperator):
+        return A
+
+    try:
+        # Lazy import: dpnp.scipy.sparse may import this module during
+        # package initialisation, so we cannot import it at module scope.
+        # pylint: disable=import-outside-toplevel
+        from dpnp.scipy.sparse import issparse
```
antonwolfy: It seems `issparse` is missed and not implemented currently

dpnp/tests/test_scipy_sparse_linalg.py
```diff
@@ -0,0 +1,896 @@
...
+from dpnp.tests.third_party.cupy import testing
+
+if is_scipy_available():
+    import scipy.sparse.linalg as scipy_sla
+
+
+# Helpers for constructing SPD, diagonally dominant, and symmetric
+# indefinite test matrices. Kept small and local, matching the style of
+# vvsort() at the top of test_linalg.py.
+def _spd_matrix(n, dtype):
```
antonwolfy: Can we add CuPy tests to `dpnp/tests/third_party/cupyx/scipy_tests/sparse_tests/test_linalg.py` to check the compatibly?

dpnp/scipy/sparse/linalg/_interface.py
abagusetty: Thanks for taking a peek, I cleaned it up

dpnp/tests/test_scipy_sparse_linalg.py
abagusetty: Done

dpnp/backend/extensions/sparse/gemv.cpp
```diff
@@ -0,0 +1,399 @@
+//*****************************************************************************
+// Copyright (c) 2025, Intel Corporation
```
vlad-perevezentsev: Please update the year to `2026` in the new files

dpnp/backend/extensions/sparse/types_matrix.hpp
```diff
@@ -0,0 +1,122 @@
...
+ * below -- both are equivalent.
+ *
+ * @tparam Tv  Value type of the sparse matrix and dense vectors.
+ */
+template <typename Tv>
+struct SparseGemvComputeTypeSupportFactory
+{
+#if defined(DPCTL_HAS_TYPE_DEFINED_ENTRY)
+    static constexpr bool
+        is_defined = std::disjunction dpnp_td_ns::TypeDefinedEntry<Tv, float>,
```
vlad-perevezentsev: Angle brackets are missing for `std::disjunction<>`

dpnp/tests/test_scipy_sparse_linalg.py
```diff
@@ -0,0 +1,898 @@
...
+        return 1e-3
+    return 1e-5
+
+
+# GMRES in dpnp.scipy.sparse.linalg._iterative uses real-valued Givens
+# rotation formulas which are incorrect for complex Arnoldi, so GMRES
+# returns wrong solutions for complex dtypes. Complex GMRES tests are
+# xfailed below. When the Givens block is fixed the xfails will flip to
+# XPASS and force an update here.
+_GMRES_CPX_XFAIL = (
```
vlad-perevezentsev: As I understand it commit [fb11515](https://github.com/IntelPython/dpnp/pull/2841/changes/fb11515e7dd6fbf47864ab88fd866d3133ccddc5) fixes the complex GMRES convergence issue so `_GMRES_CPX_XFAIL` is unnecessary
At least without that and changed  `_GMRES_DTYPES` all tests pass on my laptop

dpnp/backend/extensions/sparse/gemv.cpp
abagusetty: Resolved in the recent commits

dpnp/backend/extensions/sparse/types_matrix.hpp
abagusetty: Thanks for taking a peek, cleaned-up this file

dpnp/tests/test_scipy_sparse_linalg.py
abagusetty: Addressed by removing the `_GMRES_CPX_XFAIL` and improving stability tests

abagusetty:
@antonwolfy @vlad-perevezentsev Could you please confirm the failing git actions: https://github.com/IntelPython/dpnp/actions/runs/27830969946/job/82761533818#step:13:2975 are from an in-progress work upstream and was wondering if the PR is GTO

```bash
  =================================== FAILURES ===================================
  _ TestNdarrayTakeErrorTypeMismatch_param_0_{indices=(2, 3), out_shape=(2, 3), shape=(3, 4, 5)}.test_output_type_mismatch _
  [gw2] linux -- Python 3.13.14 /home/runner/miniconda3/envs/test/bin/python
  
  self = <<cupy.core_tests.test_ndarray.TestNdarrayTakeErrorTypeMismatch_param_0_{indices=(2, 3), out_shape=(2, 3), shape=(3, 4, 5)} testMethod=test_output_type_mismatch>  parameter: {'shape': (3, 4, 5), 'indices': (2, 3), 'out_shape': (2, 3)}>
  
      def test_output_type_mismatch(self):
          for xp in (numpy, cupy):
              a = testing.shaped_arange(self.shape, xp, numpy.int32)
              i = testing.shaped_arange(self.indices, xp, numpy.int32) % 3
              o = testing.shaped_arange(self.out_shape, xp, numpy.float32)
  >           with pytest.raises(TypeError):
                   ^^^^^^^^^^^^^^^^^^^^^^^^
  E           Failed: DID NOT RAISE TypeError
  
  ../../../miniconda3/envs/test/lib/python3.13/site-packages/dpnp/tests/third_party/cupy/core_tests/test_ndarray.py:572: Failed
  _ TestNdarrayTakeErrorTypeMismatch_param_1_{indices=(), out_shape=(), shape=()}.test_output_type_mismatch _
  [gw2] linux -- Python 3.13.14 /home/runner/miniconda3/envs/test/bin/python
  
  self = <<cupy.core_tests.test_ndarray.TestNdarrayTakeErrorTypeMismatch_param_1_{indices=(), out_shape=(), shape=()} testMethod=test_output_type_mismatch>  parameter: {'shape': (), 'indices': (), 'out_shape': ()}>
  
      def test_output_type_mismatch(self):
          for xp in (numpy, cupy):
              a = testing.shaped_arange(self.shape, xp, numpy.int32)
              i = testing.shaped_arange(self.indices, xp, numpy.int32) % 3
              o = testing.shaped_arange(self.out_shape, xp, numpy.float32)
  >           with pytest.raises(TypeError):
                   ^^^^^^^^^^^^^^^^^^^^^^^^
  E           Failed: DID NOT RAISE TypeError
  
  ../../../miniconda3/envs/test/lib/python3.13/site-packages/dpnp/tests/third_party/cupy/core_tests/test_ndarray.py:572: Failed
```

antonwolfy:
@abagusetty, it should work now. The fail was caused by new NumPy 2.5 release, but already resolved in #2959

dpnp/scipy/sparse/linalg/_iterative.py
```diff
@@ -0,0 +1,1189 @@
...
+    y = psolve(r1)
+
+    # beta1 = <r1, y>   -- one host sync (setup only).
+    # Transferred to host immediately because beta1 seeds ~5 host-side
+    # scalars (beta, qrnorm, phibar, rhs1) used in Python arithmetic
+    # and branches every iteration.  Keeping it as a 0-D device array
+    # would cascade implicit syncs or 0-D allocations throughout the
+    # recurrence -- and the < 0 / == 0 guards below would each trigger
+    # an implicit __bool__ sync of their own.
+    beta1 = float(dpnp.inner(r1, y))
```
antonwolfy: The below example is failing here:

```python
import dpnp
from dpnp.scipy.sparse.linalg import minres

# 2x2 Hermitian, positive definite: [[2, 1-1j], [1+1j, 3]]
A = dpnp.array([[2.0 + 0j, 1.0 - 1j],
                [1.0 + 1j, 3.0 + 0j]], dtype=dpnp.complex128)
b = dpnp.array([1.0 + 0j, 2.0 + 0j], dtype=dpnp.complex128)

x, info = minres(A, b)
Cell In[7], line 9
      5 A = dpnp.array([[2.0 + 0j, 1.0 - 1j],
      6                 [1.0 + 1j, 3.0 + 0j]], dtype=dpnp.complex128)
      7 b = dpnp.array([1.0 + 0j, 2.0 + 0j], dtype=dpnp.complex128)
----> 9 x, info = minres(A, b)

File /localdisk/work/antonvol/code/dpnp/dpnp/scipy/sparse/linalg/_iterative.py:871, in minres(A, b, x0, rtol, shift, maxiter, M, callback, show, check)
    862 y = psolve(r1)
    864 # beta1 = <r1, y>   -- one host sync (setup only).
    865 # Transferred to host immediately because beta1 seeds ~5 host-side
    866 # scalars (beta, qrnorm, phibar, rhs1) used in Python arithmetic
   (...)    869 # recurrence -- and the < 0 / == 0 guards below would each trigger
    870 # an implicit __bool__ sync of their own.
--> 871 beta1 = float(dpnp.inner(r1, y))
    873 if beta1 < 0:
    874     raise ValueError("indefinite preconditioner")

File /localdisk/work/antonvol/code/dpnp/dpnp/dpnp_array.py:348, in dpnp_array.__float__(self)
    346 def __float__(self, /):
    347     """Convert a zero-dimensional array to a Python float object."""
--> 348     return self._array_obj.__float__()

File /localdisk/work/antonvol/code/dpnp/dpnp/tensor/_usmarray.pyx:1227, in dpnp.tensor._usmarray.usm_ndarray.__float__()
   1225     _check_0d_scalar_conversion(self)
   1226     view = _as_zero_dim_ndarray(self)
-> 1227     return view.__float__()
   1228
   1229 raise ValueError(

TypeError: float() argument must be a string or a real number, not 'complex'
```

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,452 @@
...
+                f"csr_matrix: indices dtype must be int32 or int64, "
+                f"got {indices.dtype}"
+            )
+        if _np.dtype(indptr.dtype).char != idx_char:
+            raise TypeError(
+                f"csr_matrix: indptr dtype ({indptr.dtype}) must match "
+                f"indices dtype ({indices.dtype})"
+            )
+
+        if dtype is not None and _np.dtype(dtype) != _np.dtype(data.dtype):
```
antonwolfy, line 273: indices/indptr might stay as shared references to caller arrays:
```python
import dpnp
from dpnp.scipy.sparse import csr_matrix

indices = dpnp.array([0, 1, 2, 3], dtype=dpnp.int64)
M = csr_matrix((dpnp.ones(4, dtype=dpnp.float32), indices, dpnp.arange(5, dtype=dpnp.int64)),
    shape=(4, 4), dtype=dpnp.float64, copy=True)

indices[0] = 999                    # mutate caller's source array
print(int(M.indices[0]))            # expected 0; prints 999
# Out: 999
```

dpnp/scipy/sparse/linalg/_iterative.py
```diff
@@ -0,0 +1,1189 @@
...
+            # H[j+1, j] = ||u||  -- one device norm, one slice store.
+            # Stored as a device 0-D scalar; we only sync if we need
+            # to read its value for the next v normalisation.
+            h_norm = dpnp.linalg.norm(u)
+            H[j + 1, j] = h_norm
+            if j < last_j:
+                # Normalise u into the next Krylov vector and store it
+                # in V. The single in-place store V[:, j+1] = v writes
+                # a contiguous column slice with a unit-stride layout.
+                v = u / h_norm
```
antonwolfy: `h_norm` might be a zero:
```python
import dpnp
from dpnp.scipy.sparse.linalg import gmres
 
 A = dpnp.diag(dpnp.array([2.0, 2.0, 2.0, 2.0]))  # exact soln reachable in 1 Krylov step
 b = dpnp.ones(4, dtype=dpnp.float64)
 
 x, info = gmres(A, b, restart=4, maxiter=20)      # expected x = [0.5, 0.5, 0.5, 0.5]
 print(info, x)
 ** On entry to DLASCL parameter number  4 had an illegal value
 ** On entry to DLASCL parameter number  4 had an illegal value
LinAlgError                               Traceback (most recent call last)
Cell In[4], line 7
      4 A = dpnp.diag(dpnp.array([2.0, 2.0, 2.0, 2.0]))  # exact soln reachable in 1 Krylov step
      5 b = dpnp.ones(4, dtype=dpnp.float64)
----> 7 x, info = gmres(A, b, restart=4, maxiter=20)      # expected x = [0.5, 0.5, 0.5, 0.5]
      8 print(info, x)

File /localdisk/work/antonvol/code/dpnp/dpnp/scipy/sparse/linalg/_iterative.py:761, in gmres(A, b, x0, rtol, atol, restart, maxiter, M, callback, callback_type)
    753 # Solve the small Hessenberg least-squares  H y = e  on the
    754 # host. The matrix is (restart+1) x restart -- typically
    755 # 21 x 20 -- so the device SVD launch overhead dominates;
   (...)    758 # restart device-side lstsq that allocated a workspace and
    759 # ran a tiny SVD kernel.
    760 H_host = dpnp.asnumpy(H)
--> 761 y_host, *_ = numpy.linalg.lstsq(H_host, e_host, rcond=None)
    762 y = dpnp.asarray(y_host, sycl_queue=queue)
    763 x = x + dpnp.dot(V, y)

File /localdisk/work/antonvol/miniforge3/envs/dpnp_dev/lib/python3.13/site-packages/numpy/linalg/_linalg.py:2539, in lstsq(a, b, rcond)
   2535     b = zeros(b.shape[:-2] + (m, n_rhs + 1), dtype=b.dtype)
   2537 with errstate(call=_raise_linalgerror_lstsq, invalid='call',
   2538               over='ignore', divide='ignore', under='ignore'):
-> 2539     x, resids, rank, s = _umath_linalg.lstsq(a, b, rcond,
   2540                                              signature=signature)
   2541 if m == 0:
   2542     x[...] = 0

File /localdisk/work/antonvol/miniforge3/envs/dpnp_dev/lib/python3.13/site-packages/numpy/linalg/_linalg.py:156, in _raise_linalgerror_lstsq(err, flag)
    155 def _raise_linalgerror_lstsq(err, flag):
--> 156     raise LinAlgError("SVD did not converge in Linear Least Squares")

LinAlgError: SVD did not converge in Linear Least Squares

```

dpnp/backend/extensions/blas/blas_py.cpp
```diff
@@ -151,6 +151,32 @@ PYBIND11_MODULE(_blas_impl, m)
...
+        // trans_op selects the op applied to A:
+        //   0 = N   y = alpha * A   @ x + beta * y
+        //   1 = T   y = alpha * A^T @ x + beta * y
+        //   2 = C   y = alpha * A^H @ x + beta * y   (F-contig only)
+        //
+        // For complex matrices the scalars must be exactly
+        // representable as their real form: callers pass {-1, 0, 1};
+        // fractional or complex scalars would lose the imaginary
+        // component on the C++ cast.
+        m.def("_gemv_alpha_beta", &blas_ns::gemv_alpha_beta,
```
antonwolfy: Why do we need a separate new wrapper over `gemv`?
We can declare default values to `alpha` and `beta` in `"_gemv"` definition and it would exactly the same, isn't it?
The new wrapper doesn't seem needed in terms of any performance gain... and there is no cost to extend `_gemv` with two more keywords arguments.

dpnp/backend/extensions/blas/blas_py.cpp
```diff
@@ -151,6 +151,32 @@ PYBIND11_MODULE(_blas_impl, m)
               py::arg("depends") = py::list());
     }
 
+    {
+        // y = alpha * op(A) * x + beta * y, the full BLAS gemv form.
+        // Used by dpnp.scipy.sparse.linalg.gmres to fuse the Arnoldi
```
antonwolfy: I don't think we need such kind of comment where we describing where we are going to use new `_gemv`. It doesn't look much informative and more like left by AI.

dpnp/backend/extensions/sparse/gemv.cpp
```diff
@@ -0,0 +1,638 @@
...
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+//*****************************************************************************
+
+#include <complex>
```
antonwolfy: Do we need that include or redundant?

dpnp/scipy/sparse/_csr.py
abagusetty: Thanks. Fixed this inside `_csr.py`
```
import dpnp
from dpnp.scipy.sparse import csr_matrix
indices = dpnp.array([0, 1, 2, 3], dtype=dpnp.int64)
M = csr_matrix((dpnp.ones(4, dtype=dpnp.float32), indices, dpnp.arange(5, dtype=dpnp.int64)),
               shape=(4, 4), dtype=dpnp.float64, copy=True)
indices[0] = 999                    # mutate caller's source array
print(int(M.indices[0]))            # expected 0; prints 999
```
Output:
```
$ python3 test_dpnp_sparse_csr_matrix.py
0
```

dpnp/scipy/sparse/linalg/_iterative.py
abagusetty: Fixed. 
```
import dpnp
from dpnp.scipy.sparse.linalg import minres
# 2x2 Hermitian, positive definite: [[2, 1-1j], [1+1j, 3]]
A = dpnp.array([[2.0 + 0j, 1.0 - 1j],
                [1.0 + 1j, 3.0 + 0j]], dtype=dpnp.complex128)
b = dpnp.array([1.0 + 0j, 2.0 + 0j], dtype=dpnp.complex128)
x, info = minres(A, b)
print('x, info: ', x, info)
```
output:
```
$ python3 test_dpnp_sparse_minres.py
x, info:  [0.25+0.5j  0.75-0.25j] 0
```

dpnp/backend/extensions/sparse/gemv.cpp
abagusetty: This header wasnt needed, it is removed

dpnp/backend/extensions/blas/blas_py.cpp
abagusetty: Cleanup the comments!

dpnp/scipy/sparse/linalg/_iterative.py
abagusetty: Fixed: 
```
$ python3 test_dpnp_sparse_gmres.py
0 [0.5 0.5 0.5 0.5]
```

dpnp/backend/extensions/blas/blas_py.cpp
abagusetty: Agreed. Removed the separate `_gemv_alpha_beta` wrapper and folded `alpha`, `beta`, `trans_op` into `_gemv` with defaults

dpnp/backend/extensions/sparse/CMakeLists.txt
```diff
@@ -0,0 +1,122 @@
...
+        ${python_module_name}
+        PRIVATE -fprofile-instr-generate -fcoverage-mapping
+    )
+endif()
+
+if(_use_onemath)
+    target_link_libraries(${python_module_name} PRIVATE ${ONEMATH_LIB})
+    target_compile_options(${python_module_name} PRIVATE -DUSE_ONEMATH)
+    if(_use_onemath_cuda)
+        target_compile_options(${python_module_name} PRIVATE -DUSE_ONEMATH_CUSPARSE)
```
antonwolfy: `USE_ONEMATH_CUSPARSE` is never used, Should we remove it?

dpnp/scipy/sparse/_csr.py
antonwolfy: It'd be helpful to add a test for that use case.

dpnp/backend/extensions/blas/gemv.cpp
```diff
@@ -84,6 +92,15 @@ static sycl::event gemv_impl(sycl::queue &exec_q,
     const T *x = reinterpret_cast<const T *>(vectorX);
     T *y = reinterpret_cast<T *>(vectorY);
 
+    // Cast alpha/beta into the matrix value type. For complex T the
+    // single-argument constructor sets the imaginary component to
+    // zero, which is exact for the GMRES use case (alpha and beta are
+    // always one of {-1, 0, 1}) and for the dpnp.dot wrapper (alpha=1,
+    // beta=0). Callers passing fractional or complex scalars through
+    // this path would lose the imaginary component silently.
```
antonwolfy: The comment sounds inaccurate, there can't be complex scalars passed ,since input arguments declared as `double`, so:
```suggestion
    // beta=0).
```

dpnp/backend/extensions/blas/gemv.cpp
```diff
@@ -182,10 +217,22 @@ std::pair<sycl::event, sycl::event>
...
+    // Conjugate-transpose is only wired up for column-major (F-contig)
+    // matrices. The row-major remap (treating a C-contig matrix as its
+    // column-major transpose) does not extend cleanly to the C op
+    // because (A^T)^H == conj(A), which oneMKL does not expose as a
+    // gemv mode. Callers needing C-mode on row-major input must
+    // F-contigify first (e.g. via dpnp.asarray(A, order="F")).
+    if (is_conj_trans && !is_matrixA_f_contig) {
+        throw py::value_error(
+            "gemv: trans_op = 2 (conjugate-transpose) requires an "
+            "F-contiguous matrix; pass dpnp.asarray(A, order='F') first.");
```
antonwolfy: I'd rephrase a bit to align with other errors style:
```suggestion
            "Input matrix is not f-contiguous, but "
            "trans_op = 2 (conjugate-transpose) requested.");
```

dpnp/backend/extensions/blas/gemv.cpp
```diff
@@ -141,14 +158,32 @@ static sycl::event gemv_impl(sycl::queue &exec_q,
...
+// ``trans_op`` is a tri-state matching oneapi::mkl::transpose:
+//      0 = N (no transpose),
+//      1 = T (plain transpose),
+//      2 = C (conjugate-transpose, complex only).
+//
+// dpnp.dot only needs N/T. C exists for the GMRES Arnoldi step, which
+// needs V^H directly: post-conjugating a T-mode result is wrong for
+// complex u since conj(V^T @ u) == V^H @ u only holds when u is real.
+static std::pair<sycl::event, sycl::event>
+    gemv_dispatch(sycl::queue &exec_q,
```
antonwolfy: Why it can't be named `gemv` function as before? Why do we need the new `gemv_dispatch`?

dpnp/backend/extensions/blas/gemv.hpp
```diff
@@ -35,12 +35,32 @@
 
 namespace dpnp::extensions::blas
 {
+// y = alpha * op(A) * x + beta * y. alpha/beta come in as double and
+// are cast to the matrix value type in the impl, so complex callers
+// should stick to 1 / 0 / -1 to avoid silently dropping the imaginary
```
antonwolfy: Also unclear, how imaginary part can be dropped when `alpha/beta come in as double`?

dpnp/backend/extensions/sparse/gemv.cpp
```diff
@@ -0,0 +1,627 @@
...
+    type_utils::validate_type_for_device<Tv>(exec_q);
+
+    const Ti *row_ptr = reinterpret_cast<const Ti *>(row_ptr_data);
+    const Ti *col_ind = reinterpret_cast<const Ti *>(col_ind_data);
+    const Tv *values = reinterpret_cast<const Tv *>(values_data);
+
+    mkl_sparse::matrix_handle_t spmat = nullptr;
+    mkl_sparse::init_matrix_handle(&spmat);
+
+    auto ev_set = mkl_sparse::set_csr_data(
```
antonwolfy: Both `init_matrix_handle` and `set_csr_data` can through an exception, but no handling

dpnp/backend/extensions/sparse/gemv.cpp
```diff
@@ -0,0 +1,627 @@
...
+        exec_q, spmat, num_rows, num_cols, nnz, mkl::index_base::zero,
+        const_cast<Ti *>(row_ptr), const_cast<Ti *>(col_ind),
+        const_cast<Tv *>(values), depends);
+
+    sycl::event ev_opt;
+    try {
+        ev_opt = mkl_sparse::optimize_gemv(exec_q, mkl_trans, spmat, {ev_set});
+    } catch (mkl::exception const &e) {
+        mkl_sparse::release_matrix_handle(exec_q, &spmat, {});
```
antonwolfy: We have to pass dependent events to `release_matrix_handle` and to handle the returned event from `release_matrix_handle`.

antonwolfy: At least we need to have a blocking wait:

```suggestion
        mkl_sparse::release_matrix_handle(exec_q, &spmat, {ev_set}).wait();
```

dpnp/backend/extensions/sparse/gemv.cpp
```diff
@@ -0,0 +1,627 @@
...
+
+    if (val_type_id < 0 || val_type_id >= dpnp_td_ns::num_types)
+        throw py::value_error("sparse_gemv_compute: val_type_id out of range.");
+
+    gemv_compute_fn_ptr_t compute_fn = gemv_compute_dispatch_table[val_type_id];
+
+    if (compute_fn == nullptr)
+        throw py::value_error("sparse_gemv_compute: unsupported value dtype.");
+
+    return compute_fn(exec_q, handle_ptr, mkl_trans, alpha, x.get_data(), beta,
```
antonwolfy: Missing `keep_args_alive` call to handle lifetime (don't release memory) of `x` and `y` arrays

dpnp/scipy/sparse/_csr.py
abagusetty: Shoot, I missed this one. Just added. Thanks!

dpnp/backend/extensions/sparse/CMakeLists.txt
abagusetty: removed!

dpnp/backend/extensions/blas/gemv.cpp
abagusetty: Fixed

dpnp/backend/extensions/blas/gemv.cpp
abagusetty: Done!

dpnp/backend/extensions/blas/gemv.cpp
abagusetty: Fixed

dpnp/backend/extensions/blas/gemv.hpp
abagusetty: It was a wrong wording. I meant, real double becomes a complex with zero imaginary part. The comments were updated accordingly

dpnp/backend/extensions/sparse/gemv.cpp
abagusetty: Fixed

dpnp/backend/extensions/sparse/gemv.cpp
abagusetty: Fixed

dpnp/backend/extensions/sparse/gemv.cpp
abagusetty: Added the missing `keep_args_alive`

dpnp/scipy/sparse/linalg/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+Sparse linear algebra interface for DPNP.
+
+This module provides a subset of :mod:`scipy.sparse.linalg`
+ functionality on top of DPNP arrays.
+
+The initial implementation focuses on the :class:`LinearOperator` interface
+and a small set of Krylov solvers (``cg``, ``gmres``, ``minres``).
+"""
+
+from __future__ import annotations
```
antonwolfy: Do we need that? Is it used somewhere?

dpnp/scipy/sparse/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+and provides a location for future sparse matrix container types.
+"""
+
+from . import linalg
+from ._base import SparseABC, issparse
+from ._csr import csr_matrix
+
+__all__ = [
+    "linalg",
+    "SparseABC",
```
antonwolfy: Why do we need to expose `SparseABC` as part of `scipy.sparse` submodule?
SciPy does not did that.

dpnp/scipy/_lib/_sparse.py
```diff
@@ -0,0 +1,27 @@
+"""Sparse base class and predicate, mirroring scipy/_lib/_sparse.py.
```
antonwolfy: Should we mimic and move the file to `dpnp/scipy/_lib/_sparse.py`?

dpnp/scipy/_lib/_sparse.py
```diff
@@ -0,0 +1,27 @@
...
+see :mod:`scipy.sparse`) is intentionally omitted -- dpnp has no
+``spmatrix`` / ``sparray`` split, so the legacy names would have no
+useful semantics. Format-specific checks should use
+``issparse(A) and A.format == "csr"`` directly, which is what the
+solver fast-path already does.
+"""
+
+from abc import ABC
+
+
```
antonwolfy, line 35:
```suggestion
__all__ = ["SparseABC", "issparse"]

```

dpnp/scipy/_lib/_sparse.py
```diff
@@ -0,0 +1,27 @@
...
+# pylint: disable-next=too-few-public-methods
+class SparseABC(ABC):
+    """Abstract base for all dpnp.scipy.sparse format classes."""
+
+
+def issparse(x):
```
antonwolfy, line 41: We need to add rendering of documentation for the new sparse submodule.
Use `doc/reference/scipy_linalg.rst` as a reference.

dpnp/scipy/_lib/_sparse.py
antonwolfy: Add a copyright

dpnp/scipy/sparse/_base.py
```diff
@@ -0,0 +1,27 @@
...
+class SparseABC(ABC):
+    """Abstract base for all dpnp.scipy.sparse format classes."""
+
+
+def issparse(x):
+    """Return True if x is a dpnp sparse matrix.
+
+    Mirrors :func:`scipy.sparse.issparse` semantics: returns True for any
+    instance of :class:`SparseABC`, False otherwise.
+    """
```
antonwolfy: Let's align documentation style with other scipy-like functions:
```suggestion
    """
    Determine whether `x` is a dpnp sparse matrix type.

    For full documentation refer to :obj:`scipy.sparse.issparse`.

    Parameters
    ----------
    x : object
        Object to check for being a dpnp sparse matrix.

    Returns
    -------
    out : bool
        ``True`` if `x` is a dpnp sparse matrix, ``False`` otherwise.

    Examples
    --------
    >>> import dpnp
    >>> from dpnp.scipy.sparse import csr_matrix, issparse
    >>> issparse(csr_matrix(dpnp.eye(3)))
    True
    >>> issparse(dpnp.eye(3))
    False
    >>> issparse(5)
    False

    """
```

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
+"""CSR matrix backed by dpnp/USM arrays.
```
antonwolfy, line 29: Copyright header is missing

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
...
+        copy of another csr_matrix.
+
+    Duplicate column indices within a row are not supported (unlike
+    scipy, which sums them); each column must appear at most once per
+    row. This matches the CSR produced by dense construction and the
+    solvers, which never generate duplicates.
+
+    Attributes
+    ----------
+    data : dpnp.ndarray
```
antonwolfy: We need to support usm_ndarray also

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
...
+        # Lazy SpMV handle state. Assigned BEFORE the dispatch below so
+        # that __del__ never sees a partially-constructed object (it can
+        # be invoked if any of the _init_* helpers raise).
+        self._spmv_handle = None
+        self._spmv_val_type_id = -1
+        self._spmv_si = None
+        self._spmv_exec_q = None
+        self._has_sorted_indices = None
+
+        if isinstance(arg1, _dpnp.ndarray):
```
antonwolfy:
```suggestion
        if _dpnp.is_supported_array_type(arg1):
```

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
...
+            self._init_from_dense(arg1, dtype=dtype)
+        elif isinstance(arg1, csr_matrix):
```
antonwolfy: Should we use `‎issparse` here?

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
...
+        data, indices, indptr = arrays
+
+        if not (
+            isinstance(data, _dpnp.ndarray)
+            and isinstance(indices, _dpnp.ndarray)
+            and isinstance(indptr, _dpnp.ndarray)
+        ):
+            raise TypeError(
+                "csr_matrix: data, indices, and indptr must be dpnp arrays"
+            )
```
antonwolfy: We can use utility function:
```suggestion
        _dpnp.check_supported_arrays_type(data, indices, indptr)
```

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
...
+
+        nrows, ncols = int(shape[0]), int(shape[1])
+        if indptr.shape[0] != nrows + 1:
+            raise ValueError(
+                f"csr_matrix: indptr length {indptr.shape[0]} != "
+                f"nrows+1 ({nrows + 1})"
+            )
+
+        q = data.sycl_queue
+        if indices.sycl_queue != q or indptr.sycl_queue != q:
```
antonwolfy: We can use a common pattern for the check. like:
```python
if dpt.get_execution_queue([...]) is None:
    raise ExecutionPlacementError(...)
```

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
...
+
+        if not self._spmv_supported():
+            return None
+
+        try:
+            # Lazy import: keeps csr_matrix importable in builds that
+            # did not compile the sparse backend extension (e.g. host-
+            # only test matrices, doc builds).
+            # pylint: disable-next=import-outside-toplevel
+            from dpnp.backend.extensions.sparse import _sparse_impl as _si
```
antonwolfy: It a bit unclear why do we need to support that use case. We never did that with other backend extensions.
Is there any use case when that extension is not built? I was assuming it's done for every configuration, even when we building docs.

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
...
+            # Backend dispatch may reject the (value, index) pair even
+            # though the Python guard above accepted them (e.g. complex
+            # support disabled in the linked oneMKL build). Fall through
+            # to the dense path silently.
+            return None
+
+        # set_csr_data + optimize_gemv must complete before any compute
+        # call can dispatch against the handle. This is the only blocking
+        # sync; subsequent matvecs return without waiting.
+        ev.wait()
```
antonwolfy: Why do we need to have that call as a blocking? We can add the returned event under control of SequentialOrderManager.
Or alternatively we no need to return the event from `_sparse_gemv_init` and to implement the wait inside.

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
...
+                    0.0,  # beta
+                    y,
+                    nrows,
+                    ncols,
+                    _manager.submitted_events,
+                )
+                _manager.add_event_pair(ht_ev, comp_ev)
+                return y
+
+        # Dense fallback. Materialises ``self`` once -- this path is
```
antonwolfy: I wonder if it is a fare option to implement the fallback silently. It does not seem present in CuPy. CuPy raises the explicit exception for unsupported dtypes. While SciPy support any dtypes, since implements own sparse kernels.

In fallback case we have to allocate the full dense array, which may cause protentional OOM issue.
And so leaving the fallback implementation under the user control might be more friendly.

Also, as an option, we can implement upcasting of input arrays to support more wide range of dtypes.

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
...
+                _si, handle, val_type_id, exec_q = handle_info
+                # Reject dtype mismatches deterministically here rather
+                # than letting the C++ layer raise: callers expect a
+                # clean TypeError for cross-dtype matvec.
+                if x.dtype != self.data.dtype:
+                    raise TypeError(
+                        f"csr_matrix.dot: x dtype {x.dtype} does not "
+                        f"match matrix dtype {self.data.dtype}"
+                    )
+                y = _dpnp.empty(nrows, dtype=self.data.dtype, sycl_queue=exec_q)
```
antonwolfy:
```suggestion
                y = _dpnp.empty_like(self.data, nrows)
```

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
...
+        # evaluates to ``None``. Probe explicitly so a real backend
+        # error (extension still healthy) is not silenced by the same
+        # ``except Exception`` that catches the shutdown race.
+        handle = getattr(self, "_spmv_handle", None)
+        si = getattr(self, "_spmv_si", None)
+        if handle is None or si is None:
+            return
+
+        release_fn = getattr(si, "_sparse_gemv_release", None)
+        if release_fn is None:
```
antonwolfy: When there might be no `_sparse_gemv_release` available?

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,524 @@
...
+            self._spmv_handle = None
+            return
+
+        try:
+            # pylint: disable-next=not-callable
+            release_fn(self._spmv_exec_q, handle, [])
```
antonwolfy: Why no dependencies need to be passed?

antonwolfy: Also `release_fn` returns the SYCL event but seems no handling

dpnp/scipy/sparse/linalg/__init__.py
abagusetty: It is removed!

dpnp/scipy/sparse/_base.py
abagusetty: Done!

dpnp/scipy/_lib/_sparse.py
abagusetty: Done!

dpnp/scipy/_lib/_sparse.py
abagusetty: Done!

dpnp/scipy/_lib/_sparse.py
abagusetty: Moved it to mimic scipy style. Good idea indeed

dpnp/scipy/sparse/_csr.py
abagusetty: Done

dpnp/scipy/_lib/_sparse.py
abagusetty: Done!

dpnp/scipy/sparse/_csr.py
abagusetty: Done!

dpnp/scipy/sparse/_csr.py
abagusetty: added this one

dpnp/scipy/sparse/_csr.py
abagusetty: done

dpnp/scipy/sparse/_csr.py
abagusetty: Done!

dpnp/scipy/sparse/_csr.py
abagusetty: Done!

dpnp/scipy/sparse/_csr.py
abagusetty: Addressed it. There is no need to handle this in the if case

dpnp/scipy/sparse/_csr.py
abagusetty: Added a a`add_event_pair` to be consistent with the rest of codebase. Remove the blocking sync

dpnp/scipy/sparse/_csr.py
abagusetty: Dense fallback removed and dot raises `TypeError` for unsupported value/index dtype (CuPy-style, no O(M·N) densify/OOM). A few tests were added too

dpnp/scipy/sparse/_csr.py
abagusetty: this is fixed. The dep-events are passed and returned-events are handled

dpnp/scipy/sparse/_csr.py
abagusetty: `_sparse_gemv_release` is always compiled. This check was unneeded and removed

abagusetty:
@antonwolfy Apologies for the delay in getting back to your comments. All of them were addressed, whenever you get a chance. Thanks again!

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,648 @@
...
+    """Compressed Sparse Row matrix on a SYCL device.
+
+    Construction
+    ------------
+    csr_matrix(D)
+        from a 2-D array (``dpnp.ndarray`` or ``usm_ndarray``).
+
+    csr_matrix((M, N), [dtype=...])
+        an empty (all-zero) matrix of shape ``(M, N)``; ``dtype``
+        defaults to float64.
```
antonwolfy, line 105: It should be a default floating point data type for the device on which the input data is allocated

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,648 @@
...
+    ----------
+    data : {dpnp.ndarray, usm_ndarray}
+        1-D array of nonzero values, shape (nnz,).
+    indices : {dpnp.ndarray, usm_ndarray}
+        1-D array of column indices, shape (nnz,).
+    indptr : {dpnp.ndarray, usm_ndarray}
+        1-D array of row pointers, shape (M+1,).
+    shape : tuple of int
+    dtype : dpnp dtype
+    nnz : int
```
antonwolfy, line 139: Missing description of attributes

dpnp/scipy/sparse/linalg/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+# THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+
+"""
+Sparse linear algebra interface for DPNP.
+
+This module provides a subset of :mod:`scipy.sparse.linalg`
+ functionality on top of DPNP arrays.
```
antonwolfy, line 33:
```suggestion
functionality on top of DPNP arrays.
```

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,648 @@
...
+    row. This matches the CSR produced by dense construction and the
+    solvers, which never generate duplicates.
+
+    Supported operations: construction, ``dot`` (matvec) via cached
+    oneMKL SpMV, ``toarray``, ``copy``. This is a solver-support
+    subset of the scipy/cupy CSR API; arithmetic, indexing, reductions,
+    transpose, format conversion and element-wise math are not
+    implemented (the most common such methods raise
+    ``NotImplementedError``). Convert with ``toarray()`` and use dpnp
+    for those.
```
antonwolfy, line 127: It seems that has to be moved after the attributes block, otherwise rendered badly

dpnp/scipy/sparse/linalg/_interface.py
```diff
@@ -0,0 +1,780 @@
...
+                "LinearOperator subclass should implement at least one of "
+                "_matvec and _matmat.",
+                RuntimeWarning,
+                stacklevel=2,
+            )
+        return obj
+
+    def __init__(self, dtype, shape):
+        if dtype is not None:
+            dtype = dpnp.dtype(dtype)
```
antonwolfy, line 151: to ensure dtype is supported:
```suggestion
            dtype = dpnp.empty(0, dtype=dtype).dtype
```

dpnp/scipy/sparse/linalg/_interface.py
```diff
@@ -0,0 +1,780 @@
...
+written for either library is portable to dpnp.
+"""
+
+# Math-heavy module: single-letter and CamelCase identifiers such as
+# A, B, M, N, X, V, H are part of the published linear-algebra API and
+# mirror SciPy/CuPy verbatim, so the snake_case rule is intentionally
+# relaxed for the whole file.
+# pylint: disable=invalid-name
+
+from __future__ import annotations
```
antonwolfy, line 42: Do we need that?

dpnp/scipy/sparse/linalg/_interface.py
```diff
@@ -0,0 +1,780 @@
...
+        self.dtype = dtype
+        self.shape = shape
+
+    def _init_dtype(self):
+        """Infer dtype via a trial matvec on an int8 zero vector.
+
+        Using ``int8`` (the lowest precedence numeric dtype) lets the
+        matvec promote to its natural output type without artificially
+        widening the result -- a float32 operator stays float32, a
+        complex64 operator stays complex64, etc.  Mirrors the behaviour
```
antonwolfy, line 167:
```suggestion
        complex64 operator stays complex64, etc. Mirrors the behavior
```

dpnp/scipy/sparse/linalg/_interface.py
```diff
@@ -0,0 +1,780 @@
...
+        x : {dpnp.ndarray, usm_ndarray}
+            An array with shape ``(N,)`` or ``(N, 1)``.
+
+        Returns
+        -------
+        out : dpnp.ndarray
+            An array with shape ``(M,)`` or ``(M, 1)`` matching the rank
+            of `x`.
+        """
+        M, N = self.shape
```
antonwolfy, line 218: need to validate type of input array:
```suggestion
        dpnp.check_supported_arrays_type(x)
        M, N = self.shape
```

antonwolfy: the same check is missing for methods below

dpnp/scipy/sparse/linalg/_interface.py
```diff
@@ -0,0 +1,780 @@
...
+        The object to wrap. It may be any of the following:
+
+          * a :class:`LinearOperator` (returned unchanged);
+          * a ``dpnp.scipy.sparse`` sparse matrix, e.g. ``csr_matrix``
+            (the iterative solvers further specialise this to a cached
+            oneMKL SpMV handle, bypassing densification);
+          * a 2-D array, ``dpnp.ndarray`` or ``usm_ndarray`` (promoted
+            via :func:`dpnp.atleast_2d`);
+          * an object exposing ``.shape`` and ``.matvec`` (and optionally
+            ``rmatvec`` / ``matmat`` / ``rmatmat`` / ``dtype``).
```
antonwolfy, line 719: It has to be 8-space indent:

```suggestion
        * a :class:`LinearOperator` (returned unchanged);
        * a ``dpnp.scipy.sparse`` sparse matrix, e.g. ``csr_matrix``
          (the iterative solvers further specialise this to a cached
          oneMKL SpMV handle, bypassing densification);
        * a 2-D array, ``dpnp.ndarray`` or ``usm_ndarray`` (promoted
          via :func:`dpnp.atleast_2d`);
        * an object exposing ``.shape`` and ``.matvec`` (and optionally
          ``rmatvec`` / ``matmat`` / ``rmatmat`` / ``dtype``).
```

dpnp/scipy/sparse/linalg/_iterative.py
```diff
@@ -0,0 +1,1047 @@
...
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+# THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+
+"""Iterative sparse linear solvers for dpnp -- pure GPU/SYCL implementation.
+
+All computation stays on the device (USM/oneMKL).  There is NO host-dispatch
```
antonwolfy, line 31:
```suggestion
All computation stays on the device (USM/oneMKL). There is NO host-dispatch
```

dpnp/scipy/sparse/linalg/_interface.py
antonwolfy: The same issue with indents in some other docstrings

dpnp/scipy/sparse/linalg/_iterative.py
```diff
@@ -0,0 +1,1047 @@
...
+    else:
+        dtype = dpnp.float64
+
+    b = b.astype(dtype, copy=False)
+    _check_dtype(b.dtype, "b")
+
+    if x0 is None:
+        x = dpnp.zeros(n, dtype=dtype, sycl_queue=b.sycl_queue)
+    else:
+        x = x0.astype(dtype, copy=True).reshape(-1)
```
antonwolfy, line 243: `x0` might be `usm_ndarray`:
```suggestion
        x = dpnp.astype(x0, dtype, copy=True).reshape(-1)
```

dpnp/tests/test_scipy_sparse_linalg.py
```diff
@@ -0,0 +1,1733 @@
...
+        m = csr_matrix((data, indices, indptr), shape=(1, 3))
+        m.sort_indices()
+        idx_after = list(dpnp.asnumpy(m.indices))
+        data_after = list(dpnp.asnumpy(m.data))
+        m.sort_indices()
+        assert m.has_sorted_indices
+        assert list(dpnp.asnumpy(m.indices)) == idx_after
+        assert list(dpnp.asnumpy(m.data)) == data_after
+
+    def test_check_sorted_detects_sorted_components(self):
```
antonwolfy, line 1509: Failed on device without fp64 with:
> ValueError: Device Intel(R) Iris(R) Xe Graphics does not provide native support for double-precision floating point type.

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,648 @@
...
+            "sycl_queue": sycl_queue,
+        }
+        self.data = _dpnp.empty(0, dtype=dtype, **common)
+        idx_dtype = _dpnp.int64
+        self.indices = _dpnp.empty(0, dtype=idx_dtype, **common)
+        self.indptr = _dpnp.zeros(nrows + 1, dtype=idx_dtype, **common)
+        self._shape = (nrows, ncols)
+        self._has_sorted_indices = True
+
+    def _init_from_components(self, arrays, shape, dtype=None, copy=False):
```
antonwolfy, line 220: Do we have a missing validation against scipy from example below?
```python
import numpy as np
import scipy.sparse as sp

data, indices, indptr = np.array([1.,2.,3.]), np.array([0,1,2],np.int32), np.array([0,1,5],np.int32)
sp.csr_matrix((data, indices, indptr), shape=(2,3))
Cell In[2], line 2
      1 data, indices, indptr = np.array([1.,2.,3.]), np.array([0,1,2],np.int32), np.array([0,1,5],np.int32)
----> 2 sp.csr_matrix((data, indices, indptr), shape=(2,3))

File /localdisk/work/antonvol/miniforge3/envs/dpnp_dev/lib/python3.13/site-packages/scipy/sparse/_compressed.py:116, in _cs_matrix.__init__(self, arg1, shape, dtype, copy, maxprint)
    113     newdtype = getdtype(dtype)
    114     self.data = self.data.astype(newdtype, copy=False)
--> 116 self.check_format(full_check=False)

File /localdisk/work/antonvol/miniforge3/envs/dpnp_dev/lib/python3.13/site-packages/scipy/sparse/_compressed.py:203, in _cs_matrix.check_format(self, full_check)
    201     raise ValueError("indices and data should have the same size")
    202 if (self.indptr[-1] > len(self.indices)):
--> 203     raise ValueError("Last value of index pointer should be less than "
    204                      "the size of index and data arrays")
    206 self.prune()
    208 if full_check:
    209     # check format validity (more expensive)

ValueError: Last value of index pointer should be less than the size of index and data arrays

import dpnp as np
import dpnp.scipy.sparse as sp

data, indices, indptr = np.array([1.,2.,3.]), np.array([0,1,2],np.int32), np.array([0,1,5],np.int32)
sp.csr_matrix((data, indices, indptr), shape=(2,3))
# Out: <2x3 csr_matrix of dtype float64 with 3 stored elements>
```

antonwolfy, line 220: dpnp might pass 7 to oneMKL -> OOB read of length-3 x for below example:
```python
sp.csr_matrix((np.array([1.,2.]), np.array([0,7],np.int32), np.array([0,1,2],np.int32)), shape=(2,3))
# Out: <2x3 csr_matrix of dtype float64 with 2 stored elements>
```
while scipy:
```python
a = sp.csr_matrix((np.array([1.,2.]), np.array([0,7],np.int32), np.array([0,1,2],np.int32)), shape=(2,3))
a.check_format()
Cell In[9], line 1
----> 1 a.check_format()

File /localdisk/work/antonvol/miniforge3/envs/dpnp_dev/lib/python3.13/site-packages/scipy/sparse/_compressed.py:212, in _cs_matrix.check_format(self, full_check)
    210 if self.nnz > 0:
    211     if self.indices.max() >= N:
--> 212         raise ValueError(f"indices must be < {N}")
    213     if self.indices.min() < 0:
    214         raise ValueError("indices must be >= 0")

ValueError: indices must be < 3
```

antonwolfy, line 220: - scipy: silent at construction, ValueError in check_format
- dpnp: row_lengths = diff = [2,-1] -> negative count to dpnp.repeat / bad offsets to oneMKL

```python
sp.csr_matrix((np.array([1.,2.,3.]), np.array([0,1,2],np.int32), np.array([0,2,1],np.int32)), shape=(2,3))
```

dpnp/backend/extensions/sparse/gemv.cpp
```diff
@@ -0,0 +1,653 @@
...
+                exec_q, mkl_trans, &alpha, cache->view, cache->A, cache->x,
+                &beta, cache->y, alg, cache->descr, cache->workspace, depends);
+            cache->optimized = true;
+
+            return mkl_sparse::spmv(exec_q, mkl_trans, &alpha, cache->view,
+                                    cache->A, cache->x, &beta, cache->y, alg,
+                                    cache->descr, {ev_opt});
+        }
+
+        return mkl_sparse::spmv(exec_q, mkl_trans, &alpha, cache->view,
```
antonwolfy, line 350: alpha/beta passed by address of stack, because oneMath `spmv` takes `const void* alpha` and so might be destroyed when `gemv_compute_impl` returns

dpnp/scipy/sparse/linalg/_interface.py
```diff
@@ -0,0 +1,780 @@
...
+        # csr_matrix.dot is 1-D-only (like cupyx); x is already 1-D here.
+        if issparse(self.A):
+            return self.A.dot(x)
+        return super()._matvec(x)
+
+    def _matmat(self, X):
+        # No native SpMM: emulate as a column loop of 1-D SpMVs (no densify).
+        if issparse(self.A):
+            return dpnp.stack(
+                [self.A.dot(X[:, i]) for i in range(X.shape[1])], axis=-1
```
antonwolfy, line 631: `self.A.dot(X[:, i])` on a C-contiguous 2-D `X` passes a non-unit-stride view.
`sparse_gemv_compute` only checks ndim/shape/dtype and reads `x.get_data()` contiguously → wrong results for `L @ X/L.matmat(X)` with ≥2 columns, no error.

dpnp/scipy/sparse/_csr.py
```diff
@@ -0,0 +1,648 @@
...
+
+        if not self._spmv_supported():
+            return None
+
+        self.sort_indices()
+
+        exec_q = self.data.sycl_queue
+        _manager = _dpu.SequentialOrderManager[exec_q]
+        # pylint: disable-next=protected-access
+        handle, val_type_id, ev = _si._sparse_gemv_init(
```
antonwolfy, line 446: The below example raises exception here:
```python
import dpnp as np
import dpnp.scipy.sparse as sp

b0 = np.zeros(4, dtype=np.float64)
A_empty = sp.csr_matrix((4, 4), dtype=np.float64)

sp.linalg.cg(A_empty, b0, rtol=1e-8)
Cell In[12], line 1
----> 1 sp.linalg.cg(A_empty, b0, rtol=1e-8)

File /localdisk/work/antonvol/code/dpnp/dpnp/scipy/sparse/linalg/_iterative.py:373, in cg(A, b, x0, rtol, tol, maxiter, M, callback, atol)
    365     warnings.warn(
    366         "'tol' is deprecated in favor of 'rtol' and will be removed in "
    367         "a future release; use 'rtol' instead.",
    368         DeprecationWarning,
    369         stacklevel=2,
    370     )
    371     rtol = tol
--> 373 A_op, M_op, x, b, dtype = _make_system(A, M, x0, b)
    374 n = b.shape[0]
    376 bnrm = dpnp.linalg.norm(b)

File /localdisk/work/antonvol/code/dpnp/dpnp/scipy/sparse/linalg/_iterative.py:271, in _make_system(A, M, x0, b)
    268         M_op = _FastMOp()
    270 # Inject fast CSR SpMV for A if available.
--> 271 fast_mv = _make_fast_matvec(A)
    272 if fast_mv is not None:
    273     _orig = A_op

File /localdisk/work/antonvol/code/dpnp/dpnp/scipy/sparse/linalg/_iterative.py:178, in _make_fast_matvec(A)
    176     return None
    177 # pylint: disable-next=protected-access
--> 178 handle_info = A._ensure_spmv_handle()
    179 if handle_info is None:
    180     return None

File /localdisk/work/antonvol/code/dpnp/dpnp/scipy/sparse/_csr.py:445, in csr_matrix._ensure_spmv_handle(self)
    443 _manager = _dpu.SequentialOrderManager[exec_q]
    444 # pylint: disable-next=protected-access
--> 445 handle, val_type_id, ev = _si._sparse_gemv_init(
    446     exec_q,
    447     0,  # trans=N (forward)
    448     _dpnp.get_usm_ndarray(self.indptr),
    449     _dpnp.get_usm_ndarray(self.indices),
    450     _dpnp.get_usm_ndarray(self.data),
    451     int(self._shape[0]),
    452     int(self._shape[1]),
    453     int(self.data.shape[0]),
    454     _manager.submitted_events,
    455 )
    457 # set_csr_data + optimize_gemv must complete before the first
    458 # compute; chain the init event through the queue's order manager
    459 # so the first _sparse_gemv_compute depends on it (non-blocking).
    460 _manager.add_event_pair(ev, ev)

RuntimeError: sparse_gemv_init: MKL exception in set_csr_data: oneapi::mkl::sparse::set_csr_data: invalid argument: nnz is = 0 provided with at least one of `ja`/`a` != nullptr
```
but passes with scipy

abagusetty:
@antonwolfy Is there a timeline for the 0.21.0 release/code freeze. Just wanted to prioritize this PR accordingly on my end

antonwolfy:
> @antonwolfy Is there a timeline for the 0.21.0 release/code freeze. Just wanted to prioritize this PR accordingly on my end

@abagusetty, as for now, we are planning the freeze by mid of September.

dpnp/scipy/sparse/_csr.py
abagusetty: Fixed!

dpnp/scipy/sparse/linalg/_interface.py
abagusetty: Fixed. Thanks for catching this. `sparse_gemv_compute` now rejects non-contiguous `x/y` rather than silently consuming it. Added few unit-tests to cover this!

dpnp/backend/extensions/sparse/gemv.cpp
abagusetty: Fixed all the instance of this!

dpnp/scipy/sparse/_csr.py
abagusetty: Fixed it to be similar like scipy. It is silent at construction but `check_format` runs the checks lazily on the first use now

dpnp/tests/test_scipy_sparse_linalg.py
```diff
@@ -0,0 +1,1733 @@
...
+        )
+        m = csr_matrix(dense)
+        assert m.has_sorted_indices
+
+    @with_requires("scipy")
+    def test_sort_indices_multirow(self):
+        import scipy.sparse as sp
+
+        # 3 rows, columns shuffled per row.
+        data = dpnp.array([3, 1, 2, 5, 4, 6], dtype=dpnp.float64)
```
antonwolfy:
```suggestion
        data = dpnp.array([3, 1, 2, 5, 4, 6], dtype=dpnp.float32)
```
otherwise failing on device without fp64 support

dpnp/scipy/sparse/linalg/_interface.py
```diff
@@ -0,0 +1,780 @@
...
+        return False
+
+
+def _get_dtype(operators, dtypes=None):
+    if dtypes is None:
+        dtypes = []
+    for obj in operators:
+        if obj is not None and hasattr(obj, "dtype") and obj.dtype is not None:
+            dtypes.append(obj.dtype)
+    return dpnp.result_type(*dtypes) if dtypes else None
```
antonwolfy, line 80: That causes tests failure on device without fp64:
```
FAILED tests/test_scipy_sparse_linalg.py::TestLinearOperatorAlgebra::test_scaled_linear_operator[float32] - ValueError: Device Intel(R) Iris(R) Xe Graphics does not provide native support for double-precision floating point type.
--
tests/test_scipy_sparse_linalg.py::TestLinearOperatorAlgebra::test_scaled_linear_operator[complex64] - ValueError: Device Intel(R) Iris(R) Xe Graphics does not provide native support for double-precision floating point type.
```

PR #2843, opened 2026-04-10
Merge master into include-dpctl-tensor

vlad-perevezentsev:
This PR synchronizes the `include-dpctl-tensor` branch with the current `master` branch


ndgrigorian (dismissed):
tests pass, LGTM

dpnp/backend/extensions/indexing/CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ set_target_properties(
 target_include_directories(
     ${python_module_name}
     PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/../../
```
antonwolfy, line 69: is that correct? master has:
```
target_include_directories(
    ${python_module_name}
    PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../../ ${CMAKE_CURRENT_SOURCE_DIR}/../common
)
```

dpnp/backend/extensions/indexing/CMakeLists.txt
antonwolfy: Does we still have `Dpctl_TENSOR_INCLUDE_DIR` after the migration?

dpnp/backend/extensions/statistics/CMakeLists.txt
```diff
@@ -71,6 +71,7 @@ set_target_properties(
 target_include_directories(
     ${python_module_name}
     PRIVATE
+        ${CMAKE_CURRENT_SOURCE_DIR}/../../
```
antonwolfy, line 74: the same comments/questions here as for indexing CMake file

dpnp/backend/extensions/indexing/CMakeLists.txt
vlad-perevezentsev: If `pre-commit` is successful,then this is the correct style for it

Maybe the two lines could be combined into one but we now have four

antonwolfy: My question wan mainly, why do we have 4 includes now here? what is changed that affected legacy list of includes?

dpnp/backend/extensions/indexing/CMakeLists.txt
vlad-perevezentsev: Great catch!
I guess I forgot to delete this in the last PRs

dpnp/backend/extensions/indexing/CMakeLists.txt
antonwolfy: To be more specific: why do we need `${CMAKE_SOURCE_DIR}/dpnp/backend/include` ?

dpnp/__main__.py
antonwolfy:
```suggestion
# Copyright (c) 2026, Intel Corporation
```

dpnp/backend/extensions/indexing/CMakeLists.txt
vlad-perevezentsev: Previously this header came from dpctl and was found via `${Dpctl_INCLUDE_DIRS}`
After the migration the explicit path is required because without it I get 
`fatal error: 'dpnp4pybind11.hpp' file not found` during the build

Also you added these changes yourself when you introduced `dpnp4pybind11.hpp `
https://github.com/IntelPython/dpnp/pull/2734/changes#diff-184866ecda4860667a4d6bf8df98717810e828d440988865ee60b113e02fea6bR67

dpnp/backend/extensions/lapack/CMakeLists.txt
antonwolfy: that include is not needed:
```suggestion
```

antonwolfy (approved):
No more comments from me

PR #2844, opened 2026-04-10
Add `dpnp.tensor` documentation

vlad-perevezentsev:
This PR provides Sphinx documentation for the `dpnp.tensor` module, adapted from dpctl.
It also updates the `dpnp.tensor` docstrings in Sphinx style and replaces references to `dpctl.tensor` with `dpnp.tensor` in error messages and comments.

>- API reference pages covering all `dpnp.tensor` public API: array object, creation, elementwise, accumulation, manipulation, indexing, linear algebra, searching, set, sorting, statistical, printing, utility functions, data types, constants, and inspection
>- Autosummary templates (`usm_ndarray`, `cython_class`, `elementwise`) for proper rendering of classes and elementwise functions
>- User guides for tensor introduction, oneAPI execution model, and DLPack exchange
>- Fix for `usm_ndarray` docstring rendering issues 


ndgrigorian:
I am not convinced the changes make sense, since I think it would be smart to update dpnp documentation to align with dpctl documentation in the future, using furo style. In which case this will have to be rolled back.

If this change is going to be made, as hard to manage as it may be, it should probably be a single commit... this way, it would be easy to revert, so we don't have to go through all of the manual work.

@antonwolfy any thoughts?

PR #2851, opened 2026-04-14
Add device-aware output dtype for `dpt.round()` with boolean input

vlad-perevezentsev:
This PR proposes device-aware output dtype resolution for `dpnp.tensor.round()` with `boolean` input to handle devices that do not support `float16`

Boolean support for round() was originally added in #2817 [6f5a792](https://github.com/IntelPython/dpnp/pull/2817/changes/6f5a792c8542143b395ba1c4e6e44e6bdaf85576) to match NumPy behavior where numpy.round(bool) returns float16 rather than an integral type like int8.
However on devices without fp16 support, returning float16 is not viable.

The bool type mapping was removed from the round kernel and an acceptance
function `_acceptance_fn_round` was added to ensure the fallback in `_find_buf_dtype`
prefers floating-point output over integral types for boolean input

result 
fp16 devices: round(bool) -> float16 
non-fp16 devices: round(bool) -> float32

### DIFF
--- a/dpnp/tensor/_type_utils.py
+++ b/dpnp/tensor/_type_utils.py
@@ -135,7 +135,7 @@ def _acceptance_fn_reciprocal(arg_dtype, buf_dt, res_dt, sycl_dev):
 
 def _acceptance_fn_round(arg_dtype, buf_dt, res_dt, sycl_dev):
     # for boolean input, prefer floating-point output over integral
-    if arg_dtype.char == "?" and res_dt.kind in "biu":
+    if arg_dtype.kind == "b" and res_dt.kind != "f":
         return False
     return True
 
@@ -195,19 +195,17 @@ def _dtype_supported_by_device_impl(
 
 
 def _find_buf_dtype(arg_dtype, query_fn, sycl_dev, acceptance_fn):
-    _fp16 = sycl_dev.has_aspect_fp16
-    _fp64 = sycl_dev.has_aspect_fp64
-
     res_dt = query_fn(arg_dtype)
     if res_dt:
-        if _dtype_supported_by_device_impl(res_dt, _fp16, _fp64):
-            return None, res_dt
+        return None, res_dt
 
+    _fp16 = sycl_dev.has_aspect_fp16
+    _fp64 = sycl_dev.has_aspect_fp64
     all_dts = _all_data_types(_fp16, _fp64)
     for buf_dt in all_dts:
         if _can_cast(arg_dtype, buf_dt, _fp16, _fp64):
             res_dt = query_fn(buf_dt)
-            if res_dt and _dtype_supported_by_device_impl(res_dt, _fp16, _fp64):
+            if res_dt:
                 acceptable = acceptance_fn(arg_dtype, buf_dt, res_dt, sycl_dev)
                 if acceptable:
                     return buf_dt, res_dt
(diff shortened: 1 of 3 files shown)


dpnp/tensor/libtensor/include/kernels/elementwise_functions/round.hpp
```diff
@@ -116,7 +116,6 @@ template <typename T>
 struct RoundOutputType
 {
     using value_type = typename std::disjunction<
-        td_ns::TypeMapResultEntry<T, bool, sycl::half>,
         td_ns::TypeMapResultEntry<T, std::uint8_t>,
```
ndgrigorian, line 119: Is there reason not to just add a `float` loop with `sycl::half`?

Tensor type finding generally works such that, if one type can't work, it will check the next available loop. We use this approach throughout elementwise funcs, where if one type would be unavailable, we just add a loop for a more common type.

dpnp/tensor/_type_utils.py
```diff
@@ -133,6 +133,13 @@ def _acceptance_fn_reciprocal(arg_dtype, buf_dt, res_dt, sycl_dev):
         return True
 
 
+def _acceptance_fn_round(arg_dtype, buf_dt, res_dt, sycl_dev):
+    # for boolean input, prefer floating-point output over integral
+    if arg_dtype.char == "?" and res_dt.kind in "biu":
```
ndgrigorian:
```suggestion
    if arg_dtype.kind == "b" and res_dt.kind != "f":
```
using kind for `bool` over char is a nit for consistency, but `res_dt.kind != "f"` should be marginally faster and is clearer

dpnp/tensor/libtensor/include/kernels/elementwise_functions/round.hpp
ndgrigorian: actually, ignore this comment, this is right change :) I had forgotten a bit myself how the type searching works...

dpnp/tensor/_type_utils.py
```diff
@@ -188,17 +195,19 @@ def _dtype_supported_by_device_impl(
 
 
 def _find_buf_dtype(arg_dtype, query_fn, sycl_dev, acceptance_fn):
+    _fp16 = sycl_dev.has_aspect_fp16
+    _fp64 = sycl_dev.has_aspect_fp64
+
     res_dt = query_fn(arg_dtype)
```
ndgrigorian: Should `_find_buf_dtype` be changed at all?

With this change if a bad loop (i.e., one that maps a dtype to a different dtype that can sometimes not be supported, like `i->d` or even `e->d`) should more often make it to having output types, but I am not sure this gets us much benefit, since those loops are malformed anyway, and it adds (maybe small) overhead to the function.

ndgrigorian (dismissed):
aside from above points LGTM

dpnp/tensor/_type_utils.py
vlad-perevezentsev: It makes sense
Done, thank you!

dpnp/tensor/_type_utils.py
vlad-perevezentsev: You are right
Adding `_acceptance_fn_round` already provides the needed fix so `_find_buf_dtype` changes are not necessary 
Thank you

antonwolfy (approved):
LGTM
Thank you @vlad-perevezentsev

PR #2854, opened 2026-04-15
Fix `meshgrid` return type

antonwolfy:
This PR fixes `dpnp.meshgrid` and `dpnp.tensor.meshgrid` to always return tuple instead of list.

Changes `meshgrid()` to consistently return a tuple to match NumPy's behavior (version 2.5+) and align with the newer version 2025.12 of Python array API spec.


dpnp/dpnp_iface_arraycreation.py
```diff
@@ -3132,7 +3132,7 @@ def meshgrid(*xi, copy=True, sparse=False, indexing="xy"):
     if copy:
         output = [dpt.copy(x) for x in output]
 
-    return [dpnp_array._create_from_usm_ndarray(x) for x in output]
```
ndgrigorian, line 3135: should we do same for `dpnp.tensor.meshgrid`?

antonwolfy: Yep, valid. Added updates to the tensor.

vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2859, opened 2026-04-16
 Use `dpctl_capi.h` header in `dpnp4pybind11.hpp`

vlad-perevezentsev:
This PR proposes replacing the individual dpctl C-API includes and import calls in `dpnp4pybind11.hpp` with `dpctl_capi.h` header and its `import_dpctl()` function.
Previous changes to the explicit imports were made to avoid conflicting with dpctl which contained a tensor;
now that dpctl no longer contains a tensor, we can use the header directly.

Additionally, fixes `RuntimeWarning` filter in `test_exp_complex_contig` 

### DIFF
--- a/dpnp/backend/include/dpnp4pybind11.hpp
+++ b/dpnp/backend/include/dpnp4pybind11.hpp
@@ -28,29 +28,8 @@
 
 #pragma once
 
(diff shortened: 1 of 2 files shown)


antonwolfy (approved):
LGTM

PR #2862, opened 2026-04-19
Include `dpctl4pybind11.hpp` in `dpnp4pybind11.hpp` and switch libtensor to `dpnp` namespace

ndgrigorian:
This PR proposes including `dpctl4pybind11.hpp` in `dpnp4pybind11.hpp`.

* dpnp4pybind11 reimplemented most of the logic from dpctl4pybind11 unnecessarily, including SYCL casters, which should not be the responsibility of dpnp to implement, and management of memory objects, also not dpnp's responsibility
* As `dpnp4pybind11.hpp` is still using the dpctl namespace, the reimplementations could potentially cause code using both `dpctl4pybind11.hpp` and `dpnp4pybind11.hpp` together to crash due to multiple definitions in one translation unit. This can be problematic when both libraries are attempting to expose headers, with dpnp depending directly on dpctl

As part of this change, this PR also changes the `dpctl` namespace in libtensor to `dpnp`


ndgrigorian:
This PR can be split to implement the namespace changes in a follow-up, only the first 3 commits (and the part about adding back `keep_args_alive`) need to be part of this PR in theory, though includes with dpctl4pybind11 could break

but regardless, it should go in before migration is complete @antonwolfy @vlad-perevezentsev

dpnp/backend/include/dpnp4pybind11.hpp
```diff
@@ -1295,26 +666,13 @@ sycl::event keep_args_alive(sycl::queue &q,
...
-{
-    for (std::size_t i = 0; i < num; ++i) {
-
-        if (exec_q != alloc_qs[i]) {
-            return false;
-        }
-    }
-    return true;
-}
+using ::dpctl::utils::queues_are_compatible;
```
antonwolfy, line 667: please add a comment to clarify in the code why we need the overload from dpctl, since it's not so much obvious

dpnp/backend/extensions/blas/dot_common.hpp
```diff
@@ -32,7 +32,7 @@
 
 #include <pybind11/pybind11.h>
 
-// dpctl tensor headers
+// dpnp tensor headers
```
antonwolfy, line 35: need to update the comment in `dpnp/backend/extensions/fft/in_place.tpp`, `dpnp/backend/extensions/fft/out_of_place.tpp`.

antonwolfy: and probably in `dpnp/linalg/dpnp_utils_linalg.py`, `dpnp/scipy/linalg/_utils.py`, since there is `use DPCTL tensor function` in the commment

dpnp/backend/include/dpnp4pybind11.hpp
```diff
@@ -1189,19 +556,24 @@ namespace utils
...
+
 struct ManagedMemory
 {
-
+    // TODO: do we need to check for memory here? Or can we assume only
+    // dpnp::tensor::usm_ndarray will be passed?
     static bool is_usm_managed_by_shared_ptr(const py::object &h)
     {
+
         if (py::isinstance<dpctl::memory::usm_memory>(h)) {
```
antonwolfy: should we use leading `::` also here?
```suggestion
        if (py::isinstance<::dpctl::memory::usm_memory>(h)) {
```

antonwolfy: ans the same is applicable below

antonwolfy: and in `dpnp/tensor/libtensor/source/device_support_queries.cpp`::119

dpnp/backend/include/dpnp4pybind11.hpp
```diff
@@ -42,7 +42,6 @@
 #include <cstddef> // for std::size_t for C++ linkage
 #include <cstdint>
 #include <memory>
-#include <stddef.h> // for size_t for C linkage
 #include <stdexcept>
 #include <type_traits>
```
antonwolfy: probably unused:
```suggestion
```

antonwolfy: and also `#include <complex>` above

dpnp/backend/include/dpnp4pybind11.hpp
```diff
@@ -890,12 +257,12 @@ class usm_ndarray : public py::object
 public:
     PYBIND11_OBJECT(usm_ndarray, py::object, [](PyObject *o) -> bool {
         return PyObject_TypeCheck(
-                   o, ::dpctl::detail::dpctl_capi::get().PyUSMArrayType_) != 0;
+                   o, ::dpnp::detail::dpnp_capi::get().PyUSMArrayType_) != 0;
```
antonwolfy: we inside `dpnp` namespace, it'd better to use the relative namespace path:
```suggestion
                   o, detail::dpnp_capi::get().PyUSMArrayType_) != 0;
```

dpnp/backend/include/dpnp4pybind11.hpp
ndgrigorian: forgot the `device_support_queries` until last commit

antonwolfy (approved):
Thank you @ndgrigorian , no more comments from me

antonwolfy:
The CI failures seems unrelating, so merging

PR #2871, opened 2026-04-23
Fix RNG crash on Windows 

vlad-perevezentsev:
This PR adds a Windows workaround for a crash on Battlemage during test runs.

In oneMKL 2026.0 `mkl_rng::mt19937` destructor may crash with `STATUS_STACK_BUFFER_OVERRUN` during `DLL_PROCESS_DETACH` when using the Level Zero backend and static storage inside a DLL.

The workaround heap-allocates `backend_sycl` singleton on Windows to avoid running its destructor at shutdown. Although `~backend_sycl()` is empty, it triggers destruction of its members (including `mkl_rng::mt19937`) which causes the crash. 
Since `backend_sycl` is a process-lifetime singleton, skipping its destruction is safe because OS reclaims all
resources at process exit.

### DIFF
--- a/dpnp/backend/src/queue_sycl.hpp
+++ b/dpnp/backend/src/queue_sycl.hpp
@@ -67,7 +67,7 @@ public:
 
     static backend_sycl &get()
     {
-#ifdef _WIN32
+#if defined(_WIN32) && INTEL_MKL_VERSION == 20260000
         // TODO: remove once MKLD-19835 is resolved
         // mt19937 (oneMKL 2026.0) destructor crashes during DLL_PROCESS_DETACH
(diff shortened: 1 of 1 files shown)


dpnp/backend/src/queue_sycl.hpp
```diff
@@ -67,8 +67,17 @@ class backend_sycl
 
     static backend_sycl &get()
     {
+#ifdef _WIN32
```
antonwolfy: Can we add any dependency on MKL version also?

vlad-perevezentsev: done

antonwolfy (approved):
Thank you @vlad-perevezentsev

PR #2877, opened 2026-04-29
Refactor tensor elementwise tests

ndgrigorian:
This PR continues work in dpctl which was intended to reduce redundancy in the `tensor` test suite, particularly in elementwise tests


antonwolfy:
Are we ready to enable tensor tests by default? Or more refactoring required to be done in the follow-up PRs?

ndgrigorian:
@antonwolfy 
I think we should see how internal CI handles the pressure of these tests. Maybe after this is merged, we can open draft PR enabling, see how a few runs go, then decide

ndgrigorian:
@antonwolfy ping

dpnp/tests/tensor/elementwise/test_usm_types.py
```diff
@@ -0,0 +1,307 @@
...
+
+
+@pytest.mark.parametrize("usm_type", _usm_types)
+class TestUnaryUSMType:
+    def unary_elementwise(self, fn, usm_type, dtype="f4"):
+        q = get_queue_or_skip()
+        x = dpt.asarray(
+            [1, 2, 3, 4], dtype=dtype, usm_type=usm_type, sycl_queue=q
+        )
+        return getattr(dpt, fn)(x)
```
antonwolfy: Do we need to assert against input `usm_type` here?

dpnp/tests/tensor/elementwise/test_usm_types.py
```diff
@@ -0,0 +1,307 @@
...
+class TestBinaryUSMType:
+    def binary_elementwise(self, fn, op1_usm_type, op2_usm_type, dtype="f4"):
+        q = get_queue_or_skip()
+        x = dpt.asarray(
+            [1, 2, 3, 4, 5, 6], dtype=dtype, usm_type=op1_usm_type, sycl_queue=q
+        )
+        y = dpt.asarray(
+            [1, 2, 3, 4, 5, 6], dtype=dtype, usm_type=op2_usm_type, sycl_queue=q
+        )
+        return getattr(dpt, fn)(x, y)
```
antonwolfy: The same here

antonwolfy (approved):
Only small nits below, but LGTM in overall.

dpnp/tests/tensor/elementwise/test_usm_types.py
ndgrigorian: we can, I only opted not to because all of the USM type logic takes place in the common parts of the elementwise functions.

But it's pretty trivial to add so I will go ahead and do it

ndgrigorian:
@antonwolfy 
implemented requested changes

antonwolfy (approved):
Thank you @ndgrigorian, LGTM

PR #2889, opened 2026-05-04
Support buffer protocol objects as advanced index keys

vlad-perevezentsev:
This PR proposes by adding support for buffer protocol objects (`array.array`, `memoryview` etc.) as advanced index keys
These changes were proposed in [#2872](https://github.com/IntelPython/dpnp/pull/2872#discussion_r3150370962) as an extension to the advanced indexing support.

### DIFF
--- a/dpnp/dpnp_array.py
+++ b/dpnp/dpnp_array.py
@@ -38,6 +38,8 @@ elements stored in a USM allocation on a SYCL device.
 
 import warnings
 
+import numpy
+
 import dpnp
 import dpnp.tensor as dpt
 import dpnp.tensor._type_utils as dtu
@@ -46,23 +48,39 @@ from . import memory as dpm
 from .exceptions import AxisError
 
 
+def _unwrap_index_element(x):
+    """
+    Unwrap a single index element for the tensor indexing layer.
+
+    Converts dpnp arrays to usm_ndarray and array-like objects (range, list)
+    to numpy arrays with intp dtype for NumPy-compatible advanced indexing.
+
+    """
+
+    if isinstance(x, dpnp_array):
+        return x.get_array()
+    if isinstance(x, (range, list)):
+        return numpy.asarray(x, dtype=numpy.intp)
+    return x
+
+
 def _get_unwrapped_index_key(key):
     """
     Get an unwrapped index key.
 
     Return a key where each nested instance of DPNP array is unwrapped into
-    USM ndarray for further processing in DPCTL advanced indexing functions.
+    USM ndarray, and array-like objects (range, list) are converted to numpy
+    arrays for further processing in advanced indexing functions.
 
     """
 
     if isinstance(key, tuple):
-        if any(isinstance(x, dpnp_array) for x in key):
-            # create a new tuple from the input key with unwrapped DPNP arrays
-            return tuple(
-                x.get_array() if isinstance(x, dpnp_array) else x for x in key
-            )
+        if any(isinstance(x, (dpnp_array, range, list)) for x in key):
+            return tuple(_unwrap_index_element(x) for x in key)
     elif isinstance(key, dpnp_array):
         return key.get_array()
+    elif isinstance(key, (range, list)):
+        return numpy.asarray(key, dtype=numpy.intp)
     return key
 
 
--- a/dpnp/dpnp_array.py
+++ b/dpnp/dpnp_array.py
@@ -57,6 +57,8 @@ def _unwrap_index_element(x):
 
     """
 
+    if isinstance(x, dpt.usm_ndarray):
+        return x
     if isinstance(x, dpnp_array):
         return x.get_array()
     if isinstance(x, range):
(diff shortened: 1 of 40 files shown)


dpnp/dpnp_array.py
```diff
@@ -71,6 +72,16 @@ def _unwrap_index_element(x):
         if arr.size == 0:
             arr = arr.astype(numpy.intp)
         return arr
+    if isinstance(x, numpy.ndarray):
+        return x
+    # convert buffer protocol objects (array.array, memoryview, etc.)
+    try:
```
ndgrigorian: seems `__buffer__` has been added with 3.12: https://docs.python.org/3/reference/datamodel.html#object.__buffer__
https://docs.python.org/3/library/collections.abc.html#collections.abc.Buffer
https://peps.python.org/pep-0688/

we should probably start to support it as well. Maybe add a TODO to check for `__buffer__`/use `collections.abc.Buffer` and then call `memoryview` if it fails

ndgrigorian: actually I misunderstood the PEP: this feature is already implemented, we can add a check for Python version and check against `__buffer__` or try `memoryview` only if it's <3.12

dpnp/dpnp_array.py
```diff
@@ -71,6 +72,16 @@ def _unwrap_index_element(x):
         if arr.size == 0:
             arr = arr.astype(numpy.intp)
         return arr
+    if isinstance(x, numpy.ndarray):
+        return x
+    # convert buffer protocol objects (array.array, memoryview, etc.)
```
antonwolfy: Can all that use cases covered converting `x` to numpy.ndarray?
Do we need so complicated if-logic?

ndgrigorian: here is CuPy implementation of a similar idea for a reference:
https://github.com/cupy/cupy/blob/main/cupy/_core/_routines_indexing.pyx#L247

can take a similar approach too in some cases (best to leave converting to `usm_ndarray` to the tensor layer, though, for queue propagation reasons)

antonwolfy: I don't think we need to fully follow the same CuPy implementation in `_prepare_slice_list`, because it seems CuPy mimics faulty NumPy behavior sometimes.

Not sure if I got the comment fully correctly, but in general I meant something like:
```python
    if x is None or x is Ellipsis or isinstance(x, (dpt.usm_ndarray, slice, numpy.ndarray)):
        return x
    if isinstance(x, dpnp_array):
        return x.get_array()

    try:
        x = numpy.asarray(x)
    except ... :
        ...
        raise ...
    ...
    return x
```

ndgrigorian: yep, agreed mostly, though problem is, we don't really want to send integers or boolean Python scalars to numpy arrays, which is why we may need something like:

```
        elif numpy.isscalar(s):
            if not isinstance(s, (bool, numpy.bool_)):
                # keep scalar int
                continue
```
which is in cupy implementation. We could handle `range`, `list`, `tuple` through numpy though

vlad-perevezentsev: I don not think `try/except` around `numpy.asarray` is necessary because invalid index types either raise naturally or produce arrays with unsupported dtypes that the tensor layer rejects with `IndexError`

Regarding `elif numpy.isscalar(s)` I replaced it with `isinstance(x, (int, numpy.generic))` because `np.isscalar(memoryview(array.array("l", [0, 1, 2])))` returns `True` 
`bool` check is not needed since bool is subclass of `int`

dpnp/dpnp_array.py
```diff
@@ -52,26 +52,31 @@ def _unwrap_index_element(x):
...
-    return x
+    # scalars (int, bool, numpy scalars) pass through to the tensor layer
+    if isinstance(x, (int, numpy.generic)):
+        return x
+
+    # convert array-like objects (range, list, buffer protocol) to numpy
+    arr = numpy.asarray(x)
+    # cast empty arrays (float64 in NumPy) to intp
+    # for correct tensor indexing
+    if arr.size == 0 and arr.dtype.kind == "f":
```
antonwolfy: Why do we need to explicitly check `arr.dtype.kind == "f"`?

vlad-perevezentsev: This should be removed because it is a leftover from the previous logic
thanks

dpnp/dpnp_array.py
```diff
@@ -52,26 +52,31 @@ def _unwrap_index_element(x):
     """
     Unwrap a single index element for the tensor indexing layer.
 
-    Converts dpnp arrays to usm_ndarray and array-like objects (range, list)
-    to numpy arrays with intp dtype for NumPy-compatible advanced indexing.
+    Converts dpnp arrays to usm_ndarray and array-like objects (range, list,
```
antonwolfy, line 55: Need to update docstring of `_get_unwrapped_index_key` as well, missing mentioning of buffer protocol.

antonwolfy (approved):
LGTM with small nit below

dpnp/dpnp_array.py
vlad-perevezentsev: done
thank you!

antonwolfy (approved):

PR #2901, opened 2026-05-06
task: add broadcast class implementation

jharlow-intel:
Adds a `broadcast` class implementation 


dpnp/dpnp_broadcast.py
```diff
@@ -0,0 +1,170 @@
...
+        # Convert all arguments to dpnp arrays
+        arrays = []
+        for arg in args:
+            if not isinstance(arg, dpnp.ndarray):
+                # Convert array-like to dpnp.ndarray
+                arg = dpnp.asarray(arg)
+            arrays.append(arg)
+
+        if len(arrays) == 0:
+            raise TypeError("broadcast() requires at least one array")
```
antonwolfy: What is the reason to raise the error here? I see that numpy accept that.

dpnp/dpnp_broadcast.py
```diff
@@ -0,0 +1,170 @@
...
+    Iterator functionality is not supported.
+
+    """
+
+    def __init__(self, *args):
+        # Convert all arguments to dpnp arrays
+        arrays = []
+        for arg in args:
+            if not isinstance(arg, dpnp.ndarray):
+                # Convert array-like to dpnp.ndarray
```
antonwolfy: Conversion might be a costly operation, why do we need that?
Also, we have to allocate new arrays only on the same SYCL queue and with coerced USM type to comply with Compute follows data.

antonwolfy: We will probably need a conversion logic only when iterator is implemented

antonwolfy: Btw, we need to check that input dpnp arrays are on the same queue, something like:
```python
exec_q = dpt.get_execution_queue(...)
if exec_q is None:
    raise ExecutionPlacementError()
```

antonwolfy: And also we need to check that every arg in args has `.shape` attribute

dpnp/dpnp_broadcast.py
jharlow-intel: Fixed!

dpnp/dpnp_broadcast.py
jharlow-intel: All implemented

dpnp/dpnp_broadcast.py
```diff
@@ -0,0 +1,183 @@
...
+        dpnp_arrays = [arg for arg in self._arrays if isinstance(arg, dpnp.ndarray)]
+        if len(dpnp_arrays) > 1:
+            exec_q = dpt.get_execution_queue(
+                tuple(array.sycl_queue for array in dpnp_arrays)
+            )
+            if exec_q is None:
+                raise dpt.ExecutionPlacementError(
+                    "Execution placement can not be unambiguously inferred "
+                    "from input arguments."
+                )
```
ndgrigorian: do we need to check for compute follows data here? The arrays can be broadcast even if they aren't, it just means device routines can't be run with both as inputs

ndgrigorian: if we do keep this check, we would need to check if they are `dpnp.tensor.usm_ndarray` as well

dpnp/dpnp_broadcast.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        Number of iterators possessed by the broadcast object.
+
+        Returns
+        -------
+        out : int
+            The number of iterators.
+
+        """
+        return len(self._arrays)
```
ndgrigorian: primarily a question for @antonwolfy and @vlad-perevezentsev since it's design-related, but NumPy and CuPy differ drastically in this class implementation

https://numpy.org/doc/2.1/reference/generated/numpy.broadcast.html
https://docs.cupy.dev/en/latest/reference/generated/cupy.broadcast.html

do we want more of the CuPy or NumPy behavior? What is the intended use-case of this class to users?

dpnp/dpnp_broadcast.py
```diff
@@ -0,0 +1,171 @@
...
+            self._nd = 0
+            return
+
+        # Compute the broadcasted shape using _broadcast_shapes
+        self._shape = _broadcast_shapes(*self._arrays)
+
+        # Calculate size and ndim
+        self._size = 1
+        for dim in self._shape:
+            self._size *= dim
```
antonwolfy: we can use `math.prod`:

```suggestion
        self._size = math.prod(self._shape) 
```

dpnp/dpnp_broadcast.py
antonwolfy: I'd say we are not going to fully align with NumPy here, aligning with CuPy looks more preferable as for me, because might help in case of CuPy to DPNP migration for some users.

In that perspective, I'd keep the current implement as it is for now, plus adding `values` property, which mimics CuPy.

dpnp/dpnp_broadcast.py
```diff
@@ -0,0 +1,171 @@
...
+    Returns
+    -------
+    broadcast : broadcast object
+        Broadcast the input parameters against one another, and
+        return an object that encapsulates the result.
+        Amongst others, it has ``shape`` and ``nd`` properties.
+
+    See Also
+    --------
+    :obj:`dpnp.broadcast_arrays` : Broadcast any number of arrays against
```
antonwolfy: Missing back-lins on `dpnp.broadcast` in docstrings of all three functions below.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 This release is compatible with NumPy 2.4.5.
 
 ### Added
+* Added `dpnp.broadcast` class implementation [#2901](https://github.com/IntelPython/dpnp/pull/2901)
```
antonwolfy: Remove blank empty line:
```suggestion
* Added `dpnp.broadcast` class implementation [#2901](https://github.com/IntelPython/dpnp/pull/2901)
```

dpnp/dpnp_broadcast.py
```diff
@@ -0,0 +1,171 @@
...
+    >>> y = np.array([4, 5, 6])
+    >>> b = np.broadcast(x, y)
+    >>> b.shape
+    (3, 3)
+    >>> b.nd
+    2
+    >>> b.size
+    9
+
+    Limitations
```
antonwolfy: `Limitations` and `Notes` should be before `Examples`

dpnp/dpnp_broadcast.py
```diff
@@ -0,0 +1,171 @@
...
+        # Compute the broadcasted shape using _broadcast_shapes
+        self._shape = _broadcast_shapes(*self._arrays)
+
+        # Calculate size and ndim
+        self._size = 1
+        for dim in self._shape:
+            self._size *= dim
+        self._nd = len(self._shape)
+
+    @property
```
antonwolfy: We probably need to add members:
```rst
   broadcast.shape
   broadcast.size
   broadcast.nd
   broadcast.ndim
   broadcast.numiter
```
to `array-manipulation.rst` to make them rendered

dpnp/dpnp_broadcast.py
```diff
@@ -0,0 +1,171 @@
...
+# THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+
+"""Implementation of broadcast class."""
+
+import dpnp
+from dpnp.tensor._manipulation_functions import _broadcast_shapes
+
+
+class broadcast:
```
antonwolfy: I don't think we need a new file `dpnp/dpnp_broadcast.py`. The class can be a part of existing `dpnp/dpnp_iface_manipulation.py` which currently holds all the broadcast functions

dpnp/dpnp_broadcast.py
```diff
@@ -0,0 +1,171 @@
...
+        Returns
+        -------
+        out : int
+            The total size (number of elements) of the broadcasted result.
+
+        """
+        return self._size
+
+    @property
+    def nd(self):
```
antonwolfy: I don't think we need `nd` method, because NumPy states that `ndim` is preferable:
> Number of dimensions of broadcasted result. For code intended for NumPy 1.12.0 and later the more consistent [ndim](https://numpy.org/doc/stable/reference/generated/numpy.ndim.html#numpy.ndim) is preferred.

dpnp/dpnp_broadcast.py
```diff
@@ -0,0 +1,171 @@
...
+    @property
+    def shape(self):
+        """
+        Shape of the broadcasted result.
+
+        Returns
+        -------
+        out : tuple
+            A tuple containing the shape of the broadcasted result.
+
```
antonwolfy: Missing `Examples` sections for the methods

antonwolfy (approved):
Hope it will be rendered properly now.
Thank you @jharlow-intel for the main implementation.

PR #2903, opened 2026-05-07
Change order for step with installing OCL CPU RT

antonwolfy:
The PR fixes the issue in GitHub workflow to scheduled run of the tests.

We have to install proper `intel-opencl-rt` package first, since otherwise installing `intel-opencl-rt` next step after installing dpnp might cause reinstall of older dpnp version, since we don't use strict conda channel priority.


antonwolfy:
The tests passed manually:
https://github.com/IntelPython/dpnp/actions/runs/25520363258/job/74902469656

ndgrigorian (approved):

PR #2905, opened 2026-05-07
Switch to conda-forge `ocl-icd` loader packages

antonwolfy:
Switches the OpenCL ICD loader run dependency from `intel-gpu-ocl-icd-system` to the conda-forge loaders when building against a DPC++ compiler and runtime of `2026.1.1` or newer:

- **Linux:** `ocl-icd-system`
- **Windows:** `khronos-opencl-icd-loader`

The conda-forge loaders require compiler/runtime `2026.1.1`+, so the choice is gated on the compiler version the package is built against. When building against an older compiler, the recipe keeps the legacy `intel-gpu-ocl-icd-system` package.

The `-system` variant on Linux is intentional: it symlinks the system-installed OpenCL ICD loader into the environment rather than bundling its own, so the Intel GPU driver's registered ICD is discovered correctly. Using the bundling `ocl-icd` package instead shadows the system loader and breaks device enumeration.

## Decoupled compiler and oneMKL floors

The single `required_compiler_and_mkl_version` constant is split into two independent minimums, because the DPC++ compiler and oneMKL do not release in lockstep: the compiler has a `2026.1.1` build on the public channels, but `mkl-devel-dpcpp` only goes up to `2026.1.0`. Keeping them coupled made a `2026.1.1` floor unsolvable.


conda-recipe/meta.yaml
```diff
@@ -51,7 +51,8 @@ requirements:
       - {{ pin_compatible('onemkl-sycl-rng', min_pin='x.x', max_pin='x') }}
       - {{ pin_compatible('onemkl-sycl-vm', min_pin='x.x', max_pin='x') }}
       - numpy
-      - intel-gpu-ocl-icd-system
+      - ocl-icd  # [linux]
```
ndgrigorian:
```suggestion
      - ocl-icd-system  # [linux]
```
shouldn't it be this? I couldn't get it to work without this

antonwolfy: we can try

conda-recipe/meta.yaml
```diff
@@ -1,7 +1,25 @@
...
+{# The minimum compiler version must stay strictly below the (exclusive) maximum, #}
+{# otherwise the ">=min,<max" compiler pin collapses to an unsolvable empty range. #}
+{% if required_compiler_version.split('.') | map('int') | list >= max_compiler_and_mkl_version.split('.') | map('int') | list %}
+{{ MIN_BUILD_CMPL_VERSION_must_be_below_MAX_BUILD_CMPL_MKL_VERSION() }}
+{% endif %}
+
+{# The conda-forge OpenCL ICD loaders (ocl-icd-system / khronos-opencl-icd-loader) #}
+{# require a compiler and runtime of 2026.1.1 or newer. Fall back to the legacy #}
+{# intel-gpu-ocl-icd-system package when building against an older compiler. #}
+{% set use_conda_forge_ocl_icd = required_compiler_version.split('.') | map('int') | list >= [2026, 1, 1] %}
```
ndgrigorian, line 22: Why do they require this? Didn't we build successfully prior to 2026.1?

antonwolfy: With `ocl-icd-system` / `khronos-opencl-icd-loader`? No, it caused clobber issue.

antonwolfy: [Here](https://cje-fm-owrp-prod04.devtools.intel.com/satg-dap-intelpython/job/intel-packages/job/dpnp/job/dev-linux-py3.14/job/test/1171/cloudbees-pipeline-explorer/) is an example where 2026.2 lkg packages does not include the fix from 2026.1.1 release yet.

ndgrigorian: That was due to Intel channel packages though, not conda-forge

ndgrigorian: the conda-forge builds had no clobber issues for me locally even prior, the headers that were part of intel-sycl-rt and intel-opencl-rt were only present on the Intel channel packages and not conda-forge packages. With conda-forge, it should be stable even prior

antonwolfy: Yep, but we still have to support the way when building dpnp with Intel channel. And that change (migration to ocl-icd from conda-forge) will work with compiler >= 2026.1.1 only.
While the build on conda-forge feedstock is always done with the latest published compiler, i.e. 2026.1.1 (hope the repack will be done soon), and so should be no issue.

ndgrigorian: Agreed, I think it is just worth noting in the comments, that conda-forge packages would work regardless, for the sake of users who may be building manually

ndgrigorian (approved):
LGTM aside from above point about updating comments

antonwolfy:
The PR will be merged, once LKG on Linux for DPC++ compiler includes fixes from 2026.1.1 release.
Otherwise the internal Jenkins workflows will be failing.

PR #2913, opened 2026-05-12
Convert .pxi include files to .pxd/.pyx in dpnp/tensor

vlad-perevezentsev:
This PR proposes to replace Cython textual includes with proper module imports in `dpnp/tensor`
.pxi files are converted into corresponding .pxd and .pyx modules for _slicing.pxi, _stride_utils.pxi and _types.pxi`

This change replaces the include `.pxi` pattern with standard Cython cimport statements and improving modularity


dpnp/tensor/_usmarray.pyx
```diff
@@ -56,9 +55,55 @@ from . import _flags
 from ._dlpack import get_build_dlpack_version
 from ._tensor_impl import default_device_fp_type
 
-include "_stride_utils.pxi"
-include "_types.pxi"
-include "_slicing.pxi"
+from ._slicing cimport *
```
antonwolfy: Is it a good practice to use star cimport?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Replaced `.pxi` includes in `dpnp.tensor` with modular `.pxd`/`.pyx` Cython imports[#2913](https://github.com/IntelPython/dpnp/pull/2913)
```
antonwolfy:
```suggestion
* Replaced `.pxi` includes in `dpnp.tensor` with modular `.pxd`/`.pyx` Cython imports [#2913](https://github.com/IntelPython/dpnp/pull/2913)
```

dpnp/tensor/_slicing.pyx
```diff
@@ -26,11 +26,16 @@
...
 
+# distutils: language = c++
+# cython: language_level=3
+
 import numbers
 from operator import index
 from cpython.buffer cimport PyObject_CheckBuffer
 from numpy import ndarray
 
+from ._usmarray cimport usm_ndarray
```
antonwolfy, line 36: Does not that produce a circular dependency? (_usmarray ↔ _slicing)

dpnp/tensor/_usmarray.pyx
```diff
@@ -56,9 +55,55 @@ from . import _flags
...
+    dtype_to_typenum,
+    type_bytesize,
+    typenum_from_format,
+)
+
+from ._slicing import _basic_slice_meta
+
+
+# Local storage for `cdef public api` constants
+# declared in _usmarray.pxd
```
antonwolfy: Is there a way to avoid redefinition? To keep the constant defined in single place, the same is applicable to UAR_* constants.

ndgrigorian: when these files were in dpctl, my thought was to move these constants into another, separate `.pxd` which could just be included by both

vlad-perevezentsev: @ndgrigorian could you clarify what you mean by a separate `.pxd` file?
Previously these constants were defined in `.pxi` files that were included directly into `_usmarray.pyx` which provided local storage for `cdef public api`
`cimport `from another `.pxd` does not work that (header is generated but constants are zero at runtime)

ndgrigorian: > @ndgrigorian could you clarify what you mean by a separate `.pxd` file? Previously these constants were defined in `.pxi` files that were included directly into `_usmarray.pyx` which provided local storage for `cdef public api` `cimport `from another `.pxd` does not work that (header is generated but constants are zero at runtime)

I see, so reason for this is subtle
https://github.com/cython/cython/issues/4369
here is Cython issue which discusses it, found by googling this issue.

The best approach, in this case, is probably to separate the constants to a .h header file which can then be cdef extern from, and then that can also be used in dpnp4pybind11 to set the constants for the CAPI

ndgrigorian: we could also use the enum approach, if preferred, but I think the header approach is cleaner

dpnp/tensor/libtensor/include/kernels/elementwise_functions/expm1.hpp
```diff
@@ -121,6 +121,10 @@ struct Expm1Functor
                 }
             }
 
+            if (x == realT(0) && y == realT(0)) {
```
ndgrigorian, line 124: seems like this change is meant for a separate PR

dpnp/tests/tensor/elementwise/test_expm1.py
```diff
@@ -185,3 +185,31 @@ def test_expm1_special_cases():
     tol = dpt.finfo(X.dtype).resolution
     with np.errstate(invalid="ignore"):
         assert_allclose(dpt.asnumpy(dpt.expm1(X)), res, atol=tol, rtol=tol)
+
+
+@pytest.mark.parametrize("dtype", ["c8", "c16"])
```
ndgrigorian: same here (meant for different PR)

dpnp/backend/include/dpnp4pybind11.hpp
```diff
@@ -32,9 +32,10 @@
 #include "dpctl4pybind11.hpp"
 
 // Include generated Cython headers for usm_ndarray
-// (struct definition and constants only)
 #include "dpnp/tensor/_usmarray.h"
 #include "dpnp/tensor/_usmarray_api.h"
+// Include usm_ndarray constants (flags, type numbers)
+#include "../../tensor/include/usm_ndarray_constants.h"
```
ndgrigorian, line 38: why a relative import here, when above, we have "dpnp/tensor/_usmarray.h"? Shouldn't "dpnp/tensor/include/usm_ndarray_constants.h" work?

dpnp/tensor/_usmarray.pyx
vlad-perevezentsev: fixed
thanks

dpnp/tensor/libtensor/include/kernels/elementwise_functions/expm1.hpp
vlad-perevezentsev: Removed
thank you!

dpnp/tests/tensor/elementwise/test_expm1.py
vlad-perevezentsev: Removed
thank you!

dpnp/backend/include/dpnp4pybind11.hpp
vlad-perevezentsev: @ndgrigorian it does not work because `dpnp/tensor/include/` is not on the include path so `_tensor_impl` and backend extensions fail with
```
dpnp/backend/include/dpnp4pybind11.hpp:38:10: fatal error: 'dpnp/tensor/include/usm_ndarray_constants.h' file not found
   38 | #include "dpnp/tensor/include/usm_ndarray_constants.h"
```

dpnp/tensor/_usmarray.h works because it is generated into ${CMAKE_BINARY_DIR} which is already on the include path.

The alternatives were either adding `dpnp/tensor/include` to all `target_include_directories` (which would also expose unnecessary dlpack headers) or copying the file into ${CMAKE_BINARY_DIR}. 
So I chose the relative include path as the simplest solution

ndgrigorian: makes sense

dpnp/tensor/_slicing.pyx
vlad-perevezentsev: `_slicing.pxd` itself does not depend on `_usmarray;`
It is only needed for `isinstance()` checks in  .pyx implementation so this circular cimport is safe

dpnp/tensor/_types.pyx
```diff
@@ -26,25 +26,30 @@
...
+cdef int UAR_LONG = UAR_LONG_VALUE
+cdef int UAR_ULONG = UAR_ULONG_VALUE
+cdef int UAR_LONGLONG = UAR_LONGLONG_VALUE
+cdef int UAR_ULONGLONG = UAR_ULONGLONG_VALUE
+cdef int UAR_FLOAT = UAR_FLOAT_VALUE
+cdef int UAR_DOUBLE = UAR_DOUBLE_VALUE
+cdef int UAR_CFLOAT = UAR_CFLOAT_VALUE
+cdef int UAR_CDOUBLE = UAR_CDOUBLE_VALUE
+cdef int UAR_TYPE_SENTINEL = UAR_TYPE_SENTINEL_VALUE
+cdef int UAR_HALF = UAR_HALF_VALUE
```
ndgrigorian: seems redundant to cdef them here and in usmarray.pyx, is there a way to do it in something common?

vlad-perevezentsev: @ndgrigorian 
the lines like  `cdef int = *_VALUE` in `_usmarray.pyx` are required to provide local storage for `cdef public api` variable declared in `_usmarray.pxd` because without it `__pyx_capi__` capsule will not have a corresponding value

I have removed the redundant constant declarations fin `_types.pxd/.pyx` 
In `_stride_utils.pxd/.pyx` only keeps `USM_ARRAY_C_CONTIGUOUS` and `USM_ARRAY_F_CONTIGUOUS` which are used internally

ndgrigorian (approved):
no more comments from my side, LGTM

PR #2914, opened 2026-05-13
Fix `dpnp.tensor.acosh` handling of complex(+-0, NaN)

vlad-perevezentsev:
This PR proposes to fix issue #2880  where `dpnp.tensor.acosh(complex(0, NaN))` returned `NaN + NaN j` instead of `NaN ± π/2 j` as required by the Array API specification

The issue was caused by an incorrect early-return branch in acosh.hpp:
```
if (std::isnan(ry)) {
    return resT{ry, ry};
}
```
This special case bypassed the general logic which already produces the correct result.

The fix removes this unnecessary branch making the behavior consistent with both the Array API specification and NumPy
```
In [1]: import dpnp.tensor as dpt

In [2]: import numpy as np

In [3]: a = dpt.asarray([complex(+0, np.nan)])

In [4]: dpt.acosh(a)
Out[4]: usm_ndarray([nan+1.57079633j])

In [5]: a = np.asarray([complex(+0, np.nan)])

In [6]: np.acosh(a)
Out[6]: array([nan+1.57079633j])
```

Also adds a new `test_acosh_zero_nan` test to cover this case

### DIFF
--- a/dpnp/tests/tensor/elementwise/test_hyperbolic.py
+++ b/dpnp/tests/tensor/elementwise/test_hyperbolic.py
@@ -38,7 +38,9 @@ from ..helper import (
 )
 from .utils import (
     _all_dtypes,
+    _complex_fp_dtypes,
     _map_to_device_dtype,
+    _real_fp_dtypes,
 )
 
 _hyper_funcs = [(np.sinh, dpt.sinh), (np.cosh, dpt.cosh), (np.tanh, dpt.tanh)]
@@ -65,7 +67,7 @@ def test_hyper_out_type(np_call, dpt_call, dtype):
 
 
 @pytest.mark.parametrize("np_call, dpt_call", _all_funcs)
-@pytest.mark.parametrize("dtype", ["f2", "f4", "f8"])
+@pytest.mark.parametrize("dtype", _real_fp_dtypes)
 def test_hyper_real_contig(np_call, dpt_call, dtype):
     q = get_queue_or_skip()
     skip_if_dtype_not_supported(dtype, q)
@@ -96,7 +98,7 @@ def test_hyper_real_contig(np_call, dpt_call, dtype):
 
 
 @pytest.mark.parametrize("np_call, dpt_call", _all_funcs)
-@pytest.mark.parametrize("dtype", ["c8", "c16"])
+@pytest.mark.parametrize("dtype", _complex_fp_dtypes)
 def test_hyper_complex_contig(np_call, dpt_call, dtype):
     q = get_queue_or_skip()
     skip_if_dtype_not_supported(dtype, q)
@@ -123,7 +125,7 @@ def test_hyper_complex_contig(np_call, dpt_call, dtype):
 
 
 @pytest.mark.parametrize("np_call, dpt_call", _all_funcs)
-@pytest.mark.parametrize("dtype", ["f2", "f4", "f8"])
+@pytest.mark.parametrize("dtype", _real_fp_dtypes)
 def test_hyper_real_strided(np_call, dpt_call, dtype):
     q = get_queue_or_skip()
     skip_if_dtype_not_supported(dtype, q)
@@ -157,7 +159,7 @@ def test_hyper_real_strided(np_call, dpt_call, dtype):
 
 
 @pytest.mark.parametrize("np_call, dpt_call", _all_funcs)
-@pytest.mark.parametrize("dtype", ["c8", "c16"])
+@pytest.mark.parametrize("dtype", _complex_fp_dtypes)
 def test_hyper_complex_strided(np_call, dpt_call, dtype):
     q = get_queue_or_skip()
     skip_if_dtype_not_supported(dtype, q)
@@ -185,7 +187,7 @@ def test_hyper_complex_strided(np_call, dpt_call, dtype):
 
 
 @pytest.mark.parametrize("np_call, dpt_call", _all_funcs)
-@pytest.mark.parametrize("dtype", ["f2", "f4", "f8"])
+@pytest.mark.parametrize("dtype", _real_fp_dtypes)
 def test_hyper_real_special_cases(np_call, dpt_call, dtype):
     q = get_queue_or_skip()
     skip_if_dtype_not_supported(dtype, q)
@@ -202,9 +204,9 @@ def test_hyper_real_special_cases(np_call, dpt_call, dtype):
     assert_allclose(dpt.asnumpy(dpt_call(yf)), Y_np, atol=tol, rtol=tol)
 
 
-@pytest.mark.parametrize("dtype", ["c8", "c16"])
+@pytest.mark.parametrize("dtype", _complex_fp_dtypes)
 def test_acosh_zero_nan(dtype):
-    # check acosh(0 + NaN j) = NaN ± πj/2
+    # check acosh(±0 + NaN j) = NaN ± π/2 j (Array API spec)
     q = get_queue_or_skip()
     skip_if_dtype_not_supported(dtype, q)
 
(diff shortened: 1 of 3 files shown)


dpnp/tests/tensor/elementwise/test_hyperbolic.py
```diff
@@ -200,3 +200,24 @@ def test_hyper_real_special_cases(np_call, dpt_call, dtype):
 
     tol = 8 * dpt.finfo(dtype).resolution
     assert_allclose(dpt.asnumpy(dpt_call(yf)), Y_np, atol=tol, rtol=tol)
+
+
+@pytest.mark.parametrize("dtype", ["c8", "c16"])
+def test_acosh_zero_nan(dtype):
```
antonwolfy, line 208: new test is failing

dpnp/tests/tensor/elementwise/test_hyperbolic.py
```diff
@@ -200,3 +200,24 @@ def test_hyper_real_special_cases(np_call, dpt_call, dtype):
 
     tol = 8 * dpt.finfo(dtype).resolution
     assert_allclose(dpt.asnumpy(dpt_call(yf)), Y_np, atol=tol, rtol=tol)
+
+
+@pytest.mark.parametrize("dtype", ["c8", "c16"])
```
antonwolfy: Please use `_complex_fp_dtypes` here

dpnp/tests/tensor/elementwise/test_hyperbolic.py
```diff
@@ -200,3 +200,24 @@ def test_hyper_real_special_cases(np_call, dpt_call, dtype):
...
+def test_acosh_zero_nan(dtype):
+    # check acosh(0 + NaN j) = NaN ± πj/2
```
antonwolfy: I guess to be consistent with left part:
```suggestion
    # check acosh(0 + NaN j) = NaN ± π/2 j
```

dpnp/tests/tensor/elementwise/test_hyperbolic.py
```diff
@@ -200,3 +200,24 @@ def test_hyper_real_special_cases(np_call, dpt_call, dtype):
...
+    q = get_queue_or_skip()
+    skip_if_dtype_not_supported(dtype, q)
+
+    x = [complex(+0.0, np.nan), complex(-0.0, np.nan)]
+
+    xf = np.array(x, dtype=dtype)
+    yf = dpt.asarray(xf, dtype=dtype, sycl_queue=q)
+
+    with np.errstate(all="ignore"):
+        Y_np = np.arccosh(xf)
```
antonwolfy: NumPy also has `acosh` alias:
```suggestion
        Y_np = np.acosh(xf)
```

dpnp/tests/tensor/elementwise/test_hyperbolic.py
```diff
@@ -200,3 +200,24 @@ def test_hyper_real_special_cases(np_call, dpt_call, dtype):
...
+
+    Y_dpt = dpt.asnumpy(dpt.acosh(yf))
+
+    for i in range(len(x)):
```
antonwolfy: we need to place a comment that NumPy returns wrong result: NaN + NaN j
and probably to add a link to the issue in NumPy repo

dpnp/tests/tensor/elementwise/test_hyperbolic.py
```diff
@@ -200,3 +200,24 @@ def test_hyper_real_special_cases(np_call, dpt_call, dtype):
...
+        assert np.isnan(Y_np[i].real) == np.isnan(Y_dpt[i].real)
```
antonwolfy: why do we need a loop? isn't the same as like:
```python
assert np.isnan(Y_np.real) == np.isnan(Y_dpt.real)
```

dpnp/tests/tensor/elementwise/test_hyperbolic.py
ndgrigorian: conda-forge numpy gives the `nan, nan` result, while wheels give the `0, nan` result, because of something to do with numpy dispatching

https://github.com/numpy/numpy/issues/29167#issuecomment-2968099392

probably, it would be best to just not use numpy for testing these kinds of edge case results

vlad-perevezentsev: Yes, I have noticed that too
It is better not to compare it to NumPy at all
Thank you!

dpnp/tests/tensor/elementwise/test_hyperbolic.py
vlad-perevezentsev: Done
Thank you!

dpnp/tests/tensor/elementwise/test_hyperbolic.py
vlad-perevezentsev: I have updated this test

dpnp/tests/tensor/elementwise/test_hyperbolic.py
antonwolfy: But both cases seem invalid, we expect `NaN ± π/2 j` based on the spec. Should we report then?

ndgrigorian: > But both cases seem invalid, we expect `NaN ± π/2 j` based on the spec. Should we report then?

PyPI isn't incorrect
```
In [1]: import numpy as np

In [2]: vals = [complex(np.nan, -np.inf), complex(-np.inf, np.nan), complex(0, np.nan)]

In [3]: x = np.array(vals)

In [4]: np.arcsinh(x[0]), np.arcsin(x[1]), np.arccosh(x[2])
Out[4]:
(np.complex128(inf+nanj),
 np.complex128(nan+infj),
 np.complex128(nan+1.5707963267948966j))
```
the last case is `NaN + pi/2`, it's only conda-forge which is incorrect.

I don't think we shouldn't file, just noting that, they are transiently aware of it and it's platform-dependent for NumPy failure

dpnp/tests/tensor/elementwise/test_hyperbolic.py
```diff
@@ -200,3 +202,21 @@ def test_hyper_real_special_cases(np_call, dpt_call, dtype):
...
+    x = [complex(+0.0, np.nan), complex(-0.0, np.nan)]
+
+    xf = np.array(x, dtype=dtype)
+    yf = dpt.asarray(xf, dtype=dtype, sycl_queue=q)
+
+    Y_dpt = dpt.asnumpy(dpt.acosh(yf))
+
+    pi_half = np.full(len(x), np.pi / 2, dtype=xf.real.dtype)
+    assert np.all(np.isnan(Y_dpt.real))
+    assert_allclose(abs(Y_dpt.imag), pi_half, atol=1e-6)
```
antonwolfy: I guess it might be simplified:
```suggestion
    assert np.isnan(Y_dpt.real).all()
    assert_allclose(np.abs(Y_dpt.imag), np.pi / 2, atol=1e-6)
```

dpnp/tests/tensor/elementwise/test_hyperbolic.py
antonwolfy: I see, I thought the same faulty result for PyPI.

antonwolfy (approved):
The only minor nit in the tests. The remaining is absolutely LGTM!
Thank you @vlad-perevezentsev

dpnp/tests/tensor/elementwise/test_hyperbolic.py
vlad-perevezentsev: `assert_allclose` in dpnp uses `strict=True` thich requires matching shapes so I use `pi_half` as an array
```
E       AssertionError: 
E       Not equal to tolerance rtol=1e-07, atol=1e-06
E       
E       (shapes (2,), () mismatch)
E        ACTUAL: array([1.570796, 1.570796])
E        DESIRED: array(1.570796)

dpnp/tests/testing/array.py:75: AssertionError
```

vlad-perevezentsev: I have set `strict=False` for that

antonwolfy (approved):

PR #2915, opened 2026-05-15
Include libtensor headers in dpnp package distribution

### LINKED ISSUE #2909: Install libtensor when installing dpnp
dpnp conda package does not install libtensor, making writing extensions using libtensor impossible

vlad-perevezentsev:
This PR fixes missing libtensor headers in the installed dpnp package #2909 

During the migration from `dpctl.tensor` to `dpnp.tensor` libtensor header files were not added to `package_data` so they were not included in the final package.

This PR adds the missing libtensor include paths to `setup.py`


ndgrigorian:
@vlad-perevezentsev 
do you think it's possible to add a test for this? We have a test for C extension in dpctl
https://github.com/IntelPython/dpctl/blob/master/dpctl/tests/test_headers.py
https://github.com/IntelPython/dpctl/blob/master/dpctl/tests/setup_c_ext.py
https://github.com/IntelPython/dpctl/blob/master/dpctl/tests/_c_ext.c

in theory, I think could do something similar with pybind11 here, though I guess it would add pybind11 as a test dependency, unless I'm missing something.

But we could also just port in some examples

ndgrigorian:
We also should add here
https://github.com/IntelPython/dpnp/blob/0599239194fbe345d8cdd17715808e2b6079dba6/dpnp/__main__.py#L55
`--includes` and `--includes-dir`, which lead to the directory with `dpnp4pybind11.hpp`

vlad-perevezentsev:
> @vlad-perevezentsev do you think it's possible to add a test for this? We have a test for C extension in dpctl https://github.com/IntelPython/dpctl/blob/master/dpctl/tests/test_headers.py https://github.com/IntelPython/dpctl/blob/master/dpctl/tests/setup_c_ext.py https://github.com/IntelPython/dpctl/blob/master/dpctl/tests/_c_ext.c
> 
> in theory, I think could do something similar with pybind11 here, though I guess it would add pybind11 as a test dependency, unless I'm missing something.
> 
> But we could also just port in some examples

This is possible via `pybind11` dependency for tests
I can try to implement this in a future PR

vlad-perevezentsev:
> We also should add here
> 
> https://github.com/IntelPython/dpnp/blob/0599239194fbe345d8cdd17715808e2b6079dba6/dpnp/__main__.py#L55
> 
> 
> `--includes` and `--includes-dir`, which lead to the directory with `dpnp4pybind11.hpp`

This will be added in the following PR #2916

setup.py
```diff
@@ -52,6 +52,11 @@
             "libdpnp_backend_c.so",
             "dpnp_backend_c.lib",
             "dpnp_backend_c.dll",
+            "tensor/libtensor/include/kernels/*.h*",
+            "tensor/libtensor/include/kernels/elementwise_functions/*.h*",
```
antonwolfy: Should we simplify something like: `tensor/libtensor/include/kernels/*/*.h*` ?

antonwolfy (approved):

setup.py
vlad-perevezentsev: done
thank you!

setup.py
```diff
@@ -52,6 +52,8 @@
             "libdpnp_backend_c.so",
             "dpnp_backend_c.lib",
             "dpnp_backend_c.dll",
+            "tensor/libtensor/include/kernels/*.h*",
+            "tensor/libtensor/include/kernels/*/*.h*",
```
antonwolfy, line 56: I believe previously there was also `tensor/libtensor/include/utils/*.h*`. Isn't  that needed still?

vlad-perevezentsev: This is still needed
It was deleted by mistake
Fixed, thank you!

PR #2921, opened 2026-05-18
Use `ubuntu-latest` runner where possible

antonwolfy:
This PR updates existing GitHub workflows to switch to use `ubuntu-latest` runner instead of `ubuntu-22.04` one where it's possible.

Note, `cron-run-tests.yaml` still assumes to run tests on both `ubuntu-22.04` and `ubuntu-24.04` to validate both deployment paths.
And `conda-package.yml` still builds on `ubuntu-22.04` to support running tests on any Ubuntu since 22.04.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2922, opened 2026-05-18
Use python 3.14 environment in GitHub workflows

antonwolfy:
This PR updates the remaining GitHub workflows and conda environments to run under python 3.14, and bumps coveralls to 4.1.0 (which supports 3.14).

### Version bumps (3.13/3.12 → 3.14)
- `check-onemath.yaml` — oneMath interface tests
- `conda-package.yml` — array API conformance tests (`python-ver`)
- `environments/building_docs.yml` — docs build
- `environments/upload_cleanup_conda_pkg.yml` — conda package upload/cleanup
- `environments/coverage.yml` — coverage job (was pinned to `3.12`)

### Coverage job fixes for python 3.14

Bumping the coverage environment from 3.12 to 3.14 surfaced two failures unrelated to dpnp's runtime code — both stem from how coverage instrumentation interacts with 3.14's `sys.monitoring`:

**1. Build failure — `undeclared identifier '__Pyx_MonitoringEventTypes_CyGen_count'`**

Coverage builds compile every extension with `CYTHON_TRACE=1` (`DPNP_GENERATE_COVERAGE=ON`), but Cython line tracing also needs the matching `# cython: linetrace=True` directive at cythonize time. Three `.pyx` files were missing it (`_slicing.pyx`, `_stride_utils.pyx`, `_types.pyx`): without the directive, Cython references `__Pyx_MonitoringEventTypes_CyGen_count` (from the shared generator/coroutine struct) but never defines it. On python ≤3.12 the guarding `CYTHON_USE_SYS_MONITORING` macro is `0`, so the code was unreachable — which is why the previous `3.12` pin hid this; on 3.13+ it becomes reachable and fails to compile. Added the directive to the three files. As a side effect, these files were previously producing no line-coverage data and are now measured.

**2. Test failures — `test_dlpack.py::TestDLPack::test_invaid_stream[scalar|dictionary|device]`**

coverage.py 7.9+ makes the `sys.monitoring` core (`sysmon`) the default on python 3.14. That core (a) does not support plugins such as the `Cython.Coverage` plugin used here, and (b) corrupts the pending exception when a C/Cython function raises — turning dpctl's correct `TypeError` from `usm_ndarray.__dlpack__` into a `SystemError`, so `assert_raises(TypeError, ...)` fails. Pinned the measurement core back to `ctrace` via `core = "ctrace"` in `[tool.coverage.run]` (`pyproject.toml`). `ctrace` supports both the Cython plugin and branch coverage, so no coverage data is lost.

### DIFF
--- a/.github/workflows/check-onemath.yaml
+++ b/.github/workflows/check-onemath.yaml
@@ -70,7 +70,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.13'] # no dpctl package on PyPI with enabled python 3.14 support
+        python: ['3.14']
         os: [ubuntu-latest] # windows-2022 - no DFT support for Windows in oneMKL
 
     runs-on: ${{ matrix.os }}
@@ -187,7 +187,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.13'] # no dpctl package on PyPI with enabled python 3.14 support
+        python: ['3.14']
         os: [ubuntu-latest] # windows-2022 - no DFT support for Windows in oneMKL
 
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/check-onemath.yaml
+++ b/.github/workflows/check-onemath.yaml
@@ -70,7 +70,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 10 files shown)


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2926, opened 2026-05-19
Fix `dpnp.tensor.expm1` handling of complex(+-0, 0) 

vlad-perevezentsev:
This PR proposes to fix issue #2878 where `dpnp.tensor.expm1(complex(-0.0, 0.0))` returned `-0.0 + 0.0j` on CPU instead of `0.0 + 0.0j` as required by the Array API specification

The fix adds an explicit special-case branch for `complex(±0, 0)` in `expm1.hpp` so CPU and GPU behavior are now consistent and compliant with the specification.

Also `test_expm1_special_cases` test was extended to cover `complex(-0.0, 0.0)`


dpnp/tests/tensor/elementwise/test_expm1.py
```diff
@@ -184,4 +186,11 @@ def test_expm1_special_cases():
 
     tol = dpt.finfo(X.dtype).resolution
     with np.errstate(invalid="ignore"):
-        assert_allclose(dpt.asnumpy(dpt.expm1(X)), res, atol=tol, rtol=tol)
+        Y = dpt.asnumpy(dpt.expm1(X))
+        assert_allclose(Y, res, atol=tol, rtol=tol)
+
+    # assert_allclose treats +0 == -0
```
antonwolfy, line 192: do we need to validate the sign in other cases also when imag=0.0 ?

vlad-perevezentsev: Changed
Thanks

vlad-perevezentsev:
@ndgrigorian this is RFR

ndgrigorian (approved):
LGTM

PR #2928, opened 2026-05-20
Improve handling of large values by inverse hyperbolic and trigonometric functions

### LINKED ISSUE #2924: `acosh` returns wrong result
The following example does not match NumPy and oneMKL:
```python
import dpnp, dpnp.tensor as dpt, numpy as np

dpnp.__version__
# Out: '0.21.0dev0+36.g6b9774b489a'

a = dpt.asarray(complex(-4.45712982e+08, 1.0))
dpt.acosh(a)
# Out: usm_ndarray(inf+0.j)

a = np.asarray(complex(-4.45712982e+08, 1.0))
np.acosh(a)
# Out: np.complex128(20.608332945268156+3.141592651346197j)

a = dpnp.asarray([complex(-4.45712982e+08, 1.0)])

dpnp.acosh(a)  # oneMKL VM is used for 1D input array
# Out: array([20.60833295+3.14159265j])
```

antonwolfy:
The `dpnp.tensor.acosh` function was returning incorrect results (infinity) for complex numbers with large negative real parts (see #2924, #2896).

The issue was caused by the threshold used `1/epsilon` to decide when to switch from acos in SYCL's experimental complex extension (`exprm_ns::acos()`) to a log-based formula. 
The SYCL `exprm_ns::acos()` returns incorrect results for large negative real parts, due to used formula `z + sqrt(z² - 1)`, which leads to `log(0) = -infinity`.

The PR proposes to change the threshold from `1/epsilon` to `sqrt(1/epsilon)/2` for all inverse trig/hyperbolic functions. This is the precision loss point where `sqrt(z² ± 1)` calculations become unstable.

Also the PR reduces code duplication by creating shared `casinh()`, `casin()`, `catanh()`, `catan()` functions in new `complex_math.hpp` header.

### DIFF
--- a/dpnp/tensor/libtensor/include/kernels/elementwise_functions/complex_math.hpp
+++ b/dpnp/tensor/libtensor/include/kernels/elementwise_functions/complex_math.hpp
@@ -181,4 +181,78 @@ T casin(const T &in)
     // reverse result back: swap real and imaginary parts
     return T{std::imag(w), std::real(w)};
 }
+
+template <typename T>
+T catanh(const T &in)
+{
+    using realT = typename T::value_type;
+
+    static constexpr realT pio2 = realT(pi) / realT(2); // PI/2
+
+    const realT x = std::real(in);
+    const realT y = std::imag(in);
+
+    if (std::isnan(x)) {
+        // atanh(NaN + I*+-Inf) = sign(NaN)*0 + I*+-PI/2
+        if (std::isinf(y)) {
+            const realT res_re = sycl::copysign(realT(0), x);
+            const realT res_im = sycl::copysign(pio2, y);
+            return T{res_re, res_im};
+        }
+
+        // all other cases involving NaN return NaN + I*NaN
+        return T{q_nan<realT>, q_nan<realT>};
+    }
+
+    if (std::isnan(y)) {
+        // atanh(+-Inf + I*NaN) = +-0 + I*NaN
+        if (std::isinf(x)) {
+            const realT res_re = sycl::copysign(realT(0), x);
+            return T{res_re, q_nan<realT>};
+        }
+
+        // atanh(+-0 + I*NaN) = +-0 + I*NaN
+        if (x == realT(0)) {
+            return T{x, q_nan<realT>};
+        }
+
+        // all other cases involving NaN return NaN + I*NaN
+        return T{q_nan<realT>, q_nan<realT>};
+    }
+
+    /*
+     * For large x or y including atanh(+-Inf + I*+-Inf) = 0 + I*+-PI/2
+     * The sign of PI/2 depends on the sign of imaginary part of the input.
+     *
+     * exprm_ns::atanh(x) is based on calculating log((1 + x) / (1 - x)) / 2,
+     * so r_eps = 1/eps is appropriate precision loss point.
+     */
+    const realT r_eps = realT(1) / std::numeric_limits<realT>::epsilon();
+    if (sycl::fabs(x) > r_eps || sycl::fabs(y) > r_eps) {
+        const realT res_re = realT(0);
+        const realT res_im = sycl::copysign(pio2, y);
+        return T{res_re, res_im};
+    }
+
+    // ordinary cases
+    return exprm_ns::atanh(exprm_ns::complex<realT>(in));
+}
+
+template <typename T>
+T catan(const T &in)
+{
+    /*
+     * catan(z) = reverse(catanh(reverse(z))),
+     * where reverse(x + I*y) = y + I*x = I*conj(z)
+     */
+
+    // reverse(z): swap real and imaginary parts
+    T reversed{std::imag(in), std::real(in)};
+
+    // compute atanh of reversed input
+    T w = catanh(reversed);
+
+    // reverse result back: swap real and imaginary parts
+    return T{std::imag(w), std::real(w)};
+}
 } // namespace dpnp::tensor::kernels::complex_math
--- a/dpnp/tensor/libtensor/include/kernels/elementwise_functions/complex_math.hpp
+++ b/dpnp/tensor/libtensor/include/kernels/elementwise_functions/complex_math.hpp
@@ -42,8 +42,15 @@
 
 namespace dpnp::tensor::kernels::complex_math
 {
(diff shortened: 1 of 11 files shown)


ndgrigorian:
> The PR proposes to changed the threshold from `1/epsilon` to `sqrt(1/epsilon)/2` for all inverse trig/hyperbolic functions. This is the precision loss point where `sqrt(z² ± 1)` calculations become unstable.

should this be filed as an issue with DPC++ compiler? Since most implementations should have logic to prevent overflow/infs in such cases

antonwolfy:
> should this be filed as an issue with DPC++ compiler? Since most implementations should have logic to prevent overflow/infs in such cases

Yes, we probably should, but I wonder that the [extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_complex.asciidoc) still in the experimental state.
And we already had such branch previously with `1/epsilon` threshold, probably due to the exact the same reason.

I also tried to implement the kernel through SYCL's C complex math functions since marked as supported [extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/C-CXX-StandardLibrary.rst):
```cpp
// Forward declare SYCL's C complex math functions with C linkage
extern "C" {
extern __DPCPP_SYCL_EXTERNAL_LIBC float __complex__ cacosf(float __complex__ z);
extern __DPCPP_SYCL_EXTERNAL_LIBC double __complex__ cacos(double __complex__ z);
}
...
        if constexpr (is_complex<argT>::value) {
            using value_type = typename argT::value_type;

            if constexpr (std::is_same_v<value_type, float>) {
                auto result = ::cacosf(*reinterpret_cast<const float __complex__*>(&in));
                return *reinterpret_cast<const std::complex<float>*>(&result);
            }
            else {
                auto result = ::cacos(*reinterpret_cast<const double __complex__*>(&in));
                return *reinterpret_cast<const std::complex<double>*>(&result);
            }
        }
...
```
but faced exactly the same #2924 issue, since the SYCL does provide `cacos` implementation, but with exactly the naive formula:
```cpp
double __complex__ w = clog(z + csqrt(__sqr(z) - 1.0));
```

Btw I'd propose to merge that PR, in background I'll submit the issue towards the compiler.

dpnp/tensor/libtensor/include/kernels/elementwise_functions/complex_math.hpp
```diff
@@ -0,0 +1,258 @@
+//*****************************************************************************
```
ndgrigorian, line 1: doesn't have to be part of this PR, but should we consider moving in the implementations of other complex transcendental functions, and even others I guess?

i.e., expm1

antonwolfy: > doesn't have to be part of this PR, but should we consider moving in the implementations of other complex transcendental functions, and even others I guess?
> 
> i.e., expm1

Yep, It might help to reduce code duplication.

dpnp/tensor/libtensor/include/kernels/elementwise_functions/complex_math.hpp
```diff
@@ -0,0 +1,258 @@
...
+#include <limits>
+
+#include <sycl/sycl.hpp>
+
+#include "sycl_complex.hpp" // for exprm_ns
+
+namespace dpnp::tensor::kernels::complex_math
+{
+static constexpr double ln2 = 0.6931471805599453094172321214581765L;
+static constexpr double pi = 3.1415926535897932384626433832795029L;
```
ndgrigorian: I am a bit concerned setting these as literals, especially double literals. Casting them into `realT` (i.e., potentially float, from double) in the device code may fail on devices w/o double precision, a similar thing occurred in dpctl some time ago. I think this is generally undefined behavior in the compiler.

It may be best to just calculate pi and ln2 directly.

ndgrigorian: i.e., similar to before, using atan or using acos

dpnp/tensor/libtensor/include/kernels/elementwise_functions/complex_math.hpp
```diff
@@ -0,0 +1,258 @@
...
+        return T{q_nan<realT>, q_nan<realT>};
+    }
+
+    /*
+     * For large x or y including acos(+-Inf + I*+-Inf).
+     * exprm_ns::acos(x) is based on calculating log(x + sqrt(x^2 - 1)),
+     * so r_eps = sqrt(1/eps)/2 is appropriate precision loss point.
+     */
+    const realT r_eps =
+        sycl::sqrt(realT(1) / std::numeric_limits<realT>::epsilon()) / 2;
```
ndgrigorian: could pull the input to the `sqrt` out and mark it constexpr, may give some small optimization

dpnp/tensor/libtensor/include/kernels/elementwise_functions/complex_math.hpp
```diff
@@ -0,0 +1,258 @@
...
+     * asinh(in) = sign(x)*log(sign(x)*in) + O(1/in^2)   as in -> infinity
+     * The above formula works for the imaginary part as well, because
+     * Im(asinh(in)) = sign(x)*atan2(sign(x)*y, fabs(x)) + O(y/in^3)
+     * as in -> infinity, uniformly in y.
+     *
+     * exprm_ns::asinh(x) is based on calculating log(x + sqrt(x^2 + 1)),
+     * so r_eps = sqrt(1/eps)/2 is appropriate precision loss point.
+     */
+    const realT r_eps =
+        sycl::sqrt(realT(1) / std::numeric_limits<realT>::epsilon()) / 2;
```
ndgrigorian: same constexpr point throughout file

dpnp/tensor/libtensor/include/kernels/elementwise_functions/complex_math.hpp
antonwolfy: It makes, I'll update to do `ln2` and `pi` constants as `realT`-typed namespace-scope template variables.
This is a true compile-time constant with zero runtime cost.

ndgrigorian (approved):
The previous comments were addressed

PR #2930, opened 2026-05-22
Resolve gh-2882

ndgrigorian:
This PR adds a fix for gh-2882

When floats were previously converted by `convert_impl` to unsigned integral types, they would pass through an equally sized, signed integral type

If the float was in bounds for the unsigned integral type but out of bounds for the signed type, however, undefined behavior could be encountered.

This PR proposes a solution of casting to a sufficiently large intermediate type first, before casting into the unsigned integral type.


antonwolfy (approved):
Thank you @ndgrigorian, LGTM

PR #2934, opened 2026-05-28
Refresh documentation styling with Furo theme

vlad-perevezentsev:
This PR proposes a refresh of `dpnp documentation` styling and layout by migrating to the `furo` theme and improving the overall API documentation appearance

- Replace `sphinx_rtd_theme` with `furo`
- Update documentation dependencies in `pyproject.toml`, `building_docs.ym`l and `base_build_docs.txt`
- Redesign the landing page using sphinx-design cards
- Add custom CSS/JS for cleaner NumPy-style API formatting
- Improve tables, section headers, parameter formatting and function signatures

dpnp logo (`dpnp.svg`) will be added in the future


doc/_static/dpnp-custom.css
```diff
@@ -0,0 +1,91 @@
+/* Autosummary tables: left-aligned */
```
antonwolfy, line 1: The tab with rendered documentation is currently named as: `dpnpData Parallel ...` - without any space or separator.
It might be better to do as dpctl: `Data Parallel Control (dpctl)`

antonwolfy, line 1: Also it has too long dpnp version:

Image

but we only need 0.21.0dev0

doc/index.rst
```diff
@@ -1,22 +1,87 @@
...
 
+===================================
 Data Parallel Extension for NumPy*
-==================================
+===================================
 
 .. module:: dpnp
     :no-index:
 
+Python package :py:mod:`dpnp` implements a subset of `NumPy*`_ that can be
```
antonwolfy: Would it be better to rephrase something like:
`DPNP is a NumPy-compatible array library that can be executed on any data parallel device. DPNP acts as a drop-in replacement of core NumPy functions and numerical data types.`

I wonder about using the word `subset of NumPy` because it doesn't bring why the user need to use DPNP and what is the subset means in terms of number of numpy functions.

doc/index.rst
```diff
@@ -1,22 +1,87 @@
...
+            :ref-type: ref
+            :expand:
+            :color: secondary
+            :click-parent:
+
+            Access API Reference
+
+    .. grid-item-card:: Tensor (dpnp.tensor)
+
+        The underlying Array API-compliant implementation based on
```
antonwolfy, line 65: Do we need to place a link to dpctl documentaion?

doc/index.rst
```diff
@@ -1,22 +1,87 @@
...
+   :caption: Contents:
 
    overview
    quick_start_guide
    reference/index
    tensor
-
-.. toctree::
-   :maxdepth: 1
-   :caption: Development information
```
antonwolfy, line 20: Do we need to have that on the main page as well with some description?
It currently provides overview for the function from the extensions.

doc/index.rst
antonwolfy: Also we started to add implementation for some SciPy-relating function and so probably need to highlight that as well.

doc/index.rst
vlad-perevezentsev: This card already links to `dpnp.tensor` page where a `Note` provides a link to dpctl documentation
I think that is sufficient

doc/index.rst
vlad-perevezentsev: done

doc/_static/dpnp-custom.css
vlad-perevezentsev: fixed
Thank you!

vlad-perevezentsev: Fixed
Thank you!

doc/index.rst
vlad-perevezentsev: It was deleted by accident
I will return it

doc/_static/dpnp-custom.css
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/* Parameter lists: no bullets, keep indentation */
+dl.field-list dd ul.simple {
+    list-style: none !important;
+    padding-left: 1.2em !important;
+}
+
+/* Dark mode */
+@media (prefers-color-scheme: dark) {
```
antonwolfy: The dark-mode blocks only override table colors. The hardcoded `background-color: #f5f5f5` on `dd p.rubric` / `.admonition-title` (line 40) is not overridden in dark mode → light text on a near-white background = unreadable section headers and admonition titles.

An an option we can either add matching dark-mode overrides, or better, use furo CSS variables (e.g. `--color-background-secondary`) which auto-adapt and would let you delete the duplicated dark-mode block entirely.

doc/conf.py
```diff
@@ -68,10 +67,14 @@
     "sphinx.ext.napoleon",
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
+    "sphinx_copybutton",
+    "sphinx_design",
     "sphinxcontrib.googleanalytics",
     "sphinxcontrib.spelling",
 ]
 
+copybutton_prompt_text = ">>> "
```
antonwolfy: `copybutton_prompt_text = ">>> "` handles only `>>>`. Multi-line doctest blocks with `...` continuation lines (see dpnp.where for an example) will copy with the `...`  prompts left in.
It would be better to cover `...` also, as an option through:
```suggestion
copybutton_prompt_text = r">>> |\.\.\. "
copybutton_prompt_is_regexp = True

```

doc/_static/dpnp-custom.js
```diff
@@ -0,0 +1,42 @@
...
+var separatorAlt = " -- ";
+
+function reformatEntry(container)
+{
+    var paragraphs = container.querySelectorAll(":scope > p");
+    if (!paragraphs.length)
+        return;
+
+    var firstP = paragraphs[0];
+    var idx = firstP.innerHTML.indexOf(separator);
```
antonwolfy: If the separator appears inside an HTML tag/attribute (e.g. a `title= `with an en dash), `substring` splits mid-tag and corrupts markup. Locating the boundary via `textContent` (or DOM walking) might be safer.

doc/_static/dpnp-custom.css
antonwolfy: `@media (prefers-color-scheme: dark)` table rules (lines 77–84) still apply if the OS is dark but the user toggles furo to light. Furo's own CSS gates these with `body:not([data-theme="light"])`.

doc/conf.py
antonwolfy: Seems dumplicating

antonwolfy (approved):
Thank you @vlad-perevezentsev, LGTM!

PR #2937, opened 2026-05-30
Rewrite `tensor.eye` implementation and kernel

ndgrigorian:
This PR proposes an update to the `tensor.eye` constructor, shifting to an approach using a 2D nd_range kernel which instead uses the condition that `col - row == k` along the diagonal, which is correct whether the array is C- or F-contiguous

This eliminates any branching in the kernel, in hopes of improving performance even marginally.


ndgrigorian:
@antonwolfy @vlad-perevezentsev used this as a chance to experiment with vtune and found that performance difference is very insignificant between implementations. Same for GPU utilization, about the same across the board.

This implementation is a bit cleaner in my opinion though, so I think still worth adding

dpnp/tensor/libtensor/include/kernels/constructors.hpp
```diff
@@ -345,9 +345,11 @@ sycl::event full_strided_impl(sycl::queue &q,
 
 typedef sycl::event (*eye_fn_ptr_t)(sycl::queue &,
                                     std::size_t nelems, // num_elements
```
vlad-perevezentsev: `nelems` is not used in the new logic and can be removed

dpnp/tensor/libtensor/include/kernels/constructors.hpp
```diff
@@ -400,9 +406,11 @@ class EyeFunctor
 template <typename Ty>
 sycl::event eye_impl(sycl::queue &exec_q,
                      std::size_t nelems,
```
vlad-perevezentsev: `nelems` is not used in the new logic and can be removed

dpnp/tensor/libtensor/include/kernels/constructors.hpp
```diff
@@ -400,9 +406,11 @@ class EyeFunctor
...
-                     const ssize_t start,
-                     const ssize_t end,
-                     const ssize_t step,
+                     const ssize_t rows,
```
vlad-perevezentsev, line 406: could you also update a changelog?

dpnp/tensor/libtensor/source/eye_ctor.cpp
```diff
@@ -82,7 +77,7 @@ std::pair<sycl::event, sycl::event>
     int dst_typenum = dst.get_typenum();
     int dst_typeid = array_types.typenum_to_lookup_id(dst_typenum);
 
-    const py::ssize_t nelem = dst.get_size();
+    const py::ssize_t nelems = dst.get_size();
```
vlad-perevezentsev, line 80: `nelems` is passed to `fn` which does not use it

dpnp/tensor/libtensor/source/eye_ctor.cpp
```diff
@@ -96,34 +91,29 @@ std::pair<sycl::event, sycl::event>
...
 
-    const py::ssize_t length = std::min({rows, cols, rows + k, cols - k});
-    const py::ssize_t end = start + step * (length - 1);
-
-    char *dst_data = dst.get_data();
-    sycl::event eye_event;
-
     auto fn = eye_dispatch_vector[dst_typeid];
+    sycl::event eye_event =
+        fn(exec_q, static_cast<std::size_t>(nelems), rows, cols, k, stride0,
```
vlad-perevezentsev: same

dpnp/tensor/libtensor/include/kernels/constructors.hpp
ndgrigorian: good catch

vlad-perevezentsev (approved):
LGTM
Thank you @ndgrigorian

PR #2927, opened 2026-05-20
Fix missing strides validation in `dpnp.tensor.usm_ndarray`

vlad-perevezentsev:
This PR proposes to fix #2881 issue.

The problem was that  `usm_ndarray` constructor did not validate user-provided strides when allocating new memory.
This made it possible to create arrays with invalid stride patterns producing negative memory displacement or oversized memory footprints.

This PR adds validation for invalid stride configurations when allocating new memory and raises `ValueError` for unsupported strides.
It also adds a new `test_ctor_invalid_strides` in `test_usm_ndarray_ctor.py` to cover these cases

- [X] Have you provided a meaningful PR description?
- [X] Have you added a test, reproducer or referred to an issue with a reproducer?
- [X] Have you tested your changes locally for CPU and GPU devices?
- [X] Have you made sure that new changes do not introduce compiler warnings?
- [ ] Have you checked performance impact of proposed changes?
- [ ] Have you added documentation for your changes, if necessary?
- [X] Have you added your changes to the changelog?

PR #2939, opened 2026-06-01
Optimize FFT logic for complex strided input to avoid oversized memory allocation

vlad-perevezentsev:
This PR fixes an issue discovered while implementing #2927 where several FFT tests started failing after adding validation for stride configurations with oversized memory footprints.

The problem was that FFT could allocate output arrays using the same strided layout as a non-contiguous input. For inputs such as `a[::2`, this resulted in oversized allocations followed by an additional copy to a contiguous array.

This fix checks whether the memory footprint implied by the input strides exceeds the number of elements in the array. If so, the input is copied to a contiguous layout before configuring the FFT descriptor allowing oneMKL FFT to produce a contiguous output directly

By avoiding oversized allocations and the extra copy this significantly improves the performance of all FFT operations in dpnp with complex strided inputs

fft_perf

Additionally, the cuFFT C-contiguous requirement is now enforced for all code paths. Previously it was only applied within `batch_fft` branch which was a latent bug for 1D non-batch inputs on CUDA.


dpnp/fft/dpnp_utils_fft.py
```diff
@@ -408,12 +408,25 @@ def _fft(a, norm, out, forward, in_place, c2c, axes, batch_fft=True):
...
+            # for the output array and unnecessary copy to contiguous
+            # after oneMKL FFT
+            _strides = dpnp.get_usm_ndarray(a).strides
+            _shape = a.shape
+            # max element offset reachable by the strides
+            max_disp = sum(
+                st * (sh - 1) for st, sh in zip(_strides, _shape) if st > 0
+            )
+            if (max_disp + 1) > a.size:
+                a = dpnp.ascontiguousarray(a)
```
ndgrigorian, line 432: what happens here for negative strides, though? It's likely possible to have an oversized array with negative strides by just reversing every dimension.

vlad-perevezentsev: Negative strides are not allowed in oneMKL FFT (line [287](https://github.com/IntelPython/dpnp/blob/master/dpnp/fft/dpnp_utils_fft.py#L287)) so the input array is copied to C-contiguous before this logic

If you want I can add check like this
```
max_disp = sum(
     abs(st) * (sh - 1) for st, sh in zip(_strides, _shape)
 )
```

as a safety step for future changes to `_copy_array`

ndgrigorian: it should be fine without it in that case, just a comment would be good

vlad-perevezentsev: done

dpnp/fft/dpnp_utils_fft.py
```diff
@@ -408,12 +408,27 @@ def _fft(a, norm, out, forward, in_place, c2c, axes, batch_fft=True):
...
-        ):  # pragma: no cover
+        if dpnp.is_cuda_backend(a):  # pragma: no cover
             a = dpnp.ascontiguousarray(a)
+        else:
+            # Check if the memory footprint of the strides exceeds
+            # the number of elements.
+            # If so, copy to contiguous to avoid oversized allocation
+            # for the output array and unnecessary copy to contiguous
+            # after oneMKL FFT
+            _strides = dpnp.get_usm_ndarray(a).strides
```
antonwolfy: Why do we need to call `get_usm_ndarray` here?

dpnp/fft/dpnp_utils_fft.py
```diff
@@ -408,12 +408,27 @@ def _fft(a, norm, out, forward, in_place, c2c, axes, batch_fft=True):
...
+            _shape = a.shape
+            # Max element displacement reachable by the strides.
+            # Negative strides are handled by _copy_array, so only
+            # positive strides are possible here
```
antonwolfy: There is another regression also, which can be handled separately:
```python
import dpnp as np

a = np.arange(4, dtype='c8')
b = np.broadcast_to(a, (3, 4))

np.fft.fft(b, axis=0)
Cell In[5], line 1
----> 1 np.fft.fft(b, axis=0)

File ~/code/dpnp/dpnp/fft/dpnp_iface_fft.py:122, in fft(a, n, axis, norm, out)
     47 """
     48 Compute the one-dimensional discrete Fourier Transform.
     49
   (...)    118
    119 """
    121 dpnp.check_supported_arrays_type(a)
--> 122 return dpnp_fft(
    123     a, forward=True, real=False, n=n, axis=axis, norm=norm, out=out
    124 )

File ~/code/dpnp/dpnp/fft/dpnp_utils_fft.py:664, in dpnp_fft(a, forward, real, n, axis, norm, out)
    658 if c2r:
    659     # input array should be Hermitian for c2r FFT
    660     a = _make_array_hermitian(
    661         a, axis, dpnp.are_same_logical_tensors(a, a_orig)
    662     )
--> 664 return _fft(
    665     a,
    666     norm=norm,
    667     out=out,
    668     forward=forward,
    669     # TODO: currently in-place is only implemented for c2c, see SAT-7154
    670     in_place=in_place and c2c,
    671     c2c=c2c,
    672     axes=axis,
    673     batch_fft=a_ndim != 1,
    674 )

File ~/code/dpnp/dpnp/fft/dpnp_utils_fft.py:447, in _fft(a, norm, out, forward, in_place, c2c, axes, batch_fft)
    443 a_strides = _standardize_strides_to_nonzero(strides, a.shape)
    444 dsc, out_strides = _commit_descriptor(
    445     a, forward, in_place, c2c, a_strides, index, batch_fft
    446 )
--> 447 res = _compute_result(dsc, a, out, forward, c2c, out_strides)
    448 res = _scale_result(res, a.shape, norm, forward, index)
    450 # Revert swapped axes

File ~/code/dpnp/dpnp/fft/dpnp_utils_fft.py:239, in _compute_result(dsc, a, out, forward, c2c, out_strides)
    231         result = dpnp_array(
    232             out_shape,
    233             dtype=out_dtype,
   (...)    236             sycl_queue=exec_q,
    237         )
    238         res_usm = result.get_array()
--> 239     ht_fft_event, fft_event = fi._fft_out_of_place(
    240         dsc, a_usm, res_usm, forward, depends=dep_evs
    241     )
    242 _manager.add_event_pair(ht_fft_event, fft_event)
    244 if not isinstance(result, dpnp_array):

ValueError: Memory addressed by the output array is not sufficiently ample.
```

antonwolfy: zero strides in case of broadcasting will also go that path, so the comment is not fully correct

dpnp/fft/dpnp_utils_fft.py
```diff
@@ -408,12 +408,27 @@ def _fft(a, norm, out, forward, in_place, c2c, axes, batch_fft=True):
...
+            max_disp = sum(
+                st * (sh - 1) for st, sh in zip(_strides, _shape) if st > 0
+            )
+            if (max_disp + 1) > a.size:
```
antonwolfy, line 431: It'd be helpful to add dedicated tests covering both copy path and no-copy path with transposed/F-contig complex input.

dpnp/fft/dpnp_utils_fft.py
```diff
@@ -408,12 +408,27 @@ def _fft(a, norm, out, forward, in_place, c2c, axes, batch_fft=True):
         a = dpnp.reshape(a, local_shape)
         index = 1
 
+    if not a.flags.c_contiguous:
         # cuFFT requires input arrays to be C-contiguous (row-major)
         # for correct execution
-        if (
-            dpnp.is_cuda_backend(a) and not a.flags.c_contiguous
-        ):  # pragma: no cover
+        if dpnp.is_cuda_backend(a):  # pragma: no cover
```
antonwolfy, line 414: Previously there was no copy for CUDA branch if `batch_fft=False`.
I guess it's indented change, based on the above comment that C-contig is a requirement for cuFFT without any exception. Then we need to update the PR description at least, mentioning that.

dpnp/fft/dpnp_utils_fft.py
vlad-perevezentsev: These cases are already covered in `TestFft.test_strided`
Initially, after this work #2927  I found a problem caused by failures in CI
https://github.com/IntelPython/dpnp/actions/runs/26162136407/job/76970701307?pr=2927#step:15:4340

vlad-perevezentsev: although these tests can be extended to F-contig

antonwolfy: It'd be better to add a separate test for that and to consider both copy and no-copy paths

dpnp/fft/dpnp_utils_fft.py
vlad-perevezentsev: to get element strides instead of byte strides

antonwolfy: Can you rename the var to reflect that. Or place the clear comment.

dpnp/fft/dpnp_utils_fft.py
vlad-perevezentsev: That is a great catch!
I will address it in the following PR
Thank you!

dpnp/fft/dpnp_utils_fft.py
vlad-perevezentsev: done

dpnp/tests/test_fft.py
```diff
@@ -234,6 +234,26 @@ def test_strided_2d(self, stride_x, stride_y):
         expected = numpy.fft.fft(a)
         assert_dtype_allclose(result, expected)
 
+    def test_non_contiguous_no_copy(self):
+        a = generate_random_numpy_array((4, 5, 6), dtype=numpy.complex64)
+        # Non-contiguous input with compact footprint (no copy needed)
+        ia = dpnp.moveaxis(dpnp.array(a), 0, -1)
+        a_np = dpnp.asnumpy(ia)
```
antonwolfy, line 241: It'd fare to have numpy.nadarray in the same non-contig state

antonwolfy: Please ignore, I misread the code

dpnp/fft/dpnp_utils_fft.py
```diff
@@ -408,12 +408,28 @@ def _fft(a, norm, out, forward, in_place, c2c, axes, batch_fft=True):
...
             a = dpnp.ascontiguousarray(a)
+        else:
+            # Check if the memory footprint of the strides exceeds
+            # the number of elements.
+            # If so, copy to contiguous to avoid oversized allocation
+            # for the output array and unnecessary copy to contiguous
+            # after oneMKL FFT
+            elem_strides = dpnp.get_usm_ndarray(a).strides
+            _shape = a.shape
```
antonwolfy: I'd rename and move away above the if branch, since the same is used at lines 439, 444 and so on below
```suggestion
            a_shape = a.shape
```

dpnp/fft/dpnp_utils_fft.py
```diff
@@ -408,12 +408,28 @@ def _fft(a, norm, out, forward, in_place, c2c, axes, batch_fft=True):
...
-        ):  # pragma: no cover
+        if dpnp.is_cuda_backend(a):  # pragma: no cover
             a = dpnp.ascontiguousarray(a)
+        else:
+            # Check if the memory footprint of the strides exceeds
+            # the number of elements.
+            # If so, copy to contiguous to avoid oversized allocation
+            # for the output array and unnecessary copy to contiguous
+            # after oneMKL FFT
+            elem_strides = dpnp.get_usm_ndarray(a).strides
```
antonwolfy: We have exactly the same call at line 443, it'd be better to move above the if-branch to call once.

antonwolfy (approved):
Thank you @vlad-perevezentsev, LGTM with small nits

dpnp/fft/dpnp_utils_fft.py
vlad-perevezentsev: done
Thank you!

antonwolfy (approved):

antonwolfy (approved):

antonwolfy (approved):

PR #2941, opened 2026-06-05
Refactor dpnp includes and add extension example

ndgrigorian:
This PR proposes a refactor of dpnp's includes and the introduction of an example extension, to be tested to make sure examples using dpnp and dpnp.tensor can be written.

The `usm_ndarray_constants.h` header and `dpnp4pybind11.hpp` are moved to `dpnp/include`, removing a relative include and aligning with previous dpctl behavior.

Also introduces dpnp-config.cmake, which enables `find_package(Dpnp)` out of the box, which is installed with dpnp, and `--cmakedir` command line option, which leads to the CMake config.


examples/pybind11/use_dpnp_array/tests/test_use_dpnp_array.py
```diff
@@ -0,0 +1,77 @@
+# *****************************************************************************
+# Copyright (c) 2016, Intel Corporation
```
vlad-perevezentsev:
```suggestion
# Copyright (c) 2026, Intel Corporation
```

examples/pybind11/use_dpnp_array/use_dpnp_array/__init__.py
```diff
@@ -0,0 +1,49 @@
+# *****************************************************************************
+# Copyright (c) 2016, Intel Corporation
```
vlad-perevezentsev:
```suggestion
# Copyright (c) 2026, Intel Corporation
```

examples/pybind11/use_dpnp_array/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
+# *****************************************************************************
+# Copyright (c) 2023, Intel Corporation
```
vlad-perevezentsev:
```suggestion
# Copyright (c) 2026, Intel Corporation
```

examples/pybind11/use_dpnp_array/setup.py
```diff
@@ -0,0 +1,38 @@
+# *****************************************************************************
+# Copyright (c) 2016, Intel Corporation
```
vlad-perevezentsev:
```suggestion
# Copyright (c) 2026, Intel Corporation
```

examples/pybind11/use_dpnp_array/setup.py
```diff
@@ -0,0 +1,38 @@
...
+# - Redistributions of source code must retain the above copyright notice,
+#   this list of conditions and the following disclaimer.
+# - Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+# - Neither the name of the copyright holder nor the names of its contributors
+#   may be used to endorse or promote products derived from this software
+#   without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"ad
```
vlad-perevezentsev:
```suggestion
# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
```

dpnp/__main__.py
```diff
@@ -84,6 +91,11 @@ def main() -> None:
         action="store_true",
         help="Path to dpnp libtensor include directory.",
     )
+    parser.add_argument(
+        "--cmakedir",
+        action="store_true",
+        help="CMake module directory, ideal for setting -DDPCTL_ROOT in CMake.",
```
vlad-perevezentsev: This looks like a direct copy from dpctl without any changes

cmake/dpnp-config.cmake
```diff
@@ -0,0 +1,68 @@
...
+    dpnp4pybind11.hpp
+    PATHS "${_dpnp_include_dir}" "${Python_INCLUDE_DIRS}"
+    PATH_SUFFIXES dpnp/include
+)
+get_filename_component(_dpnp_dir "${Dpnp_INCLUDE_DIR}" DIRECTORY)
+
+find_path(
+    Dpnp_TENSOR_INCLUDE_DIR
+    kernels
+    utils
```
vlad-perevezentsev: `find_path()` uses OR semantics for multiple names so this check will succeed if either `kernels` or `utils` is present and it does not guarantee that both directories exist

examples/pybind11/use_dpnp_array/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+project(
+    use_dpnp_array
+    VERSION 0.1
+    LANGUAGES CXX
+    DESCRIPTION "Example of using dpnp::tensor::usm_ndarray type caster"
+)
+
+set(SYCL_MODULE_PATH "${CMAKE_SOURCE_DIR}/../../../dpnp/backend/cmake/Modules")
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${SYCL_MODULE_PATH})
+find_package(IntelSYCL REQUIRED PATHS ${SYCL_MODULE_PATH} NO_DEFAULT_PATH)
```
vlad-perevezentsev: Since dpctl already ships `IntelSYCLConfig.cmake` could this be replaced with
`find_package(IntelSYCL REQUIRED PATHS "${Dpctl_DIR}" NO_DEFAULT_PATH)` after `find_package(Dpctl REQUIRED)`?
Or does it make sense to install `IntelSYCLConfig.cmake` into `dpnp/resources/cmake/` and use "`${Dpnp_DIR}"` instead?

ndgrigorian: It's a good point that dpctl should already install it, I mostly had local builds in mind. I'm not sure if it makes sense to install it to `dpnp/resources/cmake`, only because it may be redundant, but it may make sense for PyPI installations, not sure, I'll think on that

examples/pybind11/use_dpnp_array/setup.py
vlad-perevezentsev: Not fixed

cmake/dpnp-config.cmake
ndgrigorian: this is how it was done in dpctl, but I can look for an alternative that verifies that both exist

.gitignore
```diff
@@ -3,6 +3,7 @@ _skbuild
 build_cython
 cython_debug
 dpnp.egg-info
```
vlad-perevezentsev: this line is now redundant

cmake/dpnp-config.cmake
```diff
@@ -0,0 +1,65 @@
...
+    Dpnp_INCLUDE_DIR
+    dpnp4pybind11.hpp
+    PATHS "${_dpnp_include_dir}" "${Python_INCLUDE_DIRS}"
+    PATH_SUFFIXES dpnp/include
+)
+get_filename_component(_dpnp_dir "${Dpnp_INCLUDE_DIR}" DIRECTORY)
+
+find_path(Dpnp_TENSOR_INCLUDE_DIR kernels PATHS "${_dpnp_dir}/tensor/libtensor/include")
+
+find_path(Dpnp_TENSOR_INCLUDE_DIR utils PATHS "${_dpnp_dir}/tensor/libtensor/include")
```
vlad-perevezentsev: According to CMake `find_path()` doc:

```
Once one of the calls succeeds the result variable will be set and stored in the cache so that no call will search again.
```
So the second entry is not actually checked once the first search succeeds
I am not sure there is a clean way to verify both directories with a single `find_path()` but since `kernels` and `utils` are expected to coexist in this directory it seems simpler to check only `kernels` here and remove this line

vlad-perevezentsev (approved):
LGTM
Thank you @ndgrigorian 
A few minor comments below:

vlad-perevezentsev (approved):
Great job!
No more comments from my side
LGTM, thank you @ndgrigorian

dpnp/__main__.py
```diff
@@ -84,6 +91,11 @@ def main() -> None:
         action="store_true",
         help="Path to dpnp libtensor include directory.",
     )
+    parser.add_argument(
+        "--cmakedir",
```
antonwolfy, line 95: Would it make sense to test `--cmakedir` / `Dpnp_ROOT` in the example also?

cmake/dpnp-config.cmake
```diff
@@ -0,0 +1,63 @@
...
+#   The version of dpnp found.
+#
+# The module will also explicitly define two cache variables:
+#
+# ``Dpnp_INCLUDE_DIR``
+# ``Dpnp_TENSOR_INCLUDE_DIR``
+#
+
+if(NOT Dpnp_FOUND)
+    find_package(Python 3.10 REQUIRED COMPONENTS Interpreter Development.Module)
```
antonwolfy, line 23: Would it make sense to have upper limit also?
```suggestion
    find_package(Python 3.10...<3.15 REQUIRED COMPONENTS Interpreter Development.Module)
```

.gitignore
```diff
@@ -27,9 +27,13 @@ dpnp_pytest.*
...
+# moved cmake scripts
+dpnp/resources/cmake
+
 *dpnp_backend*
 dpnp/include/dpnp/tensor/*.h
 dpnp/**/*.cpython*.so
 dpnp/**/*.pyd
 *~
 core
+*.so
```
antonwolfy, line 38: We have a pattern with `...*so` already above. Do we need to be more specific here and to ignore only ones generated in example folder?

.github/workflows/conda-package.yml
```diff
@@ -269,6 +269,119 @@ jobs:
...
+      - name: ReInstall dpnp
+        if: steps.install_dpnp.outcome == 'failure'
+        run: |
+          mamba install ${{ env.package-name }}=${{ env.PACKAGE_VERSION }} python=${{ matrix.python }} ${{ env.TEST_CHANNELS }}
+        env:
+          TEST_CHANNELS: '-c ${{ env.channel-path }} ${{ env.channels-list }}'
+
+      - name: Install example requirements
+        run: |
+          mamba install pytest dpcpp_linux-64">=2026.0.0" cmake ninja scikit-build ${{ env.TEST_CHANNELS }}
```
antonwolfy: As an option we can add a dedicated env file to `dpnp/enviroments` folder and install all the requirements in `Setup miniconda` step

cmake/dpnp-config.cmake
```diff
@@ -0,0 +1,63 @@
...
+            ERROR_QUIET
+        )
+    endif()
+endif()
+
+find_path(
+    Dpnp_INCLUDE_DIR
+    dpnp4pybind11.hpp
+    PATHS "${_dpnp_include_dir}" "${Python_INCLUDE_DIRS}"
+    PATH_SUFFIXES dpnp/include
```
antonwolfy, line 45: Do we need to use `NO_DEFAULT_PATH` here ?

CMakeLists.txt
```diff
@@ -327,6 +327,21 @@ else()
     message(FATAL_ERROR "Unsupported system.")
 endif()
 
+# install dpnp-config.cmake for find_package() support
+install(
+    FILES ${CMAKE_SOURCE_DIR}/cmake/dpnp-config.cmake
+    DESTINATION dpnp/resources/cmake
+)
+install(FILES ${CMAKE_SOURCE_DIR}/cmake/dpnp-config.cmake DESTINATION lib/cmake/dpnp)
```
antonwolfy: We can reduce duplication:
```suggestion
foreach(_dest IN ITEMS dpnp/resources/cmake lib/cmake/dpnp)
    install(FILES ${CMAKE_SOURCE_DIR}/cmake/dpnp-config.cmake DESTINATION ${_dest})
endforeach()
```

CMakeLists.txt
```diff
@@ -327,6 +327,21 @@ else()
...
+
+# install dpnp4pybind11.hpp and usm_ndarray_constants.h into include folder
+install(
+    DIRECTORY ${CMAKE_SOURCE_DIR}/dpnp/include/
```
antonwolfy: It actually means the recurses into ALL subdirs. But there is another rule at `dpnp/CMakeLists.txt:147-156` which installs the Cython-generated headers from the build.
Thus the rule will grab `dpnp/tensor/_usmarray.h` and `_usmarray_api.h` and install them to `<prefix>/dpnp/include/dpnp/tensor`. Exactly the same as rules at `dpnp/CMakeLists.txt:147-156` do.

So the same files might be written twice to the same location.

antonwolfy: We might need to specify explicitly:
```cmake
install(FILES
    ${CMAKE_SOURCE_DIR}/dpnp/include/dpnp4pybind11.hpp
    ${CMAKE_SOURCE_DIR}/dpnp/include/usm_ndarray_constants.h
    DESTINATION dpnp/include)
```
or to add `PATTERN "dpnp/tensor" EXCLUDE` to the end of install command.

cmake/dpnp-config.cmake
ndgrigorian: I think it's generally best to keep only the lower bound unless explicitly the upper bound is needed. Most likely, it would build with 3.15 anyway as well

.gitignore
ndgrigorian: we may not, I'll remove it and check

.github/workflows/conda-package.yml
ndgrigorian: it's a good idea

dpnp/__main__.py
ndgrigorian: not sure how you mean, `find_package` effectively tests already `Dpnp_ROOT`. If the example runs, `Dpnp_ROOT` should be set.

.gitignore
ndgrigorian: in terms of do we need to be, no, we don't need to be. I prefer the less specific approach since we don't need `.so` but I can make it more specific

antonwolfy: I'm fine to ignore through `*.so`, but then we need to remove the old one with more specific path

cmake/dpnp-config.cmake
ndgrigorian: we didn't in the dpctl config, but could be added

ndgrigorian: I guess looking into it, `NO_DEFAULT_PATH` just prevents searching default locations, where it can't be located. It's not a problem not to have it, but I added it anyway

dpnp/__main__.py
antonwolfy: As per my understanding `examples/pybind11/use_dpnp_array/setup.py` calls `skbuild.setup()` with no `cmake_args`, and CI runs `setup.py build_ext --inplace` with no` -DDpnp_ROOT`. So in CI `find_package(Dpnp)` resolves purely via the conda prefix (`lib/cmake/dpnp`), never via `--cmakedir`.

While the test `test_cli_options.py::test_cmakedir` checks the printed path, but nothing verifies that `-DDpnp_ROOT=$(python -m dpnp --cmakedir)` actually makes `find_package(Dpnp)` succeed.

cmake/dpnp-config.cmake
```diff
@@ -0,0 +1,69 @@
...
+# all listed variables are TRUE
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(
+    Dpnp
+    REQUIRED_VARS Dpnp_INCLUDE_DIR Dpnp_TENSOR_INCLUDE_DIR
+    VERSION_VAR Dpnp_VERSION
+)
+
+mark_as_advanced(Dpnp_INCLUDE_DIR)
+mark_as_advanced(Dpnp_TENSOR_INCLUDE_DIR)
```
antonwolfy, line 69: Seems we don't have any test/example where `Dpnp_TENSOR_INCLUDE_DIR` is used.

antonwolfy: But it's fine to leave that for the follow-up

dpnp/__main__.py
ndgrigorian: Ah, I see what you mean now. It's doable in a sense, we could pass the path to the example instead, but then it doesn't test the Conda package before exactly.

I'm not sure if one way or the other is better for testing. I think leaving it the current way is the preferred way to approach extensions, as the other way is more for local builds.

cmake/dpnp-config.cmake
ndgrigorian: yeah, we should probably add another example for that.

dpnp/__main__.py
antonwolfy: It seems fine to keep the current test. We can extend/add testing with `--cmakedir` in the follow-up.

antonwolfy (approved):
Thank you @ndgrigorian, no more comments. LGTM!

PR #2945, opened 2026-06-09
Fix tests for `dpctl` change to return tuples

ndgrigorian:
dpctl returns tuples now rather than lists for devices, subdevices, etc. as these should be immutable to users


vlad-perevezentsev (approved):
LGTM
Thank you @ndgrigorian

PR #2912, opened 2026-05-12
Tune tolerance in matvec test and disable large dim in matmul test

antonwolfy:
Few tests are failing with the latest blas libs and numpy:
```bash
  FAILED test_product.py::TestMatmul::test_strided1[-2-float32] - AssertionError: 
  Not equal to tolerance rtol=1.6e-05, atol=1.6e-05
  
  Mismatched elements: 1 / 10000 (0.01%)
  Mismatch at index:
   [6, 5, 4, 8]: 0.13449859619140625 (ACTUAL), 0.13451732695102692 (DESIRED)
  Max absolute difference among violations: 1.873076e-05
  Max relative difference among violations: 0.00013924
   ACTUAL: array([[[[ 9.825830e+01,  8.411874e+01,  2.118525e+02, ...,
            -1.437696e+02, -1.788609e+00,  3.740148e+01],
           [ 8.804939e+01, -2.840893e+01, -1.463431e+02, ...,...
   DESIRED: array([[[[ 9.825830e+01,  8.411874e+01,  2.118525e+02, ...,
            -1.437696e+02, -1.788605e+00,  3.740148e+01],
           [ 8.804940e+01, -2.840893e+01, -1.463431e+02, ...,...
  FAILED test_product.py::TestMatvec::test_axes[axes0] - AssertionError: 
  Not equal to tolerance rtol=8e-15, atol=8e-15
  
  Mismatched elements: 1 / 96 (1.04%)
  Mismatch at index:
   [0, 0, 0, 1]: 0.36551382063033344 (ACTUAL), 0.36551382063031923 (DESIRED)
  Max absolute difference among violations: 1.42108547e-14
  Max relative difference among violations: 3.88791173e-14
   ACTUAL: array([[[[  17.130334,    0.365514,  -16.568001],
           [  52.87157 ,  -75.25945 , -105.961006]],
  ...
   DESIRED: array([[[[  17.130334,    0.365514,  -16.568001],
           [  52.87157 ,  -75.25945 , -105.961006]],
  ...
  FAILED test_product.py::TestMatvec::test_axes[axes1] - AssertionError: 
  Not equal to tolerance rtol=8e-15, atol=8e-15
  
  Mismatched elements: 1 / 96 (1.04%)
  Mismatch at index:
   [0, 0, 0, 1]: 0.36551382063033344 (ACTUAL), 0.36551382063031923 (DESIRED)
  Max absolute difference among violations: 1.42108547e-14
  Max relative difference among violations: 3.88791173e-14
[...]

PR #2949, opened 2026-06-15
Re-enable dim=4 in strided matmul test with proper cross-BLAS tolerances

antonwolfy:
## Summary

gh-2912 introduced two temporary test workarounds after a CI failure on `ubuntu-latest`:
- widened the tolerance on `TestMatvec::test_axes` to `factor=40`;
- disabled the `dim=4` iteration in `TestMatmul::test_strided1`.

Both failures share the same benign root cause, so this PR re-enables the disabled coverage and replaces the ad-hoc mitigations with documented, dtype-aware tolerances:

- **`test_axes`** (matvec, float64) — keeps `factor=40`. dpnp computes `matvec` via oneMKL `gemm_batch` while NumPy uses OpenBLAS `gemv`; the summation order differs and the float64 result deviates at the noise floor (~1.4e-14), which exceeds the default `8e-15` tolerance.
- **`test_strided1`** (matmul) — re-enables `dim=4`. dpnp copies the strided input to c-contiguous and runs it through oneMKL `gemm_batch`, again against NumPy's OpenBLAS. For float32, the length-20 contraction deviates by ~1.87e-5, just above the old `factor=16` bound (~1.81e-5), so the tolerance is bumped to `factor=24` **for float32 only** (pass-bound ~2.27e-5). Integer input is compared exactly and is unaffected.

## Root cause

Neither discrepancy is a dpnp correctness bug. Both are expected floating-point non-associativity between two different BLAS backends (oneMKL vs OpenBLAS), surfaced when the conda-forge NumPy in CI advanced its bundled OpenBLAS. There is nothing to fix upstream, so documented tolerances are the correct handling and let us keep full test coverage (including `dim=4`).

### DIFF
--- a/dpnp/tests/test_product.py
+++ b/dpnp/tests/test_product.py
@@ -887,6 +887,13 @@ class TestMatmul:
         ids=["-2", "2", "(-2, 2)", "(2, -2)"],
     )
     def test_strided1(self, dtype, stride):
+        # dpnp copies the strided input into a c-contiguous array and runs the
+        # batched product through oneMKL gemm_batch, while NumPy uses OpenBLAS.
+        # The two backends accumulate in a different order, so the float32
+        # result deviates at the noise floor (~2e-5 for a length-20 contraction
+        # with dim=4), which exceeds the default tolerance. Integer input is
+        # exact, so the factor only affects the float32 case.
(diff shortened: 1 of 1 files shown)


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2950, opened 2026-06-16
Removed support for arrays of 2-dimensional vectors in dpnp.cross

antonwolfy:
NumPy 2.5 expires the deprecation of 2-dimensional vectors in `numpy.cross`: the function now requires (arrays of) 3-dimensional vectors and raises `ValueError` otherwise.

This PR brings `dpnp.cross` in line with that behavior.
Dpnp had already deprecated 2-D vectors in `dpnp.cross` back in 0.17.0 (via `DeprecationWarning`), so this removes the deprecated path entirely.

Note, `dpnp.cross` now rejects 2-D vectors unconditionally, including when running tests against `NumPy < 2.5` (where numpy.cross still computes a 2-D result). This is an intentional.

### DIFF
--- a/dpnp/tests/test_product.py
+++ b/dpnp/tests/test_product.py
@@ -1,7 +1,12 @@
 import dpctl
 import numpy
 import pytest
-from numpy.testing import assert_allclose, assert_array_equal, assert_raises
+from numpy.testing import (
+    assert_allclose,
+    assert_array_equal,
+    assert_raises,
+    assert_raises_regex,
+)
 
 import dpnp
 from dpnp.dpnp_utils import map_dtype_to_device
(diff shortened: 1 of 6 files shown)


dpnp/tests/test_product.py
```diff
@@ -149,31 +148,41 @@ def test_linalg(self, axis):
...
 
-    def test_error(self):
-        a = dpnp.arange(3)
-        b = dpnp.arange(4)
-        # Incompatible vector dimensions
-        with pytest.raises(ValueError):
-            dpnp.cross(a, b)
+    @testing.with_requires("numpy>=2.5")
+    @pytest.mark.parametrize("xp", [numpy, dpnp])
+    def test_2x2(self, xp):
```
vlad-perevezentsev, line 153: Does it make sense to cover this case for dpnp only?
To increase coverage and avoid waiting for coverage to switch to NumPy 2.5

antonwolfy: I'd prefer to keep that. NumPy 2.5 will be released soon and IMHO it's better to test the use case works in the same way for both numpy and dpnp.

vlad-perevezentsev: Ok, agreed

vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2951, opened 2026-06-16
Add test to cover scalar overflow use case in dpnp.where

antonwolfy:
This PR adds test coverage for the scalar-overflow behavior of `dpnp.where` (fixed with NumPy 2.5) and aligns the existing shape-mismatch error test with NumPy.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2953, opened 2026-06-17
Return complex arrays from eig / eigvals

antonwolfy:
NumPY 2.5 changed `numpy.linalg.eig` and `numpy.linalg.eigvals` so that for general (non-symmetric) matrices they always return complex-typed eigenvalues, instead of value-dependently casting the result back to a real type when the imaginary parts happen to be zero. For real input, float64 now yields complex128 and float32 yields complex64.

dpnp.linalg.eig / eigvals are thin NumPy fallbacks (the OneMKL geev routine is not yet wired up), so they inherit this behavior directly from the installed NumPy.

This PR brings dpnp's documentation in line with the new always-complex output.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2954, opened 2026-06-18
Return unitary vh from `dpnp.linalg.svd` for singular Hermitian inputs

antonwolfy:
`dpnp.linalg.svd(..., hermitian=True)` could return a non-unitary `vh` when the input array is singular. In the Hermitian path, the sign of each eigenvalue is factored out of the eigenvectors into `vh`. The sign was computed with `dpnp.sign(s)`, which returns `0` for zero eigenvalues, zeroing out the corresponding rows of `vh` and breaking unitarity.

The fix replaces `dpnp.sign(s)` with `dpnp.copysign(1.0, s)`, which always yields ±1, mirroring the change included into NumPy 2.5.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2955, opened 2026-06-18
Increase test coverage in dpnp.nextafter for signed zeros

antonwolfy:
This PR adds a test verifying that `dpnp.nextafter` returns the correctly signed zero when stepping between `+0.0` and `-0.0`.

It mirrors the regression test added with NumPy 2.5, which fixed `nextafter` returning the wrong-signed zero for the equal-zero case where the destination's sign must determine the sign of the resulting zero.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2958, opened 2026-06-18
Remove in-place modification of `shape` attribute for `numpy.ndarray`

antonwolfy:
NumPy 2.5 is going to deprecate in-place assignment to `ndarray.shape`.

Previously `_as_zero_dim_ndarray` in `dpnp.tensor` relied on this pattern (`view.shape = tuple()`) to turn a size-1 host buffer into a 0-dim array before extracting a Python scalar.

This PR replaces it with `reshape(())`, which is the recommended non-deprecated alternative and returns a true
0-dim array.


vlad-perevezentsev (approved):
LGTM
Thank  you @antonwolfy

PR #2959, opened 2026-06-19
Allow same-kind casting of the result into `out` in dpnp.take and dpnp.compress

antonwolfy:
Previously `dpnp.take` and `dpnp.compress` raised a `TypeError` whenever the provided `out` array had a dtype different from the input array (strictly [followed](https://data-apis.org/array-api/latest/API_specification/generated/array_api.take.html) Python array API):
> out (array) – an array having the same data type as x.

This PR relaxes that check to allow casting the result into an `out` array of a **different but same-kind** dtype (e.g. `int32` -> `int64`, `int32` -> `float32`), aligning the behavior closer to dpnp documentation [clamming](https://intelpython.github.io/dpnp/reference/generated/dpnp.take.html):
> It should be of the appropriate shape and dtype.

and aligning with recent changes done in NumPy 2.5.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2963, opened 2026-06-22
Normalize all line endings to LF via .gitattributes

antonwolfy:
Enforces LF (`\n`) line endings across the repository to prevent CRLF/LF inconsistencies between platforms. A few tracked files had crept in with CRLF endings; this normalizes them and adds a `.gitattributes` rule so the policy is enforced automatically going forward.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2965, opened 2026-06-22
Return tuple instead of list from atleast_1d/2d/3d and ogrid

antonwolfy:
NumPy returns a tuple of arrays (not a list) from several functions that yield multiple arrays.
This PR aligns dpnp with that behavior.

### What is done

- `dpnp.atleast_1d`, `dpnp.atleast_2d`, `dpnp.atleast_3d` and `dpnp.ogrid` now return a tuple of arrays instead of a list. Single-input
  `atleast_*d` and single-slice `ogrid` still return a bare array, and `mgrid` still returns a single stacked array (matching NumPy).
- Added/extended tests to assert the return container type matches NumPy. While here, added the same parity checks for other tuple-returning functions that lacked them (`nonzero`, `where`, `ix_`, `tril_indices`, `triu_indices`, `unravel_index`, `unstack`).

The split family (`array_split`, `split`, `hsplit`, `vsplit`, `dsplit`) and `histogramdd` still return lists, matching current NumPy.


ndgrigorian (approved):
LGTM

PR #2967, opened 2026-06-23
Deprecate setting the `shape` attribute of `dpnp.ndarray` and `dpnp.tensor.usm_ndarray`

antonwolfy:
Setting the array shape in place via the `shape` attribute (`a.shape = ...`) is now **deprecated** for both `dpnp.ndarray` and `dpnp.tensor.usm_ndarray`. Assigning to `.shape` emits a `DeprecationWarning` while preserving the existing behavior for now. 

This aligns `dpnp` with NumPy 2.5, which deprecated the same in-place shape assignment.

Users should reshape arrays with `dpnp.reshape` / `dpnp.tensor.reshape` (or the `reshape` method) instead:

```python
# Deprecated
a.shape = (3, 8)

# Preferred
a = dpnp.reshape(a, (3, 8))
```

### DIFF
--- a/dpnp/dpnp_iface_arraycreation.py
+++ b/dpnp/dpnp_iface_arraycreation.py
@@ -3276,7 +3276,7 @@ class OGridClass:
     See Also
     --------
     :obj:`dpnp.mgrid` : Work like :obj:`dpnp.ogrid` but returns dense
-        (or fleshed out) mesh grids..
+        (or fleshed out) mesh grids.
     :obj:`dpnp.meshgrid`: Return coordinate matrices from coordinate vectors.
     :obj:`dpnp.r_` : Array concatenator.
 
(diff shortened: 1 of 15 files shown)


dpnp/dpnp_iface_arraycreation.py
```diff
@@ -3247,21 +3256,30 @@ class OGridClass:
...
-    out : dpnp.ndarray or tuple of dpnp.ndarray
+    out : one dpnp.ndarray or tuple of dpnp.ndarray
         If the input is a single slice, returns an array.
         If the input is multiple slices, returns a tuple of arrays, with
         only one dimension not equal to 1.
 
+    See Also
+    --------
+    :obj:`dpnp.mgrid` : Work like :obj:`dpnp.ogrid` but returns dense
+        (or fleshed out) mesh grids..
```
vlad-perevezentsev:
```suggestion
        (or fleshed out) mesh grids.
```

vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy !

PR #2971, opened 2026-06-24
Fix `dpnp.mgrid`/`dpnp.ogrid` inconsistency for complex steps with non-integer magnitude

antonwolfy:
For a complex step, the integer part of its magnitude is the number of points to generate (start and stop inclusive). The single-slice and tuple-of-slices code paths disagreed when the step magnitude was **not** an integer (e.g. `2.5j`):
- the tuple path used `int(abs(step)) - 1` as the divisor (correct), while
- the single-slice path used the float magnitude `abs(step) - 1` as the divisor.

As a result `dpnp.mgrid[0:10:2.5j]` and `dpnp.mgrid[0:10:2.5j,]` produced different arrays, and the single-slice form disagreed with NumPy.

```python
>>> dpnp.mgrid[0:10:2.5j]      # before: [0. , 6.667]   after: [ 0., 10.]
>>> dpnp.mgrid[0:10:2.5j,][0]  # always: [ 0., 10.]     (matches NumPy)
```


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2983, opened 2026-07-01
Extend dependabot to cover pip, conda, and pre-commit

antonwolfy:
This PR extends the Dependabot configuration so the pinned dependency versions used across CI are tracked and bumped automatically, and consolidates pre-commit hook updates into Dependabot instead of a dedicated workflow.

Previously only the `github-actions` ecosystem was covered.

Three new `updates` entries:

- **`pip`** (`/environments`) — updates the pinned PyPI packages in the `environments/*.txt` files (`base_build_docs.txt`, `coverage.txt`, `dpctl_pkg.txt`). Dependabot scans the directory for requirements-like files regardless of name and reads any `--index-url` (e.g. the dppy dev channel in `dpctl_pkg.txt`) directly from the file.
- **`conda`** (`/environments`) — updates the plain conda-forge `pkg=X.Y.Z` pins in the env YAML files consumed by `setup-miniconda`.
- **`pre-commit`** (`/`) — updates hook versions in `.pre-commit-config.yaml`.

Also PR deletes `.github/workflows/pre-commit-autoupdate.yml`. Its flow is now handled by the `pre-commit` Dependabot entry.


ndgrigorian (approved):
I don't see an issue, only way to test is to merge, so LGTM

PR #2986, opened 2026-07-02
Fix `dpnp.interp` returning nan at exact knot when fp contains inf

antonwolfy:
`dpnp.interp` returned `NaN` when a query point landed exactly on an interior knot of `xp` and the adjacent `fp` value was `Inf`. Interpolating at such a point should return the knot value directly.

The root cause is IEEE-754 arithmetic in the linear formula: when `fp[x_idx+1]` is `inf`, `slope` becomes `Inf`, and since `x_val - xp[x_idx] == 0` at the knot, the result is `inf * 0 = nan`. The existing NaN-recovery fallback also yields `nan` in this case.

```python
import dpnp as np

x = np.array([2.0])
xp = np.array([1.0, 2.0, 3.0, 4.0])
fp = np.array([1.0, 2.0, np.inf, 4.0])
np.interp(x, xp, fp)
# before: array([nan])
# after:  array([2.])   (matches numpy.interp)
```


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #2996, opened 2026-07-09
task: refactor ASV benchmarks

jharlow-intel:
Fixes and refactors ASV benchmarking to mirror dpbench implementations.


benchmarks/benchmarks/dpbench/workloads/__init__.py
```diff
@@ -1,6 +1,5 @@
-# -*- coding: utf-8 -*-
 # *****************************************************************************
-# Copyright (c) 2026, Intel Corporation
+# Copyright (c) 2020, Intel Corporation
```
antonwolfy: Should be an year, when file is created (here and in other places):
```suggestion
# Copyright (c) 2026, Intel Corporation
```

benchmarks/benchmarks/common.py
```diff
@@ -52,10 +52,16 @@
     "int64",
     "float64",
     "complex64",
-    "longfloat",
+    # numpy.longfloat is an alias of numpy.longdouble that was removed in
+    # NumPy 2.0; use numpy.longdouble, which exists on both 1.x and 2.x.
+    "longdouble",
     "complex128",
```
antonwolfy: dpnp does not support these data types:
```suggestion
```

benchmarks/benchmarks/dpbench/workloads/black_scholes.py
```diff
@@ -0,0 +1,135 @@
...
+    "M": {"nopt": 134217728, "seed": 777777},
+    "L": {"nopt": 268435456, "seed": 777777},
+}
+# Presets actually exercised by ASV. Larger presets require several GiB of
+# device memory; add them here to benchmark bigger problem sizes.
+ASV_PRESETS = ["S"]
+
+
+def initialize(nopt, seed, types_dict):
+    import numpy as np
```
antonwolfy: The same name used to import dpnp, it'd be better to avoid that (here and in other places):
```suggestion
    import numpy
```

benchmarks/benchmarks/dpbench/workloads/black_scholes.py
```diff
@@ -0,0 +1,135 @@
...
+dpBench (https://github.com/IntelPython/dpbench), and the metadata below
+mirrors ``dpbench/configs/bench_info/black_scholes.toml``.
+"""
+
+import dpnp as np
+
+# --- dpBench benchmark metadata (see black_scholes.toml) --------------------
+
+NAME = "black_scholes"
+PRECISION = "double"
```
antonwolfy, line 44: Some iGPU might not support double precision. Do we need to support fallback run with float precision? Or that does not make a lot of sense?

benchmarks/benchmarks/dpbench/workloads/black_scholes.py
```diff
@@ -0,0 +1,135 @@
...
+# Data-size presets, copied verbatim from dpBench.
+PRESETS = {
+    "S": {"nopt": 524288, "seed": 777777},
+    "M16Gb": {"nopt": 67108864, "seed": 777777},
+    "M": {"nopt": 134217728, "seed": 777777},
+    "L": {"nopt": 268435456, "seed": 777777},
+}
+# Presets actually exercised by ASV. Larger presets require several GiB of
+# device memory; add them here to benchmark bigger problem sizes.
+ASV_PRESETS = ["S"]
```
antonwolfy: I wonder if the result will be unstable when running with small preset.
What is about `M16Gb` preset or more smart logic to select the preset depending on available memory?
Or that is not needed and ASV runs demonstrate stable results with even `S` preset?

pyproject.toml
```diff
@@ -74,6 +74,10 @@ readme = {file = "README.md", content-type = "text/markdown"}
 requires-python = ">=3.10,<3.15"
 
 [project.optional-dependencies]
+benchmark = [
+  "asv>=0.6",
+  "scipy"
```
antonwolfy, line 82: Why do we need to install scipy?

benchmarks/benchmarks/dpbench/workloads/black_scholes.py
```diff
@@ -0,0 +1,135 @@
...
+    "t",
+    "rate",
+    "volatility",
+    "call",
+    "put",
+]
+# Arguments that are arrays and therefore copied to the device.
+ARRAY_ARGS = ["price", "strike", "t", "call", "put"]
+# Arguments that the kernel writes into.
+OUTPUT_ARGS = ["call", "put"]
```
antonwolfy, line 60: Is that expected that `OUTPUT_ARGS` is never used? Do we need to have a validation step which consumes that?

benchmarks/benchmarks/dpbench/_dpbench_runner.py
```diff
@@ -0,0 +1,165 @@
...
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+# THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+
+"""Minimal re-implementation of dpBench's benchmark execution model for ASV.
+
+dpBench (https://github.com/IntelPython/dpbench) drives its benchmarks through
+a fairly heavy runner that spawns a sub-process per framework, resolves TOML
+configuration, validates results against a reference and persists timings to a
```
antonwolfy, line 33: Seems there is no validation against the reference is implemented / migrated. Or is that intentional?

benchmarks/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+Benchmarking dpnp using Airspeed Velocity.
+Read more about ASV [here](https://asv.readthedocs.io/en/stable/index.html).
+
+## Usage
+
+Unlike a pure-Python project, dpnp is a SYCL/DPC++ extension that requires the
+Intel oneAPI compiler and a lengthy build, so ASV does not build dpnp itself:
+`build_command` in `asv.conf.json` is empty and the benchmarks are run against
+an **existing environment** that already has dpnp installed.
```
antonwolfy, line 11: Isn't the below `pip install ".[benchmark]"` command will trigger full dpnp rebuild?

pyproject.toml
jharlow-intel: blackscholess benchmark uses scipy.special.erf it seems

benchmarks/benchmarks/dpbench/workloads/black_scholes.py
jharlow-intel: that is merely the benchmarks' default precision value, the runner will do both single and double precision, skipping if it is not supported by the igpu

benchmarks/benchmarks/dpbench/workloads/black_scholes.py
jharlow-intel: consumed now!

benchmarks/benchmarks/dpbench/_dpbench_runner.py
jharlow-intel: not intentional and validation is now implemented

benchmarks/README.md
jharlow-intel: Yes, fixed that thanks for pointing it out

benchmarks/asv.conf.json
```diff
@@ -1,89 +1,24 @@
...
-    // The URL or local path of the source code repository for the
-    // project being benchmarked
+    "project_url": "https://github.com/IntelPython/dpnp",
+    "show_commit_url": "https://github.com/IntelPython/dpnp/commit/",
     "repo": "..",
-
-    // List of branches to benchmark. If not provided, defaults to "master"
-    // (for git) or "tip" (for mercurial).
     "branches": [
         "HEAD"
```
vchamarthi: after refactoring asv config need real branch names:

```suggestion
        "master",
        "dev-milestone"
```

jharlow-intel: Fixed!

benchmarks/README.md
```diff
@@ -0,0 +1,267 @@
...
+
+```bash
+asv run --environment existing:/full/conda/path/envs/dpnp_env/bin/python \
+    --launch-method spawn
+```
+
+`asv.conf.json` sets `branches` to `HEAD` rather than to named branches, so that
+results recorded on a feature branch are picked up. With named branches
+`asv publish` reports `Couldn't find <hash> in branches (...)` and silently
+drops them.
```
antonwolfy: That says the config sets branches to `HEAD` so feature-branch results are picked up, and warns that named branches make asv publish silently drop results. But `benchmarks/asv.conf.json:7-10` actually sets `["master", "dev-milestone"]` — and the diff shows this PR changed it from `["HEAD"]`.

By the README's own reasoning the documented workflow will silently drop results on any feature branch. Either the config regressed or the README is stale.

benchmarks/benchmarks/dpbench/workloads/black_scholes.py
```diff
@@ -0,0 +1,176 @@
...
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+# THE POSSIBILITY OF SUCH DAMAGE.
+# *****************************************************************************
+
+"""Black-Scholes formula workload.
+
+The dpnp implementation, the NumPy reference and the data initialization are
+copied verbatim from dpBench (https://github.com/IntelPython/dpbench), and the
```
antonwolfy, line 32: Verbatim dpBench code relicensed with no provenance in the header — the workloads and runner state in their docstrings they're "copied verbatim from dpBench" (Apache-2.0), yet each carries only dpnp's standard BSD-3-Clause header `_dpbench_runner.py:1-27` with no SPDX tag or original-license note.

Both projects are Intel-copyright so relicensing is likely fine legally, and prose-level attribution exists (README + docstrings) — but the file-header license lineage is silent and worth an explicit maintainer sign-off.

benchmarks/benchmarks/dpbench/_dpbench_runner.py
```diff
@@ -0,0 +1,288 @@
...
+    ``rel_error``. Raises :exc:`ValueError` naming the offending argument
+    instead of returning a bool, so a wrong result fails the benchmark rather
+    than being silently timed.
+    """
+    for name, ref in expected.items():
+        val = actual[name]
+        if numpy.allclose(ref, val):
+            continue
+        error = relative_error(ref, val)
+        if error >= rel_error:
```
antonwolfy: If arrays contain NaN, `relative_error` is NaN and `NaN >= rel_error` is False, so validate() won't raise.

benchmarks/benchmarks/dpbench/workloads/black_scholes.py
antonwolfy: Do we need to add something like?
```
# Portions of this file are derived from IntelPython/dpBench
# (https://github.com/IntelPython/dpbench), originally licensed
# Apache-2.0, Copyright (c) 2022-2023 Intel Corporation, and
# relicensed here under BSD-3-Clause by the copyright holder.
```
Or since it's another IntelPython project there is no need to any provenance?

benchmarks/benchmarks/dpbench/_dpbench_runner.py
```diff
@@ -0,0 +1,288 @@
...
+        actual = {
+            arg: dpnp.asnumpy(self.kwargs[arg])
+            for arg in self.workload.OUTPUT_ARGS
+        }
+        validate(expected, actual)
+
+    def _reference_outputs(self):
+        """Run the NumPy reference on freshly initialized host data."""
+        # A fresh initialization is required: the kernel writes into its output
+        # arrays, so ``self._host_data`` no longer holds their initial values.
```
antonwolfy: Stale comment — references `self._host_data`, which no longer exists.

benchmarks/benchmarks/bench_dpbench.py
```diff
@@ -0,0 +1,155 @@
...
+    def setup(self, preset, precision):
+        queue = default_queue()
+        skip_unsupported_dtype(queue, runner.float_dtype(precision))
+
+        if preset not in self.WORKLOAD.PRESETS:
+            raise SkipNotImplemented(
+                f"{self.WORKLOAD.NAME} has no {preset} preset."
+            )
+
+        if not runner.preset_fits(self.WORKLOAD, preset, queue.sycl_device):
```
antonwolfy: The fit verdict is computed with `itemsize=8 (f8)` for both the single and double param cells:
```suggestion
        if not runner.preset_fits(self.WORKLOAD, preset, queue.sycl_device, precision):
```

benchmarks/benchmarks/bench_elementwise.py
```diff
@@ -26,105 +26,107 @@
 # THE POSSIBILITY OF SUCH DAMAGE.
 # *****************************************************************************
 
-import numpy
+"""Benchmarks for unary elementwise math functions, dpnp against NumPy."""
```
antonwolfy: Elementwise covers 26 unary transcendental ufuncs — a subset of NumPy's `bench_ufunc.py`, which benchmarks all 95 ufuncs via one parametrized method.
Two notes:
- `degrees` and `rad2deg` are aliases of the same ufunc, so they benchmark the identical kernel twice
- the notable gaps vs. NumPy are unary rounding/sign (floor, ceil, trunc, rint, sign, abs) and binary arithmetic (add, multiply, divide, power, hypot, arctan2)

antonwolfy: It runs 26 unary transcendental `ufuncs` over `_DTYPES = [float64, float32, int64, int32]`.

Integer input from `arange` promotes to float and mostly lands out of domain (nan/inf), and mixes native-FP timings with int→float-promotion timings. For reference, NumPy's own ASV suite restricts its transcendental/strided benchmarks to float dtypes ('e','f','d') and filters others by `ufunc.types`.

`_DTYPES` is shared with `bench_linalg.py` (where ints are valid), so suggest a separate float-only axis `(_FLOAT_DTYPES = [float64, float32])` used only in the elementwise class. This also removes the risk that a dpnp `ufunc` rejecting int input where NumPy promotes it shows up as an asymmetric failure.

benchmarks/benchmarks/bench_linalg.py
```diff
@@ -26,153 +26,49 @@
...
-        self.a = np.arange(size * size, dtype=dt).reshape((size, size))
 
-        # self.ac = self.a.copy()
-        # self.at = self.a.T
-        # self.atc = self.a.T.copy()
-        # self.b = np.arange(240000.0).reshape(400, 600)
-        self.b = np.arange(size * size, dtype=dt).reshape((size, size))
-        # self.c = np.arange(600)
-        # self.d = np.arange(400)
+class MatMul:
```
antonwolfy, line 52: `MatMul` covers `dot`/`matmul`/`inner`/`einsum` over a clean square-size sweep (16→1024) and both executors — the size sweep is actually more systematic than NumPy's Eindot.

Two suggestions vs. NumPy's bench_linalg.py:
- add a transposed/non-contiguous variant (e.g. dot(a, a.T)) — NumPy's Eindot benches these deliberately because they hit different BLAS/stride kernels, and dpnp/oneMKL is likely no different;
- if broadening later, NumPy also covers decompositions (svd/det/solve) and norm, which dpnp supports via oneMKL LAPACK. Including integer dtypes here is fine (unlike the transcendental case, integer matmul is well-defined).

benchmarks/README.md
jharlow-intel: README is stale, thanks for picking up on this. Gireesh informed me I needed to change the asv.conf.json to include those branches but I didn't review the documentation here afterwards. Will do so now

benchmarks/benchmarks/dpbench/workloads/black_scholes.py
jharlow-intel: I had to assume that because we own both repos, I can copy IP/code from either without having to worry about licensing. Open to adding the clause though if we really think it's necessary.

I believe one goal is to eventually decommission dpbench anyways and start moving our benchmarks to be in-repo for everything like this

benchmarks/benchmarks/bench_elementwise.py
jharlow-intel: Cleaned up the gaps and double kernel benchmarking. Super good catch on that one!

jharlow-intel: implemented, thanks for the suggestion

benchmarks/benchmarks/bench_linalg.py
jharlow-intel: done, and broadened _now_ just for the sake of it

benchmarks/benchmarks/bench_random.py
```diff
@@ -26,30 +26,36 @@
...
-class Sample(Benchmark):
-    executors = {"dpnp": dpnp, "numpy": numpy}
-    params = [["dpnp", "numpy"], [2**16, 2**20, 2**24]]
+    params = [_EXECUTOR_NAMES, _SIZES_1D]
     param_names = ["executor", "size"]
 
     def setup(self, executor, size):
-        self.executor = self.executors[executor]
+        self.executor = _EXECUTORS[executor]
+        self.sync = make_synchronizer(executor)
```
antonwolfy, line 58: The listed random functions does not accept `dtype` keyword and it might be a case when dpnp return float32 result on device without fp64 support, while numpy always returns float64.
I see there are two possible options:
- skip running bench_random on device without fp64
- bench only rnadom function with accept `dtype` as an input keyword and so use `skip_unsupported_dtype` as in other benchmarks

benchmarks/benchmarks/bench_random.py
```diff
@@ -26,30 +26,36 @@
...
         np = self.executor
-        np.random.rand(size)
+        self.sync(np.random.rand(size))
 
     def time_randn(self, executor, size):
         np = self.executor
-        np.random.randn(size)
+        self.sync(np.random.randn(size))
 
     def time_random_sample(self, executor, size):
```
antonwolfy: rand and random_sample are the identical functions

benchmarks/benchmarks/dpbench/_dpbench_runner.py
```diff
@@ -0,0 +1,291 @@
...
+
+def validate(expected, actual, rel_error=1e-05):
+    """Check that ``actual`` matches ``expected`` closely enough.
+
+    A mismatch is tolerated only while the relative error stays below
+    ``rel_error`` and is finite. Raises :exc:`ValueError` naming the offending
+    argument instead of returning a bool, so a wrong result fails the benchmark
+    rather than being silently timed.
+    """
+    for name, ref in expected.items():
```
antonwolfy, line 191: It probably would be better to use `numpy.testing.assert_allclose` and tol per dtype:
```python
_VALIDATION_TOL = {
    "single": {"rtol": 1e-3, "atol": 1e-4},
    "double": {"rtol": 1e-6, "atol": 1e-9},
}

def validate(expected, actual, precision="double"):
    tol = _VALIDATION_TOL[precision]
    for name, ref in expected.items():
        assert_allclose(
            actual[name],
            ref,
            equal_nan=True,
            err_msg=f"Validation failed for {name!r}",
            **tol,
        )
```

benchmarks/benchmarks/bench_random.py
jharlow-intel: Fixed

benchmarks/benchmarks/bench_random.py
jharlow-intel: Implemented it, LLM summary of changes made:
> Did both, since they turn out to be complementary rather than alternatives. Sample keeps the module-level functions and now skips its dpnp points without fp64 via skip_unsupported_dtype, because none of them accepts a dtype (and there is no dpnp.random.default_rng). Added TypedSample with an explicit dtype axis, pairing dpnp's RandomState.uniform/normal against NumPy's default_rng().random/standard_normal — the only dtype-capable entry points on each side — so float32 is still compared on a device without fp64.

benchmarks/benchmarks/dpbench/_dpbench_runner.py
jharlow-intel: pretty much added exactly as you suggested, slight changes such as no precision default

antonwolfy (approved):
Thank you @jharlow-intel, no more comments from me.
Happy to see these move in-repo.

PR #3001, opened 2026-07-15
Release transient SYCL queues at the end of each test file to avoid OOM on low-memory GPUs

antonwolfy:
Running the full test suite in a single process steadily drains device memory and can end in a `RuntimeError: ... UR_RESULT_ERROR_OUT_OF_RESOURCES` (seen intermittently on a low-memory iGPU, where the kernel logs `VM worker error: -12` / `exec queue reset detected`).

Root cause: every distinct `dpctl.SyclQueue` a test creates is retained **for the whole session** as a key in dpctl's `SequentialOrderManager` (the map is keyed by queue identity and only ever grows). Each retained entry holds a host-task event that pins the backing USM allocation, so the memory is never released. Over a full run the map accumulates ~300 queues
and free device memory bleeds from several GB down to ~1 GB, occasionally crossing the device limit.

This is amplified on integrated GPUs, where "device memory" is system RAM, so the leaked USM competes directly with the host and the container's memory budget.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #3006, opened 2026-07-22
Replace deprecated `nd_item::barrier()` with `sycl::group_barrier()` in accumulators and gemm kernels

antonwolfy:
The Intel DPC++ compiler (starting with the 2026.2.0) emits a deprecation warning for the `nd_item::barrier()` member function:
> warning: 'barrier' is deprecated: use sycl::group_barrier() free function
> instead [-Wdeprecated-declarations]

Notes on correctness:
- `nd_item::barrier()` is a **work-group** barrier, so it maps to `group_barrier(it.get_group(), …)`, not a sub-group barrier. The surrounding code depends on work-group-wide synchronization of shared-local-memory buffers.
- The old `fence_space::local_space` argument controlled *which memory* was fenced (local only). `group_barrier`'s optional argument is instead a `memory_scope`; the chosen `work_group` scope fences both local and global memory across the work-group — a superset of the previous behavior, so it is strictly safe. On the target Intel devices a work-group barrier lowers to the same hardware barrier regardless, so there is no performance impact.

This PR replaces the deprecated member function to use of `sycl::group_barrier()`.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #3009, opened 2026-07-29
Add `dpnp.tensor.broadcast_shapes` to align with the 2025.12 Python array API

antonwolfy:
The 2025.12 revision of the Python array API standard adds a standalone [`broadcast_shapes`](https://data-apis.org/array-api/2025.12/API_specification/generated/array_api.broadcast_shapes.html) function that broadcasts one or more *shapes* (as opposed to arrays) against one another.

The function was already present in the top-level `dpnp` namespace but missing from `dpnp.tensor`. This PR closes that gap.

### DIFF
--- a/dpnp/tensor/_manipulation_functions.py
+++ b/dpnp/tensor/_manipulation_functions.py
@@ -264,11 +264,14 @@ def broadcast_shapes(*shapes):
             shapes against one another.
 
     Raises:
+        TypeError: if a shape contains a non-integer dimension.
         ValueError: if the input shapes are not broadcast-compatible.
     """
     if len(shapes) == 0:
         return ()
-    return _broadcast_shape_impl([tuple(sh) for sh in shapes])
+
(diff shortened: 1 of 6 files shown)


dpnp/tensor/_manipulation_functions.py
```diff
@@ -248,6 +248,32 @@ def broadcast_arrays(*args):
...
+            shapes against one another.
+
+    Raises:
+        TypeError: if a shape contains a non-integer dimension.
+        ValueError: if the input shapes are not broadcast-compatible.
+    """
+    if len(shapes) == 0:
+        return ()
+
+    normalized = [tuple(operator.index(dim) for dim in sh) for sh in shapes]
```
ndgrigorian:
```suggestion
    normalized = [tuple(map(operator.index, sh)) for sh in shapes]
```
`map` is often a bit more efficient than a generator expression from my understanding

ndgrigorian (approved):
other than efficiency point LGTM

PR #3012, opened 2026-08-05
Link `dpnp_backend_c` against only the MKL SYCL domains it uses

antonwolfy:
`dpnp_backend_c` linked the `MKL::MKL_SYCL` umbrella target, whose link interface pulls in *every* MKL SYCL domain. This target is defined by MKL's `MKLConfig.cmake`, so the exact DLL set it drags in depends on the MKL build, not on dpnp.

This PR links the backend against only the MKL SYCL domains its sources use:
* **BLAS** — `oneapi::mkl::blas::scal`
* **RNG** — `oneapi::mkl::rng` engines/distributions and the VSL C API
* **VM** — `oneapi::mkl::vm` vector math (`div`, `sqr`, `add`, `powx`, `sqrt`, …)

`dft`, `lapack`, `sparse`, `stats`, and `data_fitting` are not referenced by any source compiled into `dpnp_backend_c`, so they are dropped from its link interface.


ndgrigorian (approved):
lgtm

PR #3014, opened 2026-08-06
Update implementation of `dpnp.putmask`

vlad-perevezentsev:
This PR proposes a new implementation of `dpnp.putmask` replacing the legacy `dpnp_putmask` implementation with dedicated SYCL kernels : a vectorized contiguous kernel and a strided kernel for F-contiguous/transposed arrays.
It also fully reworks the `putmask` tests by adding `TestPutMask`

Performance results on PVC are below:

image


dpnp/backend/extensions/indexing/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # *****************************************************************************
-# Copyright (c) 2025, Intel Corporation
+# Copyright (c) 2026, Intel Corporation
```
antonwolfy, line 3: It's year when file was created, no need to update

```suggestion
# Copyright (c) 2025, Intel Corporation
```

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,102 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
...
+    elif a.shape == values.shape:
+        a[mask] = values[mask]
+
+    else:
+        # numpy putmask cycles values by the C-order flat index of the
+        # destination (values.flat[c % N]), independent of memory layout, so
+        # values must always be flattened in C-order.
+        values_1d = values.ravel(order="C")
+        if a.dtype != values_1d.dtype:
+            values_1d = dpnp.astype(
```
antonwolfy, line 1886: Seems no test is covering that

dpnp/backend/extensions/indexing/indexing_py.cpp
```diff
@@ -1,5 +1,5 @@
 //*****************************************************************************
-// Copyright (c) 2025, Intel Corporation
+// Copyright (c) 2026, Intel Corporation
```
antonwolfy, line 2:
```suggestion
// Copyright (c) 2025, Intel Corporation
```

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,102 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
...
+    Examples
+    --------
+    >>> import dpnp as np
+    >>> x = np.arange(6).reshape(2, 3)
+    >>> np.putmask(x, x>2, x**2)
+    >>> x
+    array([[ 0,  1,  2],
+           [ 9, 16, 25]])
+
+    If `values` is smaller than `x1` it is repeated:
```
antonwolfy, line 1851:
```suggestion
    If `values` is smaller than `a` it is repeated:
```

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,102 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
...
+
+    See Also
+    --------
+    :obj:`dpnp.place` : Change elements of an array based on conditional and
+                        input values.
+    :obj:`dpnp.put` : Replaces specified elements of an array with given values.
+    :obj:`dpnp.take` : Take elements from an array along an axis.
+    :obj:`dpnp.copyto` : Copies values from one array to another.
+
+
```
antonwolfy, line 1841: Double blank like:
```suggestion
```

dpnp/dpnp_iface_indexing.py
```diff
@@ -1,5 +1,5 @@
 # *****************************************************************************
-# Copyright (c) 2016, Intel Corporation
+# Copyright (c) 2026, Intel Corporation
```
antonwolfy, line 2:
```suggestion
# Copyright (c) 2016, Intel Corporation
```

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,102 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
...
+    If `values` is not the same size as `a` and `mask` then it will
+    repeat. This gives behavior different from ``a[mask] = values``.
+
     For full documentation refer to :obj:`numpy.putmask`.
 
+    Parameters
+    ----------
+    a : {dpnp.ndarray, usm_ndarray}
+        Target array.
+    mask : {dpnp.ndarray, usm_ndarray}
```
antonwolfy, line 1822: Why do we limit only to supported array types?

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,102 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
...
+        Boolean mask array. It has to be the same shape as `a`.
+    values : {dpnp.ndarray, usm_ndarray, scalar}
```
antonwolfy, line 1824: And here the same

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,102 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
...
+    repeat. This gives behavior different from ``a[mask] = values``.
+
     For full documentation refer to :obj:`numpy.putmask`.
 
+    Parameters
+    ----------
+    a : {dpnp.ndarray, usm_ndarray}
+        Target array.
+    mask : {dpnp.ndarray, usm_ndarray}
+        Boolean mask array. It has to be the same shape as `a`.
```
antonwolfy, line 1823: In NumPy it actually has to have the same size, since supports array_like

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,102 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
...
+        )
+
+    elif a.shape == values.shape:
+        a[mask] = values[mask]
+
+    else:
+        # numpy putmask cycles values by the C-order flat index of the
+        # destination (values.flat[c % N]), independent of memory layout, so
+        # values must always be flattened in C-order.
+        values_1d = values.ravel(order="C")
```
antonwolfy, line 1884: `values` might be usm_ndarray without `ravel` method

dpnp/backend/kernels/indexing/putmask.hpp
```diff
@@ -0,0 +1,303 @@
...
+                          std::size_t values_size)
+        : dst_(dst), mask_u8_(reinterpret_cast<const std::uint8_t *>(mask)),
+          vals_(vals), indexer_(indexer), values_size_(values_size)
+    {
+    }
+
+    void operator()(sycl::id<1> wid) const
+    {
+        // empty `values` is a no-op (also guards the read below)
+        if (values_size_ == 0) {
```
antonwolfy, line 72: I wonder why do we have the guard inside the kernels now? We never did that for others.

dpnp/backend/kernels/indexing/putmask.hpp
```diff
@@ -0,0 +1,303 @@
...
+    const T *vals_ = nullptr;
+    const TwoOffsets_IndexerT indexer_;
+    const std::size_t values_size_ = 0;
+
+public:
+    PutMaskStridedFunctor(T *dst,
+                          const bool *mask,
+                          const T *vals,
+                          const TwoOffsets_IndexerT &indexer,
+                          std::size_t values_size)
```
antonwolfy, line 63: I believe sizes should be const here and after
```suggestion
                          const std::size_t values_size)
```

dpnp/tests/third_party/cupy/indexing_tests/test_insert.py
antonwolfy, line 221: Can be aligned now:
```suggestion
            with pytest.raises(ValueError):
```

dpnp/backend/extensions/indexing/putmask.cpp
```diff
@@ -0,0 +1,346 @@
...
+
+    dpnp::tensor::validation::AmpleMemory::throw_if_not_ample(dst, nelems);
+
+    char *dst_p = dst.get_data();
+    const char *mask_p = mask.get_data();
+    const char *values_p = values.get_data();
+    const std::size_t values_size = values.get_size();
+
+    // the contig kernel cycles `values` by the memory-linear index, which
+    // matches numpy's C-order `values.flat` only for C-contiguous data
```
antonwolfy, line 182: but no check here that `values.is_c_contiguous()` has to be true

dpnp/backend/kernels/indexing/putmask.hpp
```diff
@@ -0,0 +1,303 @@
...
+    const T *values_ = nullptr;
+    std::size_t nelems_ = 0;
+    std::size_t val_size_ = 0;
+
+public:
+    PutMaskContigFunctor(T *dst,
+                         const bool *mask,
+                         const T *values,
+                         std::size_t nelems,
+                         std::size_t val_size)
```
antonwolfy, line 107:
```suggestion
                         const std::size_t nelems,
                         const std::size_t val_size)
```

dpnp/backend/kernels/indexing/putmask.hpp
```diff
@@ -0,0 +1,303 @@
...
+        if constexpr (enable_sg_loadstore && !is_complex_v<T>) {
+            auto sg = ndit.get_sub_group();
+            const std::uint32_t sgSize = sg.get_max_local_range()[0];
+            const std::size_t lane_id = sg.get_local_id()[0];
+
+            const std::size_t base =
+                elems_per_wi * (ndit.get_group(0) * ndit.get_local_range(0) +
+                                sg.get_group_id()[0] * sgSize);
+
+            const bool values_no_repeat = (val_size_ >= nelems_);
```
antonwolfy, line 133: can be moved away of if-else block to reduce duplication

dpnp/backend/kernels/indexing/putmask.hpp
```diff
@@ -0,0 +1,303 @@
...
+                                idx + static_cast<std::size_t>(k) * sgSize;
+                            val_vec[k] = values_[g % val_size_];
+                        }
+                    }
+
+                    sycl::vec<T, vec_sz> out_vec;
+#pragma unroll
+                    for (std::uint8_t vec_id = 0; vec_id < vec_sz; ++vec_id) {
+                        out_vec[vec_id] =
+                            (mask_vec[vec_id] != static_cast<std::uint8_t>(0))
```
antonwolfy, line 178: for consistency with other places in the kernel:
```suggestion
                            (mask_vec[vec_id])
```

dpnp/backend/kernels/indexing/putmask.hpp
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        const std::size_t lin = wid.get(0);
+        auto offset = indexer_(static_cast<ssize_t>(lin));
+
+        const dpnp::tensor::ssize_t dst_off = offset.get_first_offset();
+        const dpnp::tensor::ssize_t mask_off = offset.get_second_offset();
+
+        if (mask_u8_[mask_off]) {
+            const std::size_t vlin = lin % values_size_;
```
antonwolfy: Do we need to align with contig kernel and use modulo operation only when it's really needed?

```suggestion
            const std::size_t vlin = values_no_repeat_ ? lin : (lin % values_size_);
```

dpnp/backend/kernels/indexing/putmask.hpp
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    void operator()(sycl::id<1> wid) const
+    {
+        // empty `values` is a no-op (also guards the read below)
+        if (values_size_ == 0) {
+            return;
+        }
+
+        const std::size_t lin = wid.get(0);
```
ndgrigorian, line 76:
```suggestion
        const std::size_t lin = wid[0];
```
Matched tensor stylistically

dpnp/backend/kernels/indexing/putmask.hpp
vlad-perevezentsev: A good point
Updated this logic, thanks

dpnp/backend/kernels/indexing/putmask.hpp
vlad-perevezentsev: The best place for this would be in putmask dispatch (putmask.cpp)
Fixed it, thanks!

dpnp/dpnp_iface_indexing.py
vlad-perevezentsev: I have added a new tests
Thanks

dpnp/backend/extensions/indexing/putmask.cpp
vlad-perevezentsev: A good point
Fixed, thanks!

dpnp/dpnp_iface_indexing.py
vlad-perevezentsev: This work was left unfinished because I initially focused on implementing an optimized kernel
I have added support for array-like `mask` and `values` in https://github.com/IntelPython/dpnp/pull/3014/commits/6a37d803472b14bee092ebaa27b1d1d943a69e9f

dpnp/dpnp_iface_indexing.py
vlad-perevezentsev: It is a great catch!
The missing tests for `usm_ndarray` revealed that the current implementation of `dpnp.putmask` did not support `usm_ndarray`.
The changes in https://github.com/IntelPython/dpnp/pull/3014/changes/43e6e37db978be2190dea166dea928bb22e1093f  propose operating on `usm_ndarray` which is similar in style to `dpnp.place`
Also I have added a new tests for that case

dpnp/dpnp_iface_indexing.py
vlad-perevezentsev: The current behavior matches [cupy ](https://github.com/cupy/cupy/blob/v14.2.0/cupy/_indexing/insert.py#L123)
A `.size` check could replace it but all three branches still need the same shape:
```
usm_a[usm_mask] = values
usm_a[usm_mask] = usm_values[usm_mask]
indexing_ext._putmask(usm_a, usm_mask, usm_values_1d, exec_q, dep_evs)
```
boolean indexing needs `mask.shape == a.shape`
Kernel packs one common shape for `dst/mask`

I could reshape `mask` to `a.shape` (`usm_mask = dpt.reshape(usm_mask, usm_a.shape)`) and remove the shape check to match numpy but reshape copies in some edge cases so I kept the shape check for this implementation

antonwolfy: Ok, let's keep the same shape check.

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,109 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
...
+        usm_type=usm_a.usm_type,
+        sycl_queue=usm_a.sycl_queue,
     )
-    mask_desc = dpnp.get_dpnp_descriptor(mask, copy_when_nondefault_queue=False)
-    values_desc = dpnp.get_dpnp_descriptor(
-        values, copy_when_nondefault_queue=False
+    usm_mask = dpt.astype(usm_mask, dpnp.bool, copy=False)
+
+    if usm_a.shape != usm_mask.shape:
+        raise ValueError("mask and data must be the same size")
```
antonwolfy:
```suggestion
        raise ValueError("mask and data must be the same shape")
```

dpnp/tests/test_indexing.py
```diff
@@ -1169,126 +1169,173 @@ def test_indices(dimension, dtype, sparse):
...
-    a = numpy.array(arr)
-    ia = dpnp.array(a)
-    m = numpy.array(mask)
-    im = dpnp.array(m)
-    v = numpy.array([100, 200])
-    iv = dpnp.array(v)
-    numpy.putmask(a, m, v)
-    dpnp.putmask(ia, im, iv)
-    assert_array_equal(a, ia)
+class TestPutMask:
```
antonwolfy, line 1172: Missing cross-queue / usm_type propagation tests in test_sycl_queue.py / test_usm_type.py

antonwolfy, line 1172: The vectorized fast path is still untested. The contig vectorized branch needs `nelems ≳ 128–256`; every array in TestPutMask is ≤ 24 elements.

dpnp/tests/test_indexing.py
```diff
@@ -1169,126 +1169,173 @@ def test_indices(dimension, dtype, sparse):
...
+        mask = generate_random_numpy_array(a.shape, dtype=dpnp.bool)
+        vals = generate_random_numpy_array((3,), dtype=dt)
+        imask, ivals = dpnp.array(mask), dpnp.array(vals)
+
+        numpy.putmask(a, mask, vals)
+        dpnp.putmask(ia, imask, ivals)
+        assert_array_equal(ia, a)
+
+    @pytest.mark.parametrize("dt", get_all_dtypes(no_none=True))
+    def test_scalar_values(self, dt):
```
antonwolfy, line 1271: The scalar unsafe-cast path is untested. Needs to cover also unsafe scalar truncation/wraparound.

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,109 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
...
+    >>> x = np.arange(5)
+    >>> np.putmask(x, x>1, np.array([-33, -44]))
+    >>> x
+    array([  0,   1, -33, -44, -33])
 
     """
 
-    x1_desc = dpnp.get_dpnp_descriptor(
-        x1, copy_when_strides=False, copy_when_nondefault_queue=False
+    dpnp.check_supported_arrays_type(a)
```
antonwolfy: Not needed, the same check is done by `dpnp.get_usm_ndarray`

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,109 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
...
+        )
+
+    if usm_a.shape == usm_values.shape:
+        usm_a[usm_mask] = usm_values[usm_mask]
+        return
+
+    # numpy putmask cycles values by the C-order flat index of the
+    # destination (values.flat[c % N]), independent of memory layout, so
+    # values must always be flattened in C-order.
+    usm_values_1d = dpnp.get_usm_ndarray(dpnp.ravel(usm_values, order="C"))
```
antonwolfy:
```suggestion
    usm_values_1d = dpnp.ravel(usm_values, order="C").get_array()
```

dpnp/backend/extensions/indexing/putmask.cpp
```diff
@@ -0,0 +1,351 @@
...
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+//*****************************************************************************
+
+#include <algorithm>
```
antonwolfy, line 29: missing `<cstdint>`

dpnp/backend/extensions/indexing/putmask.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    check_same_size({&dst, &mask}, names);
+    const int nd = dst.get_ndim();
+    check_num_dims({&mask}, nd, names);
+
+    check_queue({&dst, &mask, &values}, names, exec_q);
+    check_writable({&dst}, names);
+
+    // values must be C-contiguous
+    check_c_contig({&values}, names);
```
antonwolfy, line 151: Missing overlap check

dpnp/backend/extensions/indexing/putmask.cpp
```diff
@@ -0,0 +1,351 @@
...
+    dpnp::tensor::validation::AmpleMemory::throw_if_not_ample(dst, nelems);
+
+    char *dst_p = dst.get_data();
+    const char *mask_p = mask.get_data();
+    const char *values_p = values.get_data();
+
+    // the contig kernel cycles `values` by the memory-linear index, which
+    // matches numpy's C-order `values.flat` only for C-contiguous data
+    const bool all_c_contig = dst.is_c_contiguous() && mask.is_c_contiguous() &&
+                              values.is_c_contiguous();
```
antonwolfy: Do we need to remove `values.is_c_contiguous()` from here?

dpnp/backend/extensions/indexing/putmask.cpp
```diff
@@ -0,0 +1,351 @@
...
+}
+
+/**
+ * @brief A factory to define pairs of supported types for which
+ * putmask function is available.
+ *
+ * @tparam T Type of input vector `dst` and `values` and of result vector `dst`.
+ */
+template <typename T>
+struct PutMaskOutputType
```
antonwolfy: It does not seem we need that, and so can be dropped.
Similarly to `dpnp/tensor/libtensor/include/kernels/where.hpp`:
```cpp
template <typename fnT, typename T, typename condT>
struct WhereStridedFactory
{
    fnT get()
    {
        fnT fn = where_strided_impl<T, condT>;
        return fn;
    }
};

template <typename fnT, typename T, typename condT>
struct WhereContigFactory
{
    fnT get()
    {
        fnT fn = where_contig_impl<T, condT>;
        return fn;
    }
};
```

dpnp/backend/extensions/indexing/putmask.cpp
```diff
@@ -0,0 +1,351 @@
...
+    const auto &mask_strides = mask.get_strides_vector();
+
+    using shT = std::vector<py::ssize_t>;
+    shT common_shape;
+    shT s_dst_strides;
+    shT s_mask_strides;
+
+    int eff_nd = nd;
+    if (nd == 0) {
+        // scalar arrays: single-element 1D iteration
```
antonwolfy: It seems unreachable defensive code (0-d dst → contig path).

dpnp/backend/extensions/indexing/putmask.cpp
vlad-perevezentsev: This is not needed if I delete `PutMaskOutputType`

dpnp/backend/extensions/indexing/putmask.cpp
vlad-perevezentsev: Yes, fixed
Thank you

dpnp/backend/extensions/indexing/putmask.cpp
vlad-perevezentsev: Good catch
I have removed `PutMaskOutputType` there

dpnp/backend/extensions/indexing/putmask.cpp
vlad-perevezentsev: Good catch 
Simplified this logic
Thanks

dpnp/backend/extensions/indexing/putmask.cpp
vlad-perevezentsev: Done, thanks

dpnp/dpnp_iface_indexing.py
vlad-perevezentsev: I have removed it
Thanks

dpnp/tests/test_indexing.py
vlad-perevezentsev: Good catch
I have added a new test for that

dpnp/tests/test_indexing.py
vlad-perevezentsev: Good catch
I have added a new test for that

dpnp/tests/test_indexing.py
vlad-perevezentsev: Added
Thanks!

dpnp/backend/extensions/indexing/putmask.cpp
```diff
@@ -0,0 +1,297 @@
...
+    const sycl::event &cpy_ev = std::get<2>(pack);
+
+    std::vector<sycl::event> all_deps = depends;
+    all_deps.push_back(cpy_ev);
+
+    sycl::event comp_ev =
+        strided_fn(exec_q, nd, nelems, shape_strides_dev, dst_p, dst_off,
+                   mask_p, mask_off, values_p, values_size, all_deps);
+
+    sycl::event cleanup_ev = dpnp::tensor::alloc_utils::async_smart_free(
```
antonwolfy, line 243: Missing `"utils/sycl_alloc_utils.hpp"` include

dpnp/backend/extensions/indexing/putmask.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    // the contig kernel cycles `values` by the memory-linear index, which
+    // matches numpy's C-order `values.flat` only for C-contiguous data
+    // (`values` is already checked to be C-contiguous above)
+    const bool all_c_contig = dst.is_c_contiguous() && mask.is_c_contiguous();
+
+    if (all_c_contig) {
+        auto contig_fn = putmask_contig_dispatch_vector[dst_values_typeid];
+
+        if (contig_fn == nullptr) {
```
antonwolfy: The guard is now dead, can be removed and aligned with where/clip

dpnp/backend/extensions/indexing/putmask.cpp
```diff
@@ -0,0 +1,297 @@
...
+    shT common_shape(dst_shape, dst_shape + nd);
+    shT s_dst_strides = dst_strides;
+    shT s_mask_strides = mask_strides;
+
+    // trivial offsets: shape and strides are passed without simplification
+    constexpr py::ssize_t dst_off = 0;
+    constexpr py::ssize_t mask_off = 0;
+
+    auto strided_fn = putmask_strided_dispatch_vector[dst_values_typeid];
+    if (strided_fn == nullptr) {
```
antonwolfy: Same dead guard

dpnp/tests/test_indexing.py
```diff
@@ -1169,126 +1169,196 @@ def test_indices(dimension, dtype, sparse):
...
-    a = numpy.array(arr)
-    ia = dpnp.array(a)
-    m = numpy.array(mask)
-    im = dpnp.array(m)
-    v = numpy.array([100, 200])
-    iv = dpnp.array(v)
-    numpy.putmask(a, m, v)
-    dpnp.putmask(ia, im, iv)
-    assert_array_equal(a, ia)
+class TestPutMask:
```
antonwolfy, line 1172: Test nice-to-haves: integer-overflow scalar cast (300 → int8), a 0-d-values unsafe-cast case, and a large strided repeat test.

dpnp/backend/extensions/indexing/putmask.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    for (int i = 0; i < std::max(nd, 1); ++i) {
+        const py::ssize_t d = (nd == 0 ? 1 : dst_shape[i]);
+        const py::ssize_t m = (nd == 0 ? 1 : mask_shape[i]);
+        nelems *= static_cast<std::size_t>(d);
+        shapes_equal = shapes_equal && (d == m);
+    }
+    if (!shapes_equal) {
+        throw py::value_error("`mask` and `dst` shapes must match");
+    }
```
antonwolfy, line 175: We probably need to decouple shape equality check to separate validation util and to remove `check_same_size({&dst, &mask}, names)` from above, since it's redundant (shape-equality is stricter then shape-equality) in the follow-up PR.

dpnp/tests/test_sycl_queue.py
```diff
@@ -1535,6 +1535,17 @@ def test_choose(device):
     assert_sycl_queue_equal(result.sycl_queue, chc.sycl_queue)
 
 
+@pytest.mark.parametrize("device", valid_dev, ids=dev_ids)
+def test_putmask(device):
```
antonwolfy: I was wrong, there is nothing to check from SYCL queue and USM type, there is no output array

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,107 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
     dpt.put_along_axis(usm_a, usm_ind, usm_vals, axis=axis, mode=mode)
 
 
-def putmask(x1, mask, values):
+def putmask(a, /, mask, values):
```
antonwolfy: The below example behaves differently comparing to NumPy:
```python
import numpy as np
import dpnp

# A NON-CANONICAL bool mask: reinterpret int8 bytes that aren't all 0/1 as bool.
# (Masks from comparisons like `a > x` are always canonical 0/1 — you only reach
#  this by deliberately viewing non-0/1 bytes as bool.)
raw  = np.array([0, 5, 0, 2, 0, 7, 0, 1], dtype=np.int8)
mask = raw.view(np.bool_)          # stored bytes: [0,5,0,2,0,7,0,1]

# NumPy: any NONZERO byte is True  -> sets positions 1,3,5,7
a_np = np.arange(8)
np.putmask(a_np, mask, 99)
print("numpy:", a_np.tolist())     # [0, 99, 2, 99, 4, 99, 6, 99]

# dpnp (PR): the bool mask passes through unchanged (astype(bool, copy=False)),
# and the scalar path uses dpt boolean-index assignment, which honors only byte==1
a_dp = dpnp.arange(8)
dpnp.putmask(a_dp, dpnp.asarray(mask), 99)
print("dpnp :", dpnp.asnumpy(a_dp).tolist())   # [0, 1, 2, 3, 4, 5, 6, 99]
```

dpnp/dpnp_iface_indexing.py
```diff
@@ -1807,30 +1803,107 @@ def put_along_axis(a, ind, values, axis, mode="wrap"):
...
+    usm_mask = dpt.astype(usm_mask, dpnp.bool, copy=False)
+
+    if usm_a.shape != usm_mask.shape:
+        raise ValueError("mask and data must be the same shape")
+
+    if dpnp.isscalar(values):
+        usm_a[usm_mask] = values
+        return
+
+    usm_values = dpnp.as_usm_ndarray(
```
antonwolfy, line 1874: Do we need to pass `dtype=usm_a.dtype` here? Otherwise:
```python
import dpnp as np

a = np.ones(2, dtype='i2')
np.putmask(a, np.ones_like(a), [2.0, 1.5])
Cell In[19], line 1
----> 1 np.putmask(a, np.ones_like(a), [2.0, 1.5])

File /localdisk/work/antonvol/code/dpnp/dpnp/dpnp_iface_indexing.py:1881, in putmask(a, mask, values)
   1874 usm_values = dpnp.as_usm_ndarray(
   1875     values,
   1876     usm_type=usm_a.usm_type,
   1877     sycl_queue=usm_a.sycl_queue,
   1878 )
   1880 if not dpt.can_cast(usm_values.dtype, usm_a.dtype):
-> 1881     raise TypeError(
   1882         f"Cannot cast array data from {usm_values.dtype} to {usm_a.dtype} "
   1883         "according to the rule 'safe'"
   1884     )
   1886 if usm_a.shape == usm_values.shape:
   1887     usm_a[usm_mask] = usm_values[usm_mask]

TypeError: Cannot cast array data from float64 to int16 according to the rule 'safe'
```
but passed with numpy:
```python
a = np.ones(2, dtype='i2')
np.putmask(a, np.ones_like(a), [2.0, 1.5])
a
# Out: array([2, 1], dtype=int16)
```

vlad-perevezentsev: Good catch
Fixed, thanks

dpnp/backend/extensions/indexing/putmask.cpp
vlad-perevezentsev: I have added it
thanks

dpnp/tests/test_indexing.py
vlad-perevezentsev: I extended tests for putamask
Thanks

dpnp/backend/extensions/indexing/putmask.cpp
vlad-perevezentsev: I added TODO comment so I do not forget 
Thanks

ndgrigorian (approved):
In general this LGTM, thank you @vlad-perevezentsev

PR #3015, opened 2026-08-07
Use pybind11 `3.1.0` to build dpnp

antonwolfy:
The PR updates `CMakeLists.txt` to bump pybind11 `3.1.0` up from `3.0.4` version.

The example under `examples/pybind11/use_dpnp_array` is updated to the same version to keep both `FetchContent` declarations in sync.


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #3020, opened 2026-08-12
Add test coverage for the reshape copy keyword

antonwolfy:
This PR adds test coverage for the `copy` keyword of `reshape`/`ndarray.reshape` to the third-party `manipulation_tests/test_shape.py` suite.


ndgrigorian (approved):
only thing is to add to changelog

antonwolfy:
> only thing is to add to changelog

@ndgrigorian , to be honest, I don't think we need that for only tests update, there is no user facing issue.

vlad-perevezentsev (approved):

ndgrigorian:
> > only thing is to add to changelog
> 
> @ndgrigorian , to be honest, I don't think we need that for only tests update, there is no user facing issue.

Fair point!

PR #3021, opened 2026-08-12
Fix abort in `dpnp.all` / `dpnp.any` over an empty axis

antonwolfy:
`dpnp.all` and `dpnp.any` aborted the process when reducing over an empty axis, e.g.:

```python
import dpnp
dpnp.all(dpnp.zeros((0, 3, 4)))          # reduces over 0 elements
dpnp.any(dpnp.zeros((0, 3, 4)), axis=0)  # zero-length reduced axis
```

`_boolean_reduction` always allocated a temporary and launched the reduction kernel. When the reduced extent is zero (a reduced dimension has length 0), the kernel is submitted with a zero-sized `nd_range`. That is a silent no-op on a SYCL runtime built with `NDEBUG`, but aborts on an assertions-enabled runtime, where `adjustNDRangePerKernel` asserts `NDR.LocalSize[0] == 0` whenever `GlobalSize[0]` is `0`:

```
Assertion `NDR.LocalSize[0] == 0' failed.
  .../sycl/source/detail/scheduler/commands.cpp
  void sycl::_V1::detail::adjustNDRangePerKernel(...)
```

This PR proposes to short-circuit in `_boolean_reduction` when the (permuted) input has no elements, building the result directly with the reduction identity instead of launching a kernel. `all` uses identity `True`, `any` uses `False`. A single `x_tmp.size == 0` check covers both empty sub-cases correctly:

- empty output (a non-reduced dimension is zero) — the result is an empty array, so the fill value is irrelevant;
- zero-length reduced axis (the output is non-empty) — the reduction identity is the answer, e.g. `all([]) is True`, `any([]) is False`.

Note, the issue was identified when building with the nightly LLVM SYCL compiler and [running](https://github.com/antonwolfy/dpnp/actions/runs/31585418522/job/94078011200) dpnp tests.
And the tests are [passing](https://github.com/antonwolfy/dpnp/actions/runs/31595647966/job/94110377635) now without any crash.


ndgrigorian:
@antonwolfy 
We could also consider adding an early-exit path to the reduction pybind11 code. In fact, one should already exist, maybe it was missed for boolean reductions?

antonwolfy:
I was thinking about that.. and it should be added to `py_boolean_reduction`, something like:
```cpp
if (dst_nelems == 0) {
    // empty result: nothing to write
    return std::make_pair(sycl::event(), sycl::event());
}
if (red_nelems == 0) {
    // empty reduction extent: result is the op identity, which this kernel
    // cannot produce; the caller must handle it
    throw py::value_error(
        "Reduction over an empty axis is not supported");
}
```
but I wondering why we already had in the code a hadnling depending on `dst_nelems == 0`:
```cpp
    // TODO: should be dst_nelems == 0?
    if ((is_src_c_contig && is_dst_c_contig) ||
        (is_src_f_contig && dst_nelems == 0)) {
```

@ndgrigorian, if you remember why it was added this way and can confirm it's not needed, I'll add the above handling to `py_boolean_reduction` and will remove the TODO and obsolete path with `dst_nelems == 0`.

ndgrigorian:
@antonwolfy 
I can't recall why the comment was added and honestly, I don't see a good reason not to just change the behavior as like you said, the path is obsolete

antonwolfy:
@ndgrigorian, I updated the implementation and it's ready for review now

dpnp/tensor/_utility_functions.py
```diff
@@ -77,37 +77,48 @@ def _boolean_reduction(x, axis, keepdims, func):
...
-        sycl_queue=exec_q,
-    )
-    hev1, ev1 = ti._copy_usm_ndarray_into_usm_ndarray(
-        src=res_tmp, dst=res, sycl_queue=exec_q, depends=[ev0]
-    )
-    _manager.add_event_pair(hev1, ev1)
+    if x_tmp.size == 0:
+        # nothing to reduce over: the result is either empty (a non-reduced
+        # dimension is zero) or filled with the reduction identity (a reduced
+        # dimension is zero, e.g. all([]) is True and any([]) is False)
```
ndgrigorian: I think we can make this very verbose comment a bit more brief

ndgrigorian (approved):
Other than this nit LGTM

ndgrigorian (approved):

PR #3024, opened 2026-08-13
Add proper exception handling for nested lists in `dpnp.repeat`

### LINKED ISSUE #2972: Add a proper exception handling for nested lists in `dpnp.repeat`
```python
import dpnp as np

np.__version__
# Out: '0.21.0dev1+27.gfdb1c53d735.dirty'

x = np.array([3])
np.repeat(x, [[4]])
Cell In[20], line 1
----> 1 np.repeat(x, [[4]])

File ~/code/dpnp/dpnp/dpnp_iface_manipulation.py:2841, in repeat(a, repeats, axis)
   2838     a = dpnp.ravel(a)
   2840 usm_arr = dpnp.get_usm_ndarray(a)
-> 2841 usm_res = dpt.repeat(usm_arr, repeats, axis=axis)
   2842 return dpnp_array._create_from_usm_ndarray(usm_res)

File ~/code/dpnp/dpnp/tensor/_manipulation_functions.py:631, in repeat(x, repeats, axis)
    629 if len_reps == 1:
    630     repeats = repeats[0]
--> 631     if repeats < 0:
    632         raise ValueError("`repeats` elements must be positive")
    633     scalar = True

TypeError: '<' not supported between instances of 'list' and 'int'
```
while NumPy raising more explicit exception:
```python
np.__version__
# Out: '2.5.0'

np.repeat(x, [[4]])
Cell In[24], line 1
----> 1 np.repeat(x, [[4]])

File ~/.local/share/mamba/envs/dpnp_dev/lib/python3.13/site-packages/numpy/_core/fromnumeric.py:481, in repeat(a, repeats, axis)
    437 @array_function_dispatch(_repeat_dispatcher)
    438 def repeat(a, repeats, axis=None):
    439     """
    440     Repeat each element of an array after themselves
    441
   (...)    479
    480     """
--> 481     return _wrapfunc(a, 'repeat', repeats, axis=axis)

File ~/.local/share/mamba/envs/dpnp_dev/lib/python3.13/site-packages/numpy/_core/fromnumeric.py:54, in _wrapfunc(obj, method, *args, **kwds)
     51     return _wrapit(obj, method, *args, **kwds)
     53 try:
---> 54     return bound(*args, **kwds)
     55 except TypeError:
     56     # A TypeError occurs if the object does have such a method in its
     57     # class, but its signature is not identical to that of NumPy's. This
   (...)     61     # Call _wrapit from within the except clause to ensure a potential
     62     # exception has a traceback chain.
     63     return _wrapit(obj, method, *args, **kwds)

ValueError: setting an array element with a sequence. The requested array would exceed the maximum number of dimension of 1.
```

antonwolfy:
This PR fixes #2972.

Passing a nested sequence as `repeats` to `dpnp.repeat` (e.g. `np.repeat(x, [[4]])`) raised an unclear `TypeError` instead of a meaningful error:

```python
>>> import dpnp as np
>>> x = np.array([3])
>>> np.repeat(x, [[4]])
TypeError: '<' not supported between instances of 'list' and 'int'
```

The sequence branch of `dpnp.tensor.repeat` special-cased a length-1 sequence by taking `repeats[0]` and comparing it to `0`, which fails when that element is itself a list. It also had no dimensionality guard, so a multi-element nested sequence could slip through as a 2-D array — unlike the `usm_ndarray` branch, which already rejects `ndim > 1`.

The fix converts the sequence to an array up front and rejects a dimensionality greater than 1, aligning with NumPy, which raises a `ValueError` in this case:

```python
>>> np.repeat(x, [[4]])
ValueError: `repeats` sequence must be 0- or 1-dimensional, got 2 dimensions
```

The scalar fast path (`_repeat_by_scalar`) is preserved for the size-1 case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@ This release is compatible with NumPy 2.5.
 * Fixed missing strides validation in `dpnp.tensor.usm_ndarray` constructor when allocating new memory [#2927](https://github.com/IntelPython/dpnp/pull/2927)
 * Fixed `dpnp.bincount` raising a `ValueError` on an empty input array instead of returning an empty `intp` array [#3018](https://github.com/IntelPython/dpnp/pull/3018)
 * Fixed `dpnp.tensor.top_k` aborting for `k=0` by returning empty result arrays without launching a zero-sized kernel [#3022](https://github.com/IntelPython/dpnp/pull/3022)
+* Fixed `dpnp.repeat` raising an unclear `TypeError` for a nested sequence of `repeats` [#3024](https://github.com/IntelPython/dpnp/issues/3024)
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,7 +79,7 @@ This release is compatible with NumPy 2.5.
 * Fixed missing strides validation in `dpnp.tensor.usm_ndarray` constructor when allocating new memory [#2927](https://github.com/IntelPython/dpnp/pull/2927)
 * Fixed `dpnp.bincount` raising a `ValueError` on an empty input array instead of returning an empty `intp` array [#3018](https://github.com/IntelPython/dpnp/pull/3018)
 * Fixed `dpnp.tensor.top_k` aborting for `k=0` by returning empty result arrays without launching a zero-sized kernel [#3022](https://github.com/IntelPython/dpnp/pull/3022)
-* Fixed `dpnp.repeat` raising an unclear `TypeError` for a nested sequence of `repeats` [#3024](https://github.com/IntelPython/dpnp/issues/3024)
+* Fixed `dpnp.repeat` raising an unclear `TypeError` for a nested sequence of `repeats` [#3024](https://github.com/IntelPython/dpnp/pull/3024)
 
 ### Security
 
--- a/dpnp/tensor/_manipulation_functions.py
+++ b/dpnp/tensor/_manipulation_functions.py
@@ -663,21 +663,25 @@ def repeat(x, repeats, /, *, axis=None):
         usm_type = x.usm_type
         exec_q = x.sycl_queue
 
-        len_reps = len(repeats)
-        if len_reps == 1:
-            repeats = repeats[0]
+        repeats = dpt.asarray(
+            repeats, dtype=dpt.int64, usm_type=usm_type, sycl_queue=exec_q
+        )
+        if repeats.ndim > 1:
+            raise ValueError(
+                "`repeats` sequence must be 0- or 1-dimensional, got "
+                f"{repeats.ndim} dimensions"
+            )
+        if repeats.size == 1:
+            scalar = True
+            repeats = int(repeats[0])
             if repeats < 0:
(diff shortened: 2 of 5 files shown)


dpnp/tensor/_manipulation_functions.py
```diff
@@ -663,21 +663,25 @@ def repeat(x, repeats, /, *, axis=None):
         usm_type = x.usm_type
         exec_q = x.sycl_queue
 
-        len_reps = len(repeats)
-        if len_reps == 1:
-            repeats = repeats[0]
+        repeats = dpt.asarray(
```
ndgrigorian: One reason for this fast path was to avoid this situation of moving the input to the device. I think it would be worthwhile to preserve that, if possible. The scalar path was significantly faster because we never call `asarray`

ndgrigorian (commented):
Also @antonwolfy does `repeat` docstring need to change?

CHANGELOG.md
```diff
@@ -79,6 +79,7 @@ This release is compatible with NumPy 2.5.
 * Fixed missing strides validation in `dpnp.tensor.usm_ndarray` constructor when allocating new memory [#2927](https://github.com/IntelPython/dpnp/pull/2927)
 * Fixed `dpnp.bincount` raising a `ValueError` on an empty input array instead of returning an empty `intp` array [#3018](https://github.com/IntelPython/dpnp/pull/3018)
 * Fixed `dpnp.tensor.top_k` aborting for `k=0` by returning empty result arrays without launching a zero-sized kernel [#3022](https://github.com/IntelPython/dpnp/pull/3022)
+* Fixed `dpnp.repeat` raising an unclear `TypeError` for a nested sequence of `repeats` [#3024](https://github.com/IntelPython/dpnp/issues/3024)
```
vlad-perevezentsev:
```suggestion
* Fixed `dpnp.repeat` raising an unclear `TypeError` for a nested sequence of `repeats` [#3024](https://github.com/IntelPython/dpnp/pull/3024)
```

dpnp/tests/tensor/test_usm_ndarray_manipulation.py
```diff
@@ -1464,6 +1464,12 @@ def test_repeat_arg_validation():
     with pytest.raises(ValueError):
         dpt.repeat(x, dpt.ones((1, 1), dtype="i8"))
 
+    # repeats nested sequence must be 0d or 1d
+    with pytest.raises(ValueError, match="0- or 1-dimensional"):
+        dpt.repeat(x, [[4]])
+    with pytest.raises(ValueError, match="0- or 1-dimensional"):
```
vlad-perevezentsev, line 1470: It would be good to add this case as well  dpt.repeat(x, [[1],[2].[3],])

dpnp/tensor/_manipulation_functions.py
vlad-perevezentsev: Agree with @ndgrigorian
We can use np.asarray() here and revert dpt.asarray in else branch

ndgrigorian (approved):
In general, this LGTM

PR #3026, opened 2026-08-13
Enable free-threaded Python for dpnp

ndgrigorian:
This PR removes `python-gil` as a dependency for dpnp and enables free-threaded builds, which are now run in public CI as part of the conda-package workflow.

Changes were made in particular to `dpnp_capi` aligning with similar changes in dpctl, print options which now align with NumPy's `ContextVar` approach, dpnp verbosity mode, and `dpnp.random` module, which has introduced locks for some state

### DIFF
--- a/.github/workflows/conda-package.yml
+++ b/.github/workflows/conda-package.yml
@@ -29,8 +29,22 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.10', '3.11', '3.12', '3.13', '3.14']
+        python: ['3.10', '3.11', '3.12', '3.13']
         os: [ubuntu-22.04, windows-2022]
+        python_spec: ['']
+        include:
+          - python: '3.14'
+            os: ubuntu-22.04
+            python_spec: '3.14.* *_cp314'
+          - python: '3.14'
+            os: windows-2022
+            python_spec: '3.14.* *_cp314'
+          - python: '3.14'
+            os: ubuntu-22.04
+            python_spec: '3.14.* *_cp314t'
+          - python: '3.14'
+            os: windows-2022
+            python_spec: '3.14.* *_cp314t'
 
     permissions:
       # Needed to cancel any previous runs that are not completed for a given workflow
@@ -46,6 +60,8 @@ jobs:
     env:
       build-conda-pkg-env: 'environments/build_conda_pkg.yml'
       build-env-name: 'build'
+      python-label: ${{ contains(matrix.python_spec, 'cp314t') && '3.14t' || matrix.python }}
+      python-conda-spec: ${{ matrix.python_spec || matrix.python }}
 
     steps:
       - name: Cancel Previous Runs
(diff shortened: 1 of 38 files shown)


dpnp/include/dpnp4pybind11.hpp
```diff
@@ -38,6 +38,7 @@
 #include "usm_ndarray_constants.h"
 
 #include <array>
+#include <atomic>
```
antonwolfy, line 41: Missing `#include <mutex>`

dpnp/random/dpnp_random_state.py
antonwolfy, line 186: Per my understanding, it hands back the live `MT19937`/`MCG59` engine object whose `.normal()`/`.uniform()` `cpdefs mutate engine state in place.
A caller using the returned engine concurrently with `normal`/`uniform` would race on the per-instance engine.
While NumPy returns a plain-data snapshot, not the live engine.

.github/workflows/conda-package.yml
```diff
@@ -29,8 +29,22 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.10', '3.11', '3.12', '3.13', '3.14']
+        python: ['3.10', '3.11', '3.12', '3.13']
         os: [ubuntu-22.04, windows-2022]
+        python_spec: ['']
+        include:
+          - python: '3.14'
```
antonwolfy, line 35: We probably need to extend `.github/workflows/check-onemath.yaml` and `.github/workflows/cron-run-tests.yaml`

antonwolfy, line 35: Please populate the changelog

dpnp/random/dpnp_random_state.py
ndgrigorian: I see, I've moved the lock into `dpnp_algo_random.pyx` instead

.github/workflows/conda-package.yml
```diff
@@ -46,6 +60,8 @@ jobs:
     env:
       build-conda-pkg-env: 'environments/build_conda_pkg.yml'
       build-env-name: 'build'
+      python-label: ${{ contains(matrix.python_spec, 'cp314t') && '3.14t' || matrix.python }}
```
antonwolfy: It's probably better to follow the same pattern as done for `.github/workflows/check-onemath.yaml`, because list of python versions will be extended in the future
```suggestion
      python-label: ${{ endsWith(matrix.python_spec, 't') && format('{0}t', matrix.python) || matrix.python }}
```

.github/workflows/cron-run-tests.yaml
```diff
@@ -38,8 +38,32 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: ['3.10', '3.11', '3.12', '3.13', '3.14']
+        python: ['3.10', '3.11', '3.12', '3.13']
         runner: [ubuntu-22.04, ubuntu-24.04, windows-2022]
+        python_spec: ['']
```
antonwolfy: It's safe to remove, otherwise the job name has trailing empty component, like: `Build (3.10, ubuntu-22.04, )`
```suggestion
```

.github/workflows/conda-package.yml
antonwolfy: still missing

examples/example_freethreading.py
```diff
@@ -0,0 +1,285 @@
...
+
+Every result is checked against the same computation run serially, since
+threading is only useful here if it does not change the answers.
+
+The examples work on a GIL-enabled build too, they just do not overlap as much,
+as only one thread runs Python code at a time there.
+
+Run it with::
+
+    python example_freethreading.py
```
antonwolfy: Do we need to extend `Test examples` job and to run the new example there?

ndgrigorian: oops, no, I don't want this example in the PR yet, I need to remake it, it was an experiment to see what AI would produce, I was not particularly happy with it

ndgrigorian: my intention is to add an example in a follow up for now

antonwolfy (approved):
Thank you @ndgrigorian, no more comments from me

PR #2850, opened 2026-04-13
Release GIL when calling `mkl_lapack::orgqr`

antonwolfy:
This PR fixes a deadlock in QR decomposition tests by adding GIL release before `mkl_lapack::orgqr` call.

The hang occurred because:
1. dpnp/dpctl submits a `host_task` to manage Python object lifetimes
2. host_task needs to acquire the GIL to decrement reference counts
3. if the main thread holds the GIL during queue submission → deadlock
4. `orgqr` is currently implemented in oneMKL as GPU-to-Host reverse offload:
```cpp
exec_q.submit([&](sycl::handler& cgh) {
  cgh.depends_on(depends);
   cgh.host_task([=]() { orgqr_host(...); });
}).wait();
```

As a solution PR proposes to release GIL using before calling the OneMKL operations. The GIL is automatically reacquired when the function returns (RAII).

- [x] Have you provided a meaningful PR description?
- [ ] Have you added a test, reproducer or referred to an issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [x] Have you checked performance impact of proposed changes?
- [ ] Have you added documentation for your changes, if necessary?
- [x] Have you added your changes to the changelog?

PR #3027, opened 2026-08-14
Release GIL around blocking oneMKL calls in BLAS/LAPACK

abagusetty:
potrf, getrf, geqrf, gesv, gesvd, heevd, syevd, ungqr and the BLAS dot/gemm family block the calling thread for the whole oneMKL call while holding the GIL. That stalls every other Python thread, including the host tasks dpnp queues to manage object lifetimes, which reacquire the GIL to drop references.

Completes the backport of #2850, which applied the same fix to orgqr.cpp; gesv_batch, gesvd_batch, heevd_batch and syevd_batch already had it.

### DIFF
--- a/dpnp/tests/test_blas_lapack_gil.py
+++ b/dpnp/tests/test_blas_lapack_gil.py
@@ -1,0 +1,176 @@
+"""Blocking oneMKL calls in the BLAS/LAPACK extensions must release the GIL.
+
+Routines such as ``potrf``, ``getrf``, ``syevd`` and ``gesv`` block the calling
+thread for the whole oneMKL call. Holding the GIL across that stalls every
+other Python thread, including the host tasks dpnp queues to manage object
+lifetimes, which reacquire the GIL. Five sibling files already wrap their call
+in ``py::gil_scoped_release`` (gh-2850 did this for ``orgqr``); these tests
+cover the rest.
+
+Progress of a competing thread is compared against ``SyclQueue.wait()``, which
+is ``nogil`` and therefore the best rate achievable on the machine.
+"""
+
+import sys
+import threading
+import time
+
+import dpctl
+import pytest
+
+import dpnp
+
+from .helper import has_support_aspect64, is_gpu_device
+
+# Large enough that every routine here actually blocks. Smaller sizes stop
+# discriminating: at n=512 cholesky returns asynchronously, and at n=1024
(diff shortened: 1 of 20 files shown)


dpnp/tests/test_blas_lapack_gil.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def test_gesv(self, mats):
+        spd, gen = mats
+        rhs = dpnp.ones(_SIZE, dtype="f8")
+        self._assert_releases_gil(
+            "gesv", spd, lambda: dpnp.linalg.solve(gen, rhs)
+        )
+
+
+@pytest.mark.skipif(not is_gpu_device(), reason="requires a GPU device")
```
ndgrigorian, line 126: Is there any reason this can't run on a CPU device? This way it would run in public CI

abagusetty: CPU can't show overlap here with onemkl already using all cores, so the ratio stays near ~1.0 for this multi-threaded case alone. Tried a few knobs to reliably perfect it for CPU but it wasnt deterministic

dpnp/tests/test_blas_lapack_gil.py
```diff
@@ -0,0 +1,164 @@
...
+            time.sleep(0)
+
+    def __exit__(self, *exc):
+        self._stop = True
+        self._thread.join(timeout=5)
+        return False
+
+
+@pytest.mark.skipif(not has_support_aspect64(), reason="requires fp64 support")
+class TestBlockingCallsReleaseGil:
```
antonwolfy, line 51: The new tests are quite heavy and would be good to mark them with `@pytest.mark.slow`

abagusetty: Thanks @antonwolfy, Yes I have marked the tests with `@pytest.mark.slow` decorator except for one `potrf`. This was the fastest of the bunch and hopefully provides a window to functionality tests. If it is too a bit much on the runners, will mark it down as well.

antonwolfy: It takes about 2m for me locally, I believe it would be good to add decorator for portf as well.

abagusetty: sounds good, will do

antonwolfy (approved):
Thank you @abagusetty for contributing the complete fix, LGTM

PR #3029, opened 2026-08-17
Skip einsum test_out_0d for NumPy older than 2.4.5

antonwolfy:
This guards `TestEinsum.test_out_0d` with `@testing.with_requires("numpy>=2.4.5")`.

The test compares `dpnp.einsum` against a reference `numpy.einsum` call that uses a 0-d `out` array with `optimize="optimal"`. That reference behavior only matches on NumPy 2.4.5 and newer, so the test is skipped on older NumPy versions to avoid spurious failures.

The test was introduced in #2987.


antonwolfy:
Superseded by #3030 (branch on the upstream repo instead of a fork).

PR #3031, opened 2026-08-17
Migrate conda packaging to rattler-build (v1 recipe format) for GitHub Actions CI

ekomarova:
### Summary
This PR adds rattler-build as the package-building tool for the Conda package GitHub Actions workflow, alongside a new `rattler_recipe.yaml` (v1 recipe schema) generated from the existing `meta.yaml`.

`meta.yaml` is left untouched and fully functional. This PR only changes what runs on GitHub Actions; it does not retire or replace the `conda-build` path.

### What changed

- `.github/workflows/conda-package.yml`: build steps now call rattler-build instead of conda-build

- `conda-recipe/rattler_recipe.yaml`: new file, the rattler-build (v1 schema) equivalent of `meta.yaml`. A handful of things had to be restructured because v1 format must always be valid plain YAML (no Jinja control-flow blocks)

- `environments/build_conda_pkg.yml`: no longer needed for use

### Why these specific changes were needed

- CONDA_BLD path — `rattler-build` outputs to `output/<platform>/` relative to the working directory by default, unlike `conda-build`'s $CONDA_PREFIX/conda-bld/

- `Compute version from git describe` step — `conda-build` runs `git describe` internally before rendering to populate GIT_DESCRIBE_TAG/GIT_DESCRIBE_NUMBER. `rattler-build` has no equivalent, so `rattler_recipe.yaml` now reads these via `env.get(...)`, and we compute + export them ourselves beforehand.

- RATTLER_CHANNELS array — `channels-list` includes `--override-channels`, which `rattler-build`'s CLI doesn't recognize, so it's stripped before use. Expanded into a bash array (instead of a bare $VAR) so each `-c <channel>` reaches `rattler-build` as a separate argument — also fixes a shellcheck SC2086 warning.

- Removed  usage of `conda-verify` — it's a conda-build-specific post-build checker; `rattler-build` validates internally, so it's no longer needed.

- Removed the two guard blocks and the `pyproject.toml` loop in `rattler_recipe.yaml` — v1 format must always be valid plain YAML, so Jinja control-flow ({% if %}/{% for %}) isn't supported. The version guard checks are dropped (an invalid override still fails to build, just via a less friendly solver error instead of a custom message), and the dependency list previously generated from `pyproject.toml's [build-system]` requires is now static — it needs manual updates if those requirements change.

- Added `tests: - script: ...` block — v1 schema requires an explicit test type; the old `test: requires:` block installed test deps without running anything, which isn't valid anymore. Added a minimal import/version smoke test
- Replace `miniconda` with `rattler-build` action in the build job — `rattler-build` is a standalone binary with no conda/Python dependency, so building conda packages with it doesn't need a full conda distribution on the runner. The installation takes about 1s


ekomarova:
@antonwolfy @ndgrigorian It looks like I've managed to stabilize GitHub CI, so please take a look at it when you have time. `rattler-build` wins in terms of build speed and dependency resolution

I would also consider getting rid of installing `miniconda` for `conda_package` test pipeline, since `rattler-build` is standalone and installed fairly quickly. According to my observations, all that `miniconda` is used for is indexing, searching, installation, and all this supports by `rattler` and `pixi`, so migrating the test part to a faster `rattler`/`pixi` also makes sense. But this is a separate task that I would not make part of this PR

conda-recipe/rattler_recipe.yaml
```diff
@@ -0,0 +1,121 @@
...
+      - python
+      # building with Intel DPC++ and do not import the MSVC runtime DLLs
+      - if: win
+        then: vc14_runtime
+      - if: win
+        then: ucrt
+
+tests:
+  - script:
+      - python -c "import dpnp; print(dpnp.__version__)"
```
antonwolfy, line 92: We need to add `run_test.bat` and `run_test.sh` running explicitly, since rattler-build will not automatically do that anymore:
```yaml
tests:
  # quick smoke checks
  - script:
      - pip check
      - python -c "import dpnp; print(dpnp.__version__)"
      - python -m dpctl -f
    requirements:
      run:
        - pip
  # full suite via the existing scripts
  - script: ${{ "run_test.bat" if win else "run_test.sh" }}
    requirements:
      run:
        - pytest
        - setuptools
        - if: match(python, "<3.14")
          then: scipy
```

.github/workflows/conda-package.yml
```diff
@@ -75,46 +73,74 @@ jobs:
...
       - name: Build conda package
         id: build_conda_pkg
         continue-on-error: true
-        run: conda-build --no-test --python "${{ env.python-conda-spec }}" --numpy 2.0 ${{ env.channels-list }} conda-recipe
+        shell: bash -el {0}
+        run: |
+          read -ra RATTLER_CHANNELS <<< "$(echo "${{ env.channels-list }}" | sed 's/--override-channels//')"
+          rattler-build build \
+            --recipe conda-recipe/rattler_recipe.yaml \
+            --no-include-recipe \
```
antonwolfy, line 120: We need to pass `--test skip `, since we are running tests only in separate dedicated job

.github/workflows/conda-package.yml
```diff
@@ -75,46 +73,74 @@ jobs:
...
+          echo "GIT_DESCRIBE_TAG=$(echo "$DESCRIBE" | sed -E 's/-[0-9]+-g[0-9a-f]+$//')" >> "$GITHUB_ENV"
+          echo "GIT_DESCRIBE_NUMBER=$(echo "$DESCRIBE" | sed -E 's/^.*-([0-9]+)-g[0-9a-f]+$/\1/')" >> "$GITHUB_ENV"
+
+      # rattler-build looks for build.bat on Windows (build.sh/build.bat is the
+      # symmetric naming it expects), while conda-build is still used for this
+      # recipe in non-GitHub CI and required bld.bat.
+      # Keep bld.bat as the single real source file and
+      # just copy it under the name rattler-build needs, so both build systems
+      # keep working off the same script without maintaining two duplicates
+      - name: Prepare build.bat for rattler-build
```
antonwolfy, line 107: It's possible to specify the names of build scripts explicitly in recipe file:
```yaml
build:
  script:
    file: ${{ "bld.bat" if win else "build.sh" }}
```
and that step can be dropped then

conda-recipe/rattler_recipe.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+source:
+  path: ..
+
+build:
+  number: ${{ env.get("GIT_DESCRIBE_NUMBER") }}
+  script:
+    env:
+      WHEELS_OUTPUT_FOLDER: ${{ env.get("WHEELS_OUTPUT_FOLDER", default="") }}
+      OVERRIDE_INTEL_IPO: ${{ env.get("OVERRIDE_INTEL_IPO", default="") }}
```
antonwolfy, line 21: `OVERRIDE_INTEL_IPO` is only applicable to Windows

conda-recipe/rattler_recipe.yaml
```diff
@@ -0,0 +1,121 @@
...
+      # conda-build used to forward GIT_DESCRIBE_NUMBER into the build script
+      # automatically; rattler-build does not, so bld.bat's
+      # `wheel tags --remove --build %GIT_DESCRIBE_NUMBER%` was silently
+      # getting an empty value on Windows, which made that command fail and
+      # broke the wheel install step further down the script.
+      GIT_DESCRIBE_NUMBER: ${{ env.get("GIT_DESCRIBE_NUMBER") }}
+  dynamic_linking:
+    missing_dso_allowlist:
+      - "*DPCTLSyclInterface*"
+      - "*/dpnp_backend_c.dll"
```
antonwolfy, line 31: `"*/dpnp_backend_c.dll"` is only applicable to Windows

conda-recipe/rattler_recipe.yaml
antonwolfy, line 21: It'd be better to update `bld.bat` also, to use more robust check: `if not "%OVERRIDE_INTEL_IPO%"==""`, because the default behavior might be changed now a bit (empty string "" rather then undefined env variable)

.github/workflows/conda-package.yml
```diff
@@ -75,46 +73,74 @@ jobs:
...
         run: |
-          echo "CONDA_BLD=$CONDA_PREFIX/conda-bld/${{ runner.os == 'Linux' && 'linux' || 'win' }}-64/" | tr "\\\\" '/' >> "$GITHUB_ENV"
+          echo "CONDA_BLD=$GITHUB_WORKSPACE/output/${{ runner.os == 'Linux' && 'linux-64' || 'win-64' }}/" >> "$GITHUB_ENV"
           echo "WHEELS_OUTPUT_FOLDER=$GITHUB_WORKSPACE${{ runner.os == 'Linux' && '/' || '\\' }}" >> "$GITHUB_ENV"
 
+      - name: Compute version from git describe
+        shell: bash -el {0}
+        run: |
+          DESCRIBE=$(git describe --tags --long)
+          echo "GIT_DESCRIBE_TAG=$(echo "$DESCRIBE" | sed -E 's/-[0-9]+-g[0-9a-f]+$//')" >> "$GITHUB_ENV"
```
antonwolfy, line 98: It seems rattler-build support that on recipe level also. But that can be left for the follow-up.

.github/workflows/conda-package.yml
```diff
@@ -75,46 +73,74 @@ jobs:
...
+        shell: bash -el {0}
+        run: cp conda-recipe/bld.bat conda-recipe/build.bat
+
       - name: Build conda package
         id: build_conda_pkg
         continue-on-error: true
-        run: conda-build --no-test --python "${{ env.python-conda-spec }}" --numpy 2.0 ${{ env.channels-list }} conda-recipe
+        shell: bash -el {0}
+        run: |
+          read -ra RATTLER_CHANNELS <<< "$(echo "${{ env.channels-list }}" | sed 's/--override-channels//')"
```
antonwolfy, line 117: To reduce the duplication we can define all using build options as `build-args` in `prefix-dev/rattler-build-action`

conda-recipe/rattler_recipe.yaml
```diff
@@ -0,0 +1,121 @@
+schema_version: 1
```
antonwolfy, line 1: Please populate changelog stating that rattler-build compatible version of recipe has ben added.

.github/workflows/conda-package.yml
ekomarova: This functionality is hidden behind the `--experimental` flag. Without it, any git call.* crashes with the error "Experimental feature: provide the --experimental flag...". This is the officially unstable rattler-build API for now

ekomarova:
@antonwolfy @vlad-perevezentsev @ndgrigorian please take a look when you have time

PR #3033, opened 2026-08-19
Wrap out-of-range float-to-integer astype casts to match NumPy

antonwolfy:
Casting an out-of-range floating-point value to an integer type is undefined behavior in C++. SYCL devices resolve it by saturating to the destination's min/max, whereas NumPy emits a plain C cast that, for narrow integer targets, truncates toward zero and wraps modulo the destination width (e.g. `float32(128)` becomes `int8(-128)`).

`convert_impl` (used by `astype` and every copy-and-cast kernel) only normalized this to NumPy's wrapping behavior for *unsigned* destinations. Signed narrow integer targets fell through to a raw `static_cast` and therefore *saturated* on device, diverging from NumPy. The same inconsistency showed up within a single operation: `dpnp.tensor.linalg.trace` with an `int8` output dtype produced different results depending on whether the value went through the element-wise `astype` path (saturated) or the reduction path (wrapped).

This PR generalizes the float-to-integer branch of `convert_impl` to funnel every conversion through a wider signed integer, relying on the well-defined integer narrowing to perform the modular wrap for both signed and unsigned narrow integer targets. 64-bit destinations keep their existing handling (unsigned routes negatives through `int64`; signed casts directly).


dpnp/tensor/libtensor/include/utils/type_utils.hpp
```diff
@@ -98,15 +98,34 @@ dstTy convert_impl(const srcTy &v)
...
+        // narrowing to perform the modular wrap.
+        if constexpr (sizeof(dstTy) < sizeof(std::int64_t)) {
+            return static_cast<dstTy>(static_cast<std::int64_t>(v));
+        }
+        else if constexpr (std::is_unsigned_v<dstTy>) {
+            // 64-bit unsigned destination: no wider signed integer is
+            // available to funnel through, so route only negative values
+            // through int64 to keep two's-complement wrapping well defined;
+            // non-negative values up to the unsigned maximum convert
+            // directly.
```
ndgrigorian: I think we should abbreviate these extremely verbose comments

antonwolfy: Good call — abbreviated the block from ~14 lines down to a 4-line rationale plus one-line notes on each branch in ba09d52.

vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #3041, opened 2026-08-25
Fix dpnp.insert ignoring out-of-bounds negative indices in multi-element obj

antonwolfy:
`dpnp.insert` has two index paths: the singleton path (`_insert_singleton_index`) already validated bounds, but the multi-element array path (`_insert_array_indices`) normalized negative indices with `indices[indices < 0] += n` without any bounds check. As a result, an out-of-bounds negative index mixed with in-bounds ones — e.g. `dpnp.insert([0, 1, 2], [-6, 0], [9, 8])` — silently produced a wrong result instead of raising.

This PR adds the same bounds validation NumPy introduced: if any index is `< -n` or `> n`, `IndexError` is raised with the standard `index {i} is out of bounds for axis {axis} with size {n}` message. In-bounds and out-of-bounds indices mixed in a single call now consistently raise.

### DIFF
--- a/dpnp/dpnp_iface_manipulation.py
+++ b/dpnp/dpnp_iface_manipulation.py
@@ -254,6 +254,15 @@ def _insert_array_indices(parameters, indices, values, obj):
         # Can safely cast the empty list to intp
         indices = indices.astype(dpnp.intp)
 
+    if indices.size > 0:
+        min_idx = int(indices.min())
+        max_idx = int(indices.max())
+        if min_idx < -n or max_idx > n:
+            oob = min_idx if min_idx < -n else max_idx
+            raise IndexError(
+                f"index {oob} is out of bounds for axis {axis} with size {n}"
+            )
+
     indices[indices < 0] += n
 
     numnew = len(indices)
--- a/dpnp/dpnp_iface_manipulation.py
+++ b/dpnp/dpnp_iface_manipulation.py
@@ -259,6 +259,10 @@ def _check_index_bounds(obj, indices, n, axis):
         min_idx, max_idx = dpnp.stack([indices.min(), indices.max()]).asnumpy()
     else:
         host_obj = numpy.asarray(obj)
+        if host_obj.dtype == dpnp.bool:
+            # a boolean mask selects positions, which (for an oversized mask)
+            # can fall out of bounds, so validate the flatnonzero result
+            host_obj = numpy.flatnonzero(host_obj)
         min_idx, max_idx = host_obj.min(), host_obj.max()
 
     min_idx, max_idx = int(min_idx), int(max_idx)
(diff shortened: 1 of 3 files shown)


dpnp/dpnp_iface_manipulation.py
```diff
@@ -254,6 +254,15 @@ def _insert_array_indices(parameters, indices, values, obj):
         # Can safely cast the empty list to intp
         indices = indices.astype(dpnp.intp)
 
+    if indices.size > 0:
+        min_idx = int(indices.min())
+        max_idx = int(indices.max())
```
vlad-perevezentsev: 2 copy to host and .wait() executions
Is there any way to reduce the host transfer here?

antonwolfy: Rework a bit to minimize impact, but no way to avoid that completely

dpnp/dpnp_iface_manipulation.py
```diff
@@ -240,6 +240,35 @@ def _calc_parameters(a, axis, obj, values=None):
...
+    """
+
+    if isinstance(obj, slice) or indices.size == 0:
+        return
+
+    if dpnp.is_supported_array_type(obj):
+        min_idx, max_idx = dpnp.stack([indices.min(), indices.max()]).asnumpy()
+    else:
+        host_obj = numpy.asarray(obj)
+        min_idx, max_idx = host_obj.min(), host_obj.max()
```
vlad-perevezentsev, line 266: NumPy checks indices after `flatnonzero` for boolean masks 
https://github.com/numpy/numpy/blob/main/numpy/lib/_function_base_impl.py#L5555 

The current logic returns a wrong result 
```
In [1]: import dpnp

In [2]: import numpy

In [3]: dpnp.insert(dpnp.array([0,1,2]), numpy.array([True,False,False,False,True]), 9)
Out[3]: array([9, 0, 1, 2, 9])

In [4]: numpy.insert(numpy.array([0,1,2]), numpy.array([True,False,False,False,True]), 9)
Cell In[4], line 1
----> 1 numpy.insert(numpy.array([0,1,2]), numpy.array([True,False,False,False,True]), 9)

File ~/miniforge3/envs/dpnp_conda/lib/python3.14/site-packages/numpy/lib/_function_base_impl.py:5576, in insert(arr, obj, values, axis)
   5574 newshape[axis] += numnew
   5575 old_mask = ones(newshape[axis], dtype=bool)
-> 5576 old_mask[indices] = False
   5578 new = empty(newshape, arr.dtype, arrorder)
   5579 slobj2 = [slice(None)] * ndim

IndexError: index 5 is out of bounds for axis 0 with size 5
```

To fix that, we need to add below 261 line
```
if host_obj.dtype == numpy.bool_:
    host_obj = numpy.flatnonzero(host_obj)
```

antonwolfy: Great catch! Fixed now.

vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy

PR #3042, opened 2026-08-25
Fix sycl::queue leak in usm_ndarray `get_queue()`/`get_device()`

vlad-perevezentsev:
This PR fixes a memory leak in the C++ accessors `usm_ndarray::get_queue()` and `get_device()` in `dpnp/include/dpnp4pybind11.hpp`

`UsmNDArray_GetQueueRef_` (dpnp's `usm_ndarray.get_queue_ref()`) returns an owning copy of the queue (it calls `DPCTLQueue_Copy` i.e. new `sycl::queue`)
Both accessors dereferenced this pointer and copied the queue out but never freed the heap copy, leaking one `sycl::queue` per call. Under queue churn this keeps `queue_impl` alive and prevents reuse of the associated Level Zero command lists / event pools.

The fix adds `UsmNDArray_RemoveQueueRef` C-API to `dpnp_capi` (it binds `DPCTLQueue_Delete`); `get_queue()/get_device()` take ownership of the returned copy with `std::unique_ptr<sycl::queue>` whose custom deleter frees it via `UsmNDArray_RemoveQueueRef`. This keeps allocation and deallocation on dpctl's side of the ABI. 

 dpctl needs no change because its `Memory_GetQueueRef` returns a borrowed ref, whereas dpnp's `UsmNDArray_GetQueueRef` is documented to return a copy so it must be deleted.


dpnp/include/dpnp4pybind11.hpp
```diff
@@ -489,17 +489,26 @@ class usm_ndarray : public py::object
         PyUSMArrayObject *raw_ar = usm_array_ptr();
 
         auto const &api = detail::dpnp_capi::get();
+        // UsmNDArray_GetQueueRef_ returns an owning copy (DPCTLQueue_Copy,
+        // i.e. `new sycl::queue`); wrap it in a unique_ptr to avoid leaking a
+        // queue per call.
         DPCTLSyclQueueRef QRef = api.UsmNDArray_GetQueueRef_(raw_ar);
```
antonwolfy, line 497: `QRef` might be NULL on OOM inside `DPCTLQueue_Copy`, while below dereferencing of NULL is UB.
It seems legacy and can be done in the follow-up PR.

dpnp/include/dpnp4pybind11.hpp
```diff
@@ -489,17 +489,26 @@ class usm_ndarray : public py::object
...
-        return *(reinterpret_cast<sycl::queue *>(QRef));
+        std::unique_ptr<sycl::queue> q_ptr{
```
antonwolfy: One point worth considering: freeing via `delete` (the `unique_ptr` default deleter) relies on an undocumented implementation detail rather than dpctl's documented ownership contract.

- `DPCTLSyclQueueRef` is documented as an *opaque* pointer, and `DPCTLQueue_Copy` is annotated `__dpctl_give`, whose contract is: "the caller now owns the object … to free it, use it exactly once as a value for a `__dpctl_take` argument" — i.e. `DPCTLQueue_Delete`.
- `delete` happens to be equivalent today only because `DPCTLQueue_Delete` is literally `delete reinterpret_cast<sycl::queue*>(QRef)` and dpnp/dpctl share the same runtime/allocator. If dpctl ever changed how it allocates the copy (pool/custom allocator), raw `delete` would break while `DPCTLQueue_Delete` would not.

Non-blocking, but to honor the documented contract we could use a `unique_ptr` with a custom deleter.

antonwolfy: A cleaner option that honors the contract *and* avoids the link dependency: add a `UsmNDArray_RemoveQueueRef` capi export that binds `DPCTLQueue_Delete`, and route the free through the capsule function pointer — mirroring how `UsmNDArray_GetQueueRef_` already works, so no consumer needs to link `libDPCTLSyclInterface`.

`dpnp/tensor/_usmarray.pyx`:
```cython
cdef api void UsmNDArray_RemoveQueueRef(c_dpctl.DPCTLSyclQueueRef QRef):
    """Delete a DPCTLSyclQueueRef previously returned by UsmNDArray_GetQueueRef"""
    c_dpctl.DPCTLQueue_Delete(QRef)   # safe on NULL
```

`dpnp/include/dpnp4pybind11.hpp` — add the struct member / ctor-init / assignment mirroring `UsmNDArray_GetQueueRef_`, then:
```cpp
DPCTLSyclQueueRef QRef = api.UsmNDArray_GetQueueRef_(raw_ar);
auto qref_deleter = [](sycl::queue *q) {
    detail::dpnp_capi::get().UsmNDArray_RemoveQueueRef_(
        reinterpret_cast<DPCTLSyclQueueRef>(q));
};
std::unique_ptr<sycl::queue, decltype(qref_deleter)> q_ptr{
    reinterpret_cast<sycl::queue *>(QRef), qref_deleter};
return *q_ptr;
```

This keeps allocation and deallocation on dpctl's side of the ABI and gives a symmetric Get/Remove pair (the capi test would be the natural place for a round-trip check).

That change is non-blocking for the PR — the current `delete` is provably equivalent under the supported build. So can be also done in the follow-up.

dpnp/include/dpnp4pybind11.hpp
```diff
@@ -489,17 +489,26 @@ class usm_ndarray : public py::object
         PyUSMArrayObject *raw_ar = usm_array_ptr();
 
         auto const &api = detail::dpnp_capi::get();
+        // UsmNDArray_GetQueueRef_ returns an owning copy (DPCTLQueue_Copy,
```
antonwolfy: Can we use a shorter comment?

vlad-perevezentsev: Done

dpnp/include/dpnp4pybind11.hpp
vlad-perevezentsev: Great point, agreed. 
Using raw delete could break if dpctl changes the allocator.

I updated it to use `DPCTLQueue_Delete` through `UsmNDArray_RemoveQueueRef` with a custom unique_ptr deleter.
I also used it in `test_usm_ndarray_capi.py` to cover and fix the leak there as well
Thanks

dpnp/include/dpnp4pybind11.hpp
vlad-perevezentsev: A good catch
I will be implemented in the follow-up PR

antonwolfy (approved):
Thank you @vlad-perevezentsev, LGTM

PR #3043, opened 2026-08-26
Fix `dpnp.linspace` returning NaN for equal infinite endpoints

antonwolfy:
`dpnp.linspace` returned `NaN` instead of the endpoint value when `start` and `stop` are equal and infinite — e.g. `dpnp.linspace(inf, inf, 4)` gave `[nan inf inf nan]`.

The bug had two independent sources:
- The array path in `dpnp_linspace` computed `delta = stop - start`, so `inf - inf = NaN` propagated through the whole result.
- The scalar path delegates to `dpnp.tensor.linspace`, whose affine kernel `LinearSequenceAffineFunctor` evaluates `start * w + stop * wc`; at the endpoints one weight is `0`, and `inf * 0 = NaN`.

The PR proposes to fix all the places.

Note, the tests compare against NumPy on versions that include the fix (NumPy >= 2.6.0) and fall back to explicit expected values on older NumPy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,7 @@ This release is compatible with NumPy 2.5.
 * Fixed `dpnp.ndarray.view` ignoring the USM element offset of a sliced array, which also caused `dpnp.einsum` to silently return wrong results for a single sliced operand with no summed index [#3037](https://github.com/IntelPython/dpnp/pull/3037)
 * Fixed `dpnp.all` and `dpnp.any` aborting when reducing over an empty axis (e.g. an array with a zero-length dimension) [#3021](https://github.com/IntelPython/dpnp/pull/3021)
 * Released the GIL before the blocking OneMKL DFT calls in the FFT extension [#3040](https://github.com/IntelPython/dpnp/pull/3040)
+* Fixed `dpnp.linspace` returning `nan` for equal infinite endpoints [#3043](https://github.com/IntelPython/dpnp/pull/3043)
 
 ### Security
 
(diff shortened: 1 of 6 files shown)


vlad-perevezentsev (approved):
LGTM
Thank you @antonwolfy 

One separate issue for a follow-up
`dpnp.linspace(-dpnp.inf, dpnp.inf, num)` returns `NaN` for the last value instead of `stop`

```
In [4]: dpnp.linspace(-dpnp.inf, dpnp.inf, 4)
Out[4]: array([nan, nan, nan, nan])

In [6]: numpy.linspace(-numpy.inf, numpy.inf, 4)
/home/sdp/miniforge3/envs/dpnp_conda/lib/python3.14/site-packages/numpy/_core/function_base.py:163: RuntimeWarning: invalid value encountered in multiply
  y *= step
/home/sdp/miniforge3/envs/dpnp_conda/lib/python3.14/site-packages/numpy/_core/function_base.py:173: RuntimeWarning: invalid value encountered in add
  y += start
Out[6]: array([nan, nan, nan, inf])
```

antonwolfy:
Just to note, NumPy behavior also doesn't seem correct, since the result should be:
```
[-inf, nan, nan, inf]
```
Since `linspace(start, stop, num, endpoint=True)` should return evenly spaced points on `[−∞, +∞]`. And MatLab returns exactly the above values.

I believe the only way to do that in dpnp/dpnp.tensor now is to have a special handling for the first and last elements of the interval.
But I wonder if that is needed for such edge case with no real use case.

antonwolfy:
It seems the same is applicable to similar use cases:
```
np.linspace(0, np.inf, 4)
# Out: array([nan, inf, inf, inf])
# Expected:  [0.,  inf, inf, inf]

# np.linspace(-np.inf, 0, 4)
# Out: array([nan,   nan,  nan, 0.])
# Expected:  [-inf, -inf, -inf, 0.]
```

PR #3045, opened 2026-08-27
Fix `dpnp.ndarray.flat` indexing edge cases

antonwolfy:
This PR reworks `dpnp.flatiter` (`dpnp.ndarray.flat`) indexing so it aligns with NumPy's flat-iterator semantics, which were tightened in NumPy 2.4.

Previously `dpnp.ndarray.flat` accepted only a single integer index and raised `TypeError` for everything else. Indexing now delegates to regular array indexing of the flattened array, so it supports the full set of flat index types and matches NumPy's error behavior.

### DIFF
--- a/dpnp/dpnp_flatiter.py
+++ b/dpnp/dpnp_flatiter.py
@@ -182,9 +182,12 @@ class flatiter:
             idx = flat_index[key]
 
         if not dpnp.isscalar(val):
-            val = dpnp.asarray(
-                val, sycl_queue=exec_q, usm_type=usm_type
-            ).ravel()
+            val = dpnp.asarray(val, sycl_queue=exec_q, usm_type=usm_type)
+            if idx.ndim == 0 and val.ndim != 0:
+                # a scalar index targets a single item, reject an array value
+                raise ValueError("Error setting single item of array.")
+
+            val = val.ravel()
             n = idx.size
             if 0 < val.size != n:
                 # cycles the values over the selection
--- a/dpnp/dpnp_flatiter.py
+++ b/dpnp/dpnp_flatiter.py
@@ -78,6 +78,14 @@ class flatiter:
         self._i = 0
 
     @staticmethod
+    def _unwrap_tuple(key):
+        # a flat iterator is 1-D, so a single-element index tuple is equivalent
+        # to its element (e.g. `flat[(idx,)]` behaves like `flat[idx]`)
+        if isinstance(key, tuple) and len(key) == 1:
+            return key[0]
+        return key
+
+    @staticmethod
     def _reject_newaxis(key):
         # newaxis (None) is valid for array indexing but not for flat indexing
         if key is None or (
@@ -119,6 +127,7 @@ class flatiter:
         return dpnp.reshape(self._arr, -1)
 
     def __getitem__(self, key):
+        key = self._unwrap_tuple(key)
         self._reject_newaxis(key)
         self._check_bounds(key)
 
@@ -126,6 +135,7 @@ class flatiter:
         return self._flatten()[key].copy()
 
     def __setitem__(self, key, val):
+        key = self._unwrap_tuple(key)
         self._reject_newaxis(key)
         self._check_bounds(key)
 
--- a/dpnp/dpnp_flatiter.py
+++ b/dpnp/dpnp_flatiter.py
@@ -82,24 +82,6 @@ class flatiter:
         self._size = a.size
         self._i = 0
 
-    @staticmethod
-    def _unwrap_tuple(key):
-        # a flat iterator is 1-D, so a single-element index tuple is equivalent
-        # to its element (e.g. `flat[(idx,)]` behaves like `flat[idx]`)
-        if isinstance(key, tuple) and len(key) == 1:
-            return key[0]
-        return key
-
-    @staticmethod
-    def _reject_invalid_key(key):
-        # newaxis (None) and multi-element tuples (a flat iterator is 1-D and
-        # takes a single index) are not valid, unlike regular array indexing
-        if key is None or (isinstance(key, tuple) and len(key) > 1):
-            raise IndexError(
-                "only integers, slices (`:`), ellipsis (`...`) and integer "
-                "or boolean arrays are valid indices"
-            )
-
     def _check_bounds(self, key):
         # fancy int indices wrap instead of raising, so check them vs NumPy
         if key is Ellipsis or isinstance(key, (slice, bool, tuple)):
@@ -131,22 +113,26 @@ class flatiter:
         if lo < -size:
             raise IndexError(f"index {lo} is out of bounds for size {size}")
 
-    def _flatten(self):
-        # C-order flat view (copy if non-contiguous)
-        return dpnp.reshape(self._arr, -1)
+    def _normalize_key(self, key):
+        # 1-D iterator: unwrap a 1-elem tuple; reject None and longer tuples
+        if isinstance(key, tuple) and len(key) == 1:
+            key = key[0]
+        if key is None or (isinstance(key, tuple) and len(key) > 1):
+            raise IndexError(
+                "only integers, slices (`:`), ellipsis (`...`) and integer "
+                "or boolean arrays are valid indices"
+            )
+        self._check_bounds(key)
+        return key
 
     def __getitem__(self, key):
-        key = self._unwrap_tuple(key)
-        self._reject_invalid_key(key)
-        self._check_bounds(key)
+        key = self._normalize_key(key)
 
         # flat always yields a copy, never a view
-        return self._flatten()[key].copy()
+        return dpnp.reshape(self._arr, -1)[key].copy()
 
     def __setitem__(self, key, val):
-        key = self._unwrap_tuple(key)
-        self._reject_invalid_key(key)
-        self._check_bounds(key)
+        key = self._normalize_key(key)
 
         if isinstance(key, tuple) and len(key) == 0:
             # NumPy rejects arr.flat[()] = val
--- a/dpnp/dpnp_flatiter.py
+++ b/dpnp/dpnp_flatiter.py
@@ -34,57 +34,89 @@ import dpnp
 class flatiter:
     """Flat iterator object to iterate over arrays."""
 
-    def __init__(self, X):
-        if type(X) is not dpnp.ndarray:
+    def __init__(self, a):
+        if type(a) is not dpnp.ndarray:
             raise TypeError(
-                "Argument must be of type dpnp.ndarray, got {}".format(type(X))
+                "Argument must be of type dpnp.ndarray, got {}".format(type(a))
             )
-        self.arr_ = X
-        self.size_ = X.size
-        self.i_ = 0
+        self._arr = a
+        self._size = a.size
+        self._i = 0
 
-    def _multiindex(self, i):
-        nd = self.arr_.ndim
-        if nd == 0:
-            if i == 0:
-                return ()
-            raise KeyError
-        elif nd == 1:
-            return (i,)
-        sh = self.arr_.shape
-        i_ = i
-        multi_index = [0] * nd
-        for k in reversed(range(1, nd)):
-            si = sh[k]
-            q = i_ // si
-            multi_index[k] = i_ - q * si
-            i_ = q
-        multi_index[0] = i_
-        return tuple(multi_index)
+    @staticmethod
+    def _reject_newaxis(key):
+        # newaxis (None) is valid for array indexing but not for flat indexing
+        if key is None or (
+            isinstance(key, tuple) and any(k is None for k in key)
+        ):
+            raise IndexError(
+                "only integers, slices (`:`), ellipsis (`...`) and integer "
+                "or boolean arrays are valid indices"
+            )
+
+    def _check_bounds(self, key):
+        # fancy int indices wrap instead of raising, so check them vs NumPy
+        if key is Ellipsis or isinstance(key, (slice, bool, tuple)):
+            return
+
+        if isinstance(key, int) or (
+            callable(getattr(key, "__index__", None))
+            and not hasattr(key, "ndim")
+        ):
+            return  # scalar int: regular indexing checks it
+
+        try:
+            idx = dpnp.asarray(key, sycl_queue=self._arr.sycl_queue)
+        except Exception:
+            return  # let regular indexing raise
+
+        if idx.dtype.kind not in "iu" or idx.size == 0:
+            return
+
+        size = self._size
+        hi, lo = int(dpnp.max(idx)), int(dpnp.min(idx))
+        if hi >= size:
+            raise IndexError(f"index {hi} is out of bounds for size {size}")
+        if lo < -size:
+            raise IndexError(f"index {lo} is out of bounds for size {size}")
+
+    def _flatten(self):
+        # C-order flat view (copy if non-contiguous)
+        return dpnp.reshape(self._arr, -1)
 
     def __getitem__(self, key):
-        idx = getattr(key, "__index__", None)
-        if not callable(idx):
-            raise TypeError(key)
-        i = idx()
-        mi = self._multiindex(i)
-        return self.arr_.__getitem__(mi)
+        self._reject_newaxis(key)
+        self._check_bounds(key)
+
+        # flat always yields a copy, never a view
+        return self._flatten()[key].copy()
 
     def __setitem__(self, key, val):
-        idx = getattr(key, "__index__", None)
-        if not callable(idx):
-            raise TypeError(key)
-        i = idx()
-        mi = self._multiindex(i)
-        return self.arr_.__setitem__(mi, val)
+        self._reject_newaxis(key)
+        self._check_bounds(key)
+
+        if isinstance(key, tuple) and len(key) == 0:
+            # NumPy rejects arr.flat[()] = val
+            raise IndexError(
+                "Assigning to a flat iterator with a 0-D index is not "
+                "supported"
+            )
+
+        # resolve key to flat positions, reusing regular indexing to validate
+        arr = self._arr
+        flat_index = dpnp.arange(
+            arr.size, sycl_queue=arr.sycl_queue, usm_type=arr.usm_type
+        )
+        positions = dpnp.reshape(flat_index[key], -1)
+        dpnp.put(arr, positions, val)
 
     def __iter__(self):
         return self
 
     def __next__(self):
-        if self.i_ < self.size_:
-            val = self.__getitem__(self.i_)
-            self.i_ = self.i_ + 1
+        if self._i < self._size:
+            val = self.__getitem__(self._i)
+            self._i = self._i + 1
             return val
         else:
             raise StopIteration
--- a/dpnp/dpnp_flatiter.py
+++ b/dpnp/dpnp_flatiter.py
@@ -35,9 +35,9 @@ class flatiter:
     """Flat iterator object to iterate over arrays."""
 
     def __init__(self, a):
-        if type(a) is not dpnp.ndarray:
+        if not isinstance(a, dpnp.ndarray):
             raise TypeError(
-                "Argument must be of type dpnp.ndarray, got {}".format(type(a))
+                f"An array must be of type dpnp.ndarray, but got {type(a)}"
             )
         self._arr = a
         self._size = a.size
--- a/dpnp/dpnp_flatiter.py
+++ b/dpnp/dpnp_flatiter.py
@@ -28,7 +28,12 @@
 
 """Implementation of flatiter."""
 
+import numpy
+
 import dpnp
+import dpnp.tensor as dpt
+
+from .dpnp_array import dpnp_array
 
 
 class flatiter:
@@ -107,16 +112,21 @@ class flatiter:
         ):
             return  # scalar int: regular indexing checks it
 
-        try:
-            idx = dpnp.asarray(key, sycl_queue=self._arr.sycl_queue)
-        except Exception:
-            return  # let regular indexing raise
+        if isinstance(key, dpnp_array):
+            idx = key
+        elif isinstance(key, dpt.usm_ndarray):
+            idx = dpnp_array._create_from_usm_ndarray(key)
+        else:
+            try:
+                idx = numpy.asarray(key)
+            except Exception:
+                return  # let regular indexing raise
 
-        if idx.dtype.kind not in "iu" or idx.size == 0:
+        if not dpnp.issubdtype(idx.dtype, dpnp.integer) or idx.size == 0:
             return
 
         size = self._size
-        hi, lo = int(dpnp.max(idx)), int(dpnp.min(idx))
+        hi, lo = int(idx.max()), int(idx.min())
         if hi >= size:
             raise IndexError(f"index {hi} is out of bounds for size {size}")
         if lo < -size:
--- a/dpnp/dpnp_flatiter.py
+++ b/dpnp/dpnp_flatiter.py
@@ -91,11 +91,10 @@ class flatiter:
         return key
 
     @staticmethod
-    def _reject_newaxis(key):
-        # newaxis (None) is valid for array indexing but not for flat indexing
-        if key is None or (
-            isinstance(key, tuple) and any(k is None for k in key)
-        ):
+    def _reject_invalid_key(key):
+        # newaxis (None) and multi-element tuples (a flat iterator is 1-D and
+        # takes a single index) are not valid, unlike regular array indexing
+        if key is None or (isinstance(key, tuple) and len(key) > 1):
             raise IndexError(
                 "only integers, slices (`:`), ellipsis (`...`) and integer "
                 "or boolean arrays are valid indices"
@@ -138,7 +137,7 @@ class flatiter:
 
     def __getitem__(self, key):
         key = self._unwrap_tuple(key)
-        self._reject_newaxis(key)
+        self._reject_invalid_key(key)
         self._check_bounds(key)
 
         # flat always yields a copy, never a view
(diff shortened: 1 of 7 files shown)


dpnp/dpnp_flatiter.py
```diff
@@ -32,59 +32,139 @@
...
-            si = sh[k]
-            q = i_ // si
-            multi_index[k] = i_ - q * si
-            i_ = q
-        multi_index[0] = i_
-        return tuple(multi_index)
+
+    def _check_bounds(self, key):
+        # fancy int indices wrap instead of raising, so check them vs NumPy
+        if key is Ellipsis or isinstance(key, (slice, bool, tuple)):
```
vlad-perevezentsev: This logic looks wrong for a `tuple`
If we pass a tuple index OOB index will not be caught
```
In [1]: import numpy, dpnp

In [2]: a = numpy.array([1,2,3])

In [3]: a_dp = dpnp.array(a)

In [4]: a.flat[(numpy.array([5]),)]
Cell In[4], line 1
----> 1 a.flat[(numpy.array([5]),)]

IndexError: index 5 is out of bounds for size 3

In [5]: a_dp.flat[(dpnp.array([5]),)]
Out[5]: array([3])
```

vlad-perevezentsev: The same as for `bool` and gives a different `shape`
```
In [6]: a.flat[True]
<ipython-input-6-e68040906b52>:1: DeprecationWarning: Indexing flat iterators with a 0-dimensional boolean index is deprecated and may be removed in a future version. (Deprecated NumPy 2.4)
  a.flat[True]
Out[6]: np.int64(1)

In [7]: a_dp.flat[True]
Out[7]: array([[1, 2, 3]])
```

dpnp/dpnp_flatiter.py
```diff
@@ -32,59 +32,139 @@
...
+                "Assigning to a flat iterator with a 0-D index is not "
+                "supported"
+            )
+
+        a = self._arr
+        exec_q = a.sycl_queue
+        usm_type = a.usm_type
+
+        # resolve key to flat positions, reusing regular indexing to validate
+        flat_index = dpnp.arange(a.size, sycl_queue=exec_q, usm_type=usm_type)
```
vlad-perevezentsev: For a scalar int key we allocate a full `arange(a.size)` just to resolve one position.
Could  you add a fast path to avoid it? Like this
```
if isinstance(key, int):
    pos = key + a.size if key < 0 else key
    if not 0 <= pos < a.size:
        raise IndexError(f"index {key} is out of bounds for size {a.size}")
    idx = dpnp.asarray(pos, sycl_queue=exec_q, usm_type=usm_type)
else:
    flat_index = dpnp.arange(a.size, sycl_queue=exec_q, usm_type=usm_type)
    idx = flat_index[key]
```

dpnp/dpnp_flatiter.py
```diff
@@ -32,59 +32,139 @@
...
+        try:
+            idx = dpnp.asarray(key, sycl_queue=self._arr.sycl_queue)
+        except Exception:
+            return  # let regular indexing raise
+
+        if idx.dtype.kind not in "iu" or idx.size == 0:
+            return
+
+        size = self._size
+        hi, lo = int(dpnp.max(idx)), int(dpnp.min(idx))
```
vlad-perevezentsev: 2 copy to host and .wait() executions
Is there any way to reduce the host transfer here?

dpnp/dpnp_flatiter.py
antonwolfy: Regarding `flat[True]` (the 0-d boolean case): numpy deprecated it in 2.4 (`DeprecationWarning: … may be removed in a future version`), so it seemed low-value to add code replicating a quirk numpy is actively removing.

antonwolfy: So I'd prefer to keep the standard array boolean-index semantics, rather than duplicating the deprecated numpy behavior.
Once the deprecation expires, numpy will also return:
```python
a.flat[True]
# Out: array([[1, 2, 3]])
```

antonwolfy: The tuple index OOB is addressed.

dpnp/dpnp_flatiter.py
antonwolfy: Only left for the use case when input `key` is dpnp.ndarray or usm.ndarray.

dpnp/tests/third_party/cupy/indexing_tests/test_iterate.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
ndgrigorian, line 1: Is this import necessary? It looks like an artifact

antonwolfy: It's need to reduce the tests diff towards CuPy latest

ndgrigorian: I would personally favor removing it since it's unused but it's mostly not much of an issue

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,180 @@
...
+                "Assigning to a flat iterator with a 0-D index is not "
+                "supported"
+            )
+
+        a = self._arr
+        exec_q = a.sycl_queue
+        usm_type = a.usm_type
+
+        # resolve key to flat positions, reusing regular indexing to validate
+        if isinstance(key, int) and not isinstance(key, bool):
```
vlad-perevezentsev: A scalar key (here and np.int64/0d array in else) silently takes the first value instead of raising a `ValueError` (like NumPy)

```
In [1]: import numpy

In [2]: import dpnp

In [3]: a = dpnp.arange(6)

In [4]: a_np = numpy.arange(6)

In [5]: a.flat[0] = [1, 2, 3]

In [6]: a_np.flat[0] = [1, 2, 3]
Cell In[6], line 1
----> 1 a_np.flat[0] = [1, 2, 3]

ValueError: Error setting single item of array.

In [7]: a.flat[dpnp.array(0)] = [1, 2, 3]

In [8]: a_np.flat[numpy.array(0)] = [1, 2, 3]
Cell In[8], line 1
----> 1 a_np.flat[numpy.array(0)] = [1, 2, 3]

ValueError: Error setting single item of array.

In [9]: a.flat[0:1] = [1, 2, 3]

In [10]: a_np.flat[0:1] = [1, 2, 3]

```

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,179 @@
...
-        idx = getattr(key, "__index__", None)
-        if not callable(idx):
-            raise TypeError(key)
-        i = idx()
-        mi = self._multiindex(i)
-        return self.arr_.__getitem__(mi)
+        key = self._normalize_key(key)
+
+        # flat always yields a copy, never a view
+        return dpnp.reshape(self._arr, -1)[key].copy()
```
vlad-perevezentsev: Can you add a fast path for a plain `int key` that indexes the array directly via the old `_multiindex` instead of flattening? It should improve performance

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,179 @@
...
+            idx = dpnp_array._create_from_usm_ndarray(key)
+        else:
+            try:
+                idx = numpy.asarray(key)
+            except (TypeError, ValueError):
+                return  # let regular indexing raise
+
+        if dpnp.issubdtype(idx.dtype, dpnp.bool):
+            # only a boolean ndarray mask is valid; reject bool scalars/lists
+            if idx.ndim > 0 and not isinstance(key, (bool, list, tuple)):
```
vlad-perevezentsev: Looks like `bool` and `tuple` are unreachable

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,179 @@
...
+        self._i = 0
+
+    def _validate_key(self, key):
+        # Ellipsis/slice/tuple need no validation here
+        if key is Ellipsis or isinstance(key, (slice, tuple)):
+            return
+
+        # a genuine scalar int (not bool): regular indexing checks it
+        if not isinstance(key, bool) and (
+            isinstance(key, int)
```
vlad-perevezentsev: This is subsumed by the `__index__` check below it

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,179 @@
...
+                )
+            idx = dpnp.asarray(pos, sycl_queue=exec_q, usm_type=usm_type)
+        elif isinstance(key, slice):
+            # slice fast path: build only the selected positions
+            start, stop, step = key.indices(a.size)
+            idx = dpnp.arange(
+                start, stop, step, sycl_queue=exec_q, usm_type=usm_type
+            )
+        else:
+            flat_index = dpnp.arange(
```
vlad-perevezentsev, line 228: when `key` is a boolean mask, `arange(a.size)[key]` is just `dpnp.nonzero(key)[0]` without the O(size)

antonwolfy: Done — the boolean-mask branch in `__setitem__` now uses `dpnp.nonzero(mask)[0]` instead of `arange(a.size)[key]`, so it no longer materializes the full-size index buffer.

I also added the matching size check in `_validate_key`: a boolean mask whose length differs from the array size now raises `IndexError` (previously `arange(size)[mask]` did that implicitly, but `nonzero` would have silently accepted a mismatched mask). Covered by a new `test_flat_bool_mask_wrong_size` test.

dpnp/dpnp_flatiter.py
antonwolfy: Done — dropped the `isinstance(key, int)` clause. A Python `int` has `__index__` and no `ndim`, so it's fully covered by the `__index__` check (with the `not isinstance(key, bool)` guard kept to keep booleans out).

dpnp/dpnp_flatiter.py
antonwolfy: Done — `__getitem__` now has a scalar fast path: for a plain `int` it resolves the multi-index via `numpy.unravel_index(pos, self._arr.shape)` and indexes `self._arr` directly, instead of `dpnp.reshape(self._arr, -1)[key]`. This avoids flattening (which copies the whole array for a non-contiguous `self._arr`), so iteration over such an array is no longer O(n) per element.

Covered by `test_flat_getitem` (now including negative and non-contiguous `2D.T` scalar indices), `test_flat_scalar_getitem_returns_copy`, and the out-of-bounds/0-d cases.

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,198 @@
...
+            idx = dpnp_array._create_from_usm_ndarray(key)
+        else:
+            try:
+                idx = numpy.asarray(key)
+            except (TypeError, ValueError):
+                return  # let regular indexing raise
+
+        if dpnp.issubdtype(idx.dtype, dpnp.bool):
+            # only a boolean ndarray mask is valid; reject bool scalars/lists
+            if idx.ndim > 0 and not isinstance(key, list):
```
vlad-perevezentsev: The check `if idx.ndim > 1` is missing and is not covered by the tests 
```
In [3]: a = dpnp.arange(6).reshape(2, 3)

In [4]: m = dpnp.array([[True, False, True], [False, True, False]])

In [5]: a_np = dpnp.asnumpy(a)

In [6]: m_np = dpnp.asnumpy(m)

In [7]: a.flat[m] = -1

In [8]: a_np.flat[m_np] = -1
Cell In[8], line 1
----> 1 a_np.flat[m_np] = -1

IndexError: too many indices for flat iterator: flat iterator is 1-dimensional, but 2 were indexed
```

```
if dpnp.issubdtype(idx.dtype, dpnp.bool):
    if idx.ndim > 1:
        raise IndexError(
            "too many indices for flat iterator: flat iterator is "
            f"1-dimensional, but {idx.ndim} were indexed"
        )
    # only a boolean ndarray mask is valid; reject bool scalars/lists
    if idx.ndim == 1 and not isinstance(key, list):
```

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,198 @@
...
+            val = val.ravel()
+            n = idx.size
+            if val.size and val.size != n:
+                # cycles the values over the selection
+                val = val[
+                    dpnp.arange(n, sycl_queue=exec_q, usm_type=usm_type)
+                    % val.size
+                ]
+
+        dpnp.put(a, idx, val)
```
vlad-perevezentsev, line 248: Calling `dpnp.put` prevents memory overlap while NumPy allows it
```
In [10]: a = dpnp.arange(6)

In [11]: a_np = numpy.arange(6)

In [12]: a.flat[:-1] = a[1:]

ValueError: Arrays index overlapping segments of memory

In [13]: a_np.flat[:-1] = a_np[1:]

In [14]: a[:-1] = a[1:]  # setitem works
```

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,198 @@
...
-        return self.arr_.__getitem__(mi)
+        key = self._normalize_key(key)
+
+        if isinstance(key, int) and not isinstance(key, bool):
+            # scalar fast path: index directly instead of flattening the array
+            pos = self._scalar_pos(key)
+            return self._arr[numpy.unravel_index(pos, self._arr.shape)].copy()
+
+        # flat always yields a copy, never a view
+        return dpnp.reshape(self._arr, -1)[key].copy()
```
vlad-perevezentsev: Advanced indexing already allocates a fresh array so the `.copy()` is a second full-size copy

We can add this 
```
res = dpnp.reshape(self._arr, -1)[key]
    if getattr(key, "ndim", 0) == 0 and not isinstance(key, list):
        # basic indexing returns a view, but flat must yield a copy
        res = res.copy()
    return res
```

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,198 @@
...
+            mask = dpnp.asarray(key, sycl_queue=exec_q, usm_type=usm_type)
+            idx = dpnp.nonzero(mask)[0]
+        else:
+            flat_index = dpnp.arange(
+                a.size, sycl_queue=exec_q, usm_type=usm_type
+            )
+            idx = flat_index[key]
+
+        if not dpnp.isscalar(val):
+            val = dpnp.asarray(val, sycl_queue=exec_q, usm_type=usm_type)
```
vlad-perevezentsev: Here and on 191 line `__setitem__` bypasses compute follows data

Passing `sycl_queue/usm_type` to `dpnp.asarray()` silently migrates an input that lives on another queue so flat setitem accepts what the rest of dpnp rejects 

```
In [2]: a = dpnp.arange(6)

In [3]: q = dpctl.SyclQueue(a.sycl_device) # diff queue, same device

In [4]: v = dpnp.array([7, 8], sycl_queue=q)

In [5]: m = dpnp.array([True, False] * 3, sycl_queue=q)

In [6]: a[0:2] = v
ValueError: Execution queue is not compatible with allocation queues

In [7]: dpnp.put(a, dpnp.array([0, 1]), v)
ExecutionPlacementError: 

In [8]: a.flat[m]
ExecutionPlacementError: arrays have different associated queues. Use `y.to_device(x.device)` to migrate.

In [9]: a.flat[0:2] = v

In [10]: a.flat[m] = -1
```

Should we align behavior everywhere?

Like this

```
if not dpnp.isscalar(val):
    if dpnp.is_supported_array_type(val):
        val = dpnp.asarray(val)          # keep its own queue and put raises
    else:
        val = dpnp.asarray(val, sycl_queue=exec_q, usm_type=usm_type)
```

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,198 @@
...
+            if idx.ndim == 0 and val.ndim != 0:
+                # a scalar index targets a single item, reject an array value
+                raise ValueError("Error setting single item of array.")
+
+            val = val.ravel()
+            n = idx.size
+            if val.size and val.size != n:
```
vlad-perevezentsev: `dpnp.put` broadcasts size-1 and 0-d values by itself so cycling only adds two full-size temporaries

I think it will be better
```
if val.size not in (0, 1, n):
```

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,198 @@
...
     def __getitem__(self, key):
-        idx = getattr(key, "__index__", None)
-        if not callable(idx):
-            raise TypeError(key)
-        i = idx()
-        mi = self._multiindex(i)
-        return self.arr_.__getitem__(mi)
+        key = self._normalize_key(key)
+
+        if isinstance(key, int) and not isinstance(key, bool):
```
vlad-perevezentsev: `_validate_key` in `_normalize_key` already rejects every boolean key so `not isinstance(key, bool)` can never fire

Same on 179 line

dpnp/dpnp_flatiter.py
```diff
@@ -28,63 +28,198 @@
...
+            )
+        elif hasattr(key, "dtype") and dpnp.issubdtype(key.dtype, dpnp.bool):
+            # boolean mask fast path
+            mask = dpnp.asarray(key, sycl_queue=exec_q, usm_type=usm_type)
+            idx = dpnp.nonzero(mask)[0]
+        else:
+            flat_index = dpnp.arange(
+                a.size, sycl_queue=exec_q, usm_type=usm_type
+            )
+            idx = flat_index[key]
```
vlad-perevezentsev: `_validate_key` already builds the index array and checks its min/max, so `arange(a.size)` + a gather just recomputes what's known — a `size`-sized temporary for nothing. Wrapping negatives isn't needed either: `dpnp.put` uses `mode="wrap"`, which handles in-range negative indices, and the bounds check has already rejected the rest.

Suggest returning `idx` from `_validate_key`, passing it through `_normalize_key`, and making the `else` in `__setitem__` an `elif idx is None`: so `arange` is only the fallback for `Ellipsis/()`.

 a.flat[[1, -2, 3]] = 0 on a 128 MiB array: 0.86 → 0.38 ms.

```
    def _validate_key(self, key):
        """
        Validate `key` as an index of the flat iterator.

        Returns the array of flat positions, if `key` is an integer array, and
        ``None`` otherwise, i.e. if the positions have to be resolved by the
        caller.

        """
        ...
        if lo < -size:
            raise IndexError(f"index {lo} is out of bounds for size {size}")
        return idx

    def _normalize_key(self, key):
        ...
        return key, self._validate_key(key)

    def __getitem__(self, key):
        key, _ = self._normalize_key(key)
        ...

    def __setitem__(self, key, val):
        key, idx = self._normalize_key(key)
        ...
        # resolve key to flat positions, an integer array key is already
        # validated and `dpnp.put` resolves its negative positions
        ...
        elif idx is None:
            # ellipsis, empty tuple or an unrecognized key: let regular
            # indexing resolve the positions and raise on an invalid key
            flat_index = dpnp.arange(
                a.size, sycl_queue=exec_q, usm_type=usm_type
            )
            idx = flat_index[key]
```

dpnp/dpnp_flatiter.py
antonwolfy: Good catch — valid, and it was worse on the setitem side: a 2-D mask whose total size matched passed the size-only check, and the setitem path's `dpnp.nonzero(mask)[0]` then used only the first-axis indices, silently writing wrong positions (NumPy raises).

Fixed: `_validate_key` now rejects `idx.ndim > 1` with the same `IndexError` NumPy raises ("too many indices for flat iterator: flat iterator is 1-dimensional, but N were indexed"), and only a 1-D mask is treated as valid. Added a test (gated numpy>=2.4 — on numpy 2.0 a 2-D bool mask raises ValueError rather than IndexError).

dpnp/dpnp_flatiter.py
antonwolfy: `dpnp.ndarray.flat` setitem is built on `dpnp.put`, which rejects overlapping memory by design, so this is consistent with the rest of dpnp rather than a flat-specific gap.

Matching NumPy here isn't cleanly feasible either: NumPy's *flat* setitem cascades on a backward overlap (`a.flat[1:] = a[:-1]` -> `[0,0,0,0,0,0]`), which a gather/scatter `put` can't reproduce; the only workaround (copying the value) yields snapshot semantics that match NumPy's *regular* assignment but not its flat behavior.

Given that, I'd prefer to keep raising a clear error rather than add a masked copy that still diverges.

dpnp/dpnp_flatiter.py
antonwolfy: Good catch. Advanced indexing (list/array) already allocates a fresh array, so the unconditional `.copy()` was a second full-size copy. Now I materialize `res = reshape(-1)[key]` and copy only when it aliases the source, detected by comparing the base allocation `usm_data._pointer` (offset-independent). This also skips the copy for a basic index on a non-contiguous array, where `reshape(-1)` already copied. Added a non-contiguous case to `test_flat_getitem_returns_copy` to pin the copy semantics on that path.

dpnp/dpnp_flatiter.py
antonwolfy: Agreed. Added a `_asarray_cfd` helper used for both the value and the boolean mask: a device array keeps its own queue (so the downstream `dpnp.put`/`dpnp.nonzero` raises `ExecutionPlacementError` on a queue mismatch), while a host input is placed on the iterator's queue. Verified that `a.flat[0:2] = v` and `a.flat[m] = -1` now raise `ExecutionPlacementError` for a cross-queue `v`/`m`, matching regular assignment, `dpnp.put` and `a.flat[m]` getitem. Added `test_flat_setitem_different_queue` to cover it.

dpnp/dpnp_flatiter.py
antonwolfy: Follow-up correction: the alias check must compare `usm_data._pointer` (the offset-independent allocation base), not the array-level `_pointer`, which is adjusted to the first element — for a slice view (non-zero offset) the latter compared unequal to the source, so the copy was wrongly skipped and leaked a writable view. Extended `test_flat_getitem_returns_copy` to run on a contiguous and a non-contiguous source, which catches exactly that.

dpnp/dpnp_flatiter.py
antonwolfy: Follow-up refinement: the coercion helper is now `_as_dpnp_array`, mirroring the dispatch in `_validate_key` — a `dpnp_array` is returned unchanged, a `usm_ndarray` is wrapped without a copy (both keep their own queue), and only a host input is placed on the iterator's queue. Also split the coverage: the cross-queue negative case moved to `test_sycl_queue.py::test_flat_setitem_diff_queue`, and I added positive persistence tests — `test_sycl_queue.py::test_flat` (result stays on the input queue across get/setitem) and `test_usm_type.py::test_flat` (slice keeps the usm type, advanced index coerces, setitem preserves it).

dpnp/dpnp_flatiter.py
antonwolfy: Done — the guard is now `if val.size not in (0, 1, n):`. Confirmed against NumPy's flatiter assignment (`iter_ass_subscript` in `iterators.c`): it wraps the value in its own iterator and cycles it, resetting when exhausted, so a size-1/0-d value fills the whole selection — exactly what `dpnp.put` produces by broadcasting. Cycling a size-1 value previously built an `arange(n)` + modulo + gather for a result `put` already gives; now that only runs for a genuine mismatch (`val.size > 1 and != n`), where `val[arange(n) % val.size]` reproduces NumPy's wrap. Empty (0) is a no-op on both sides. Added `test_flat_setitem_single_value_broadcasts` to pin the size-1/0-d case to NumPy.

dpnp/dpnp_flatiter.py
antonwolfy: Right, removed in both getitem and setitem. `_normalize_key` runs `_validate_key`, which always raises for a boolean key (a Python `bool` skips the scalar-int branch, gets `numpy.asarray`-ed to bool dtype, and hits the "boolean indices for iterators are not supported" raise), so `key` is never a `bool` at the scalar fast path — the `not isinstance(key, bool)` guard (there because `bool` subclasses `int`) could never fire. Now just `isinstance(key, int)`, with a short comment on why the bool case is already handled.

dpnp/dpnp_flatiter.py
antonwolfy: Done. `_validate_key` now returns the validated positions for an integer-array key (`None` otherwise), and `_normalize_key` — renamed to `_prepare_key` since it now yields both — passes them through. The setitem integer-array path feeds that `idx` straight to `dpnp.put`, dropping the `arange(a.size)` + gather; `arange` is now only the `elif idx is None` fallback for ellipsis / unrecognized keys. Negative wrapping is left to `dpnp.put(mode="wrap")` — `_validate_key` already guarantees indices are in `[-size, size)`, and I confirmed `put` on a raw negative resolves to the same position as the old `arange[key]` gather. Bounds/OOB still raise in `_validate_key` before `put`, and a cross-queue index still raises `ExecutionPlacementError` (now via `put`). Added `test_flat_setitem_int_array_value` (positive and negative indices with a per-element value) to lock the order/wrap against NumPy.

antonwolfy:
Ready for the next review round. Since the last one:

- Addressed all outstanding review comments: the redundant full-size copy in getitem, compute-follows-data in setitem (value and mask), value cycling now deferred to `dpnp.put`'s broadcast, the dead `bool` guard on the scalar fast path, and reuse of the validated index array instead of rebuilding `arange(size)`.
- Ran a full self-review pass on top: aligned the wrong-size boolean-mask error message with NumPy, expanded test coverage (n>1 cycling, in-bounds empty selection, non-integer array index, 1-element-tuple assignment, and queue/usm_type placement in `test_sycl_queue`/`test_usm_type`), noted the assignment fixes in the changelog, and minor tidy-ups.
- Verified the `numpy>=2.4` gates against numpy 2.0.2: all are required and no ungated test needs a floor.

All flatiter tests pass locally.

PR #3048, opened 2026-08-29
Add CI workflow to build and test dpnp with the open source LLVM SYCL compiler

antonwolfy:
This PR adds a manually-triggered CI workflow that builds and tests dpnp against the open source `intel/llvm` DPC++ (LLVM SYCL) nightly toolchain, instead of the released Intel oneAPI compiler. This gives early warning when an upstream compiler change breaks dpnp, and exercises the source against a second, independent SYCL implementation. Building against the open source nightly surfaced one real correctness bug in dpnp (fixed here) and required a few build-system adjustments, all included.

### What is added

- `.github/workflows/os-llvm-sycl-build.yml` — `workflow_dispatch` job that downloads the latest `intel/llvm` nightly compiler, the OpenCL CPU runtime (oclcpuexp) and oneTBB, builds dpctl and dpnp from source against the nightly (dpnp built against the oneMath interface), and runs the full `dpnp/tests` suite.
- `.github/workflows/bump-sycl-deps.yml` — weekly (Saturday) workflow that opens a PR when a newer oclcpuexp driver release or oneTBB release is available, keeping the pins in `os-llvm-sycl-build.yml` current. The nightly compiler itself is discovered at run time and is not pinned.
- `dpnp/backend/cmake/Modules/IntelSYCLConfig.cmake` — parse the open source compiler version string (`clang version X.Y.Z`) in the vendored module, which is forced for this build because the nightly does not ship `add_sycl_to_target()`.

### Correctness fix

Building the ufunc extension with the open source nightly exposed an ODR violation in `dpnp/backend/extensions/ufunc/elementwise_functions/populate.hpp`: the `MACRO_POPULATE_DISPATCH_*` macros defined the helper structs `ContigFactory`, `StridedFactory` and `TypeMapFactory` with the same name in the shared `dpnp::extensions::ufunc::impl` namespace across every per-ufunc translation unit, each with a different body. The Intel compiler inlines these away, but the open source compiler emits them as weak COMDAT symbols and folds the identically-named instantiations across translation units, so several binary ufuncs got their dispatch and result-type tables populated from the wrong translation unit. Observable effect: `fmin`, `fmod`, `heaviside`, `logaddexp2` and `ldexp` silently computed the elementwise maximum, and the integer result-type entries for `fmin`, `fmax`, `fmod`, `gcd` and `lcm` were dropped. The fix gives the factories per-ufunc-unique names so no two translation units share a symbol.

### DIFF
--- a/.github/workflows/bump-sycl-deps.yml
+++ b/.github/workflows/bump-sycl-deps.yml
@@ -44,6 +44,9 @@ jobs:
           releases-only: true
           # Driver releases are tagged like "2026-WW28"; ignore nightly-* tags.
           regex: '^\d{4}-WW\d+$'
+          # Pick the most recently published matching release. Must stay false:
+          # sort-tags: true would semver-sort and mis-parse the YYYY-WWNN tag.
+          sort-tags: false
 
       - name: Find latest oneTBB release tag
         id: tbb
@@ -52,13 +55,16 @@ jobs:
           repository: uxlfoundation/oneTBB
           releases-only: true
           regex: '^v\d+\.\d+\.\d+$'
+          # Most recently published matching release (see note above).
+          sort-tags: false
 
       - name: Resolve asset names and derive values
         id: resolve
+        env:
+          DRIVER_TAG: ${{ steps.driver.outputs.tag }}
+          TBB_TAG: ${{ steps.tbb.outputs.tag }}
         run: |
           set -euo pipefail
-          DRIVER_TAG='${{ steps.driver.outputs.tag }}'
-          TBB_TAG='${{ steps.tbb.outputs.tag }}'
 
           # The oclcpuexp filename embeds its own version and is not derivable
           # from the tag, so read it back from the driver release assets.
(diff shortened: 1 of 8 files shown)


antonwolfy:
Both new workflows are passing:
https://github.com/antonwolfy/dpnp/actions/runs/33247287920
https://github.com/antonwolfy/dpnp/actions/runs/33247311140

.github/workflows/os-llvm-sycl-build.yml
```diff
@@ -0,0 +1,197 @@
+name: Build with Open Source LLVM SYCL compiler
+
+on:
+  # To be able to be triggered manually
+  workflow_dispatch:
```
ndgrigorian, line 9: Is there a reason to trigger manually vs just doing with each PR? Or we could run nightly?

.github/workflows/bump-sycl-deps.yml
```diff
@@ -0,0 +1,139 @@
+name: Bump open source SYCL build deps
```
ndgrigorian, line 1: would be nice to add this same chron job to dpctl

.github/workflows/os-llvm-sycl-build.yml
antonwolfy: It's the leftover while I was troubleshooting the workflow on fork.
Thank you for noticing that

.github/workflows/bump-sycl-deps.yml
antonwolfy: Yes, I'm planning to do that. But firstly wanted to ensure it works properly in dpnp, I'm wondering if there is no permission issue (about which one the Claude warned me).
So planning to wait till the first bumping PR triggered successfully.

ndgrigorian (approved):
Seems to be working so LGTM

PR #3055, opened 2026-09-02
Fix nonzero bool nonstandard bytes

### LINKED ISSUE #3054: Boolean arrays with elements are stored as bytes other than 0 or 1 give wrong results
```python3
import numpy as np
import dpnp

a = np.array([0, 1, 2, 255, 3, 0], dtype=np.uint8)
b = np.array([1, 1, 0, 2, 0, 7], dtype=np.uint8)
na, nb = a.view(bool), b.view(bool)
da = dpnp.asarray(a).view(dpnp.bool)
db = dpnp.asarray(b).view(dpnp.bool)

print(np.equal(na, nb))        # [False  True False  True False False]
print(dpnp.equal(da, db))      # [False  True False False False False]

print(np.argmax(na), int(dpnp.argmax(da)))          # 1  vs  4
print(np.unique(na).size, dpnp.unique(da).size)     # 2  vs  4
print(np.isin(na, nb))         # all True
print(dpnp.isin(da, db))       # [ True  True False False False  True]
```

output:
```bash
[False  True False  True False False]
[False  True False False False False]
1 4
2 2
[ True  True  True  True  True  True]
[ True  True  True False False  True]
```

abagusetty:
Boolean arrays may contain non-zero bytes like `0x02` or `0xFF` that NumPy treats as True but dpnp does not. 

Fixes #3054 

### DIFF
--- a/dpnp/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpnp/tensor/libtensor/include/kernels/accumulators.hpp
@@ -74,9 +74,17 @@ struct NonZeroIndicator
     {
         static constexpr outputT out_one(1);
         static constexpr outputT out_zero(0);
-        static constexpr inputT val_zero(0);
 
-        return (val == val_zero) ? out_zero : out_one;
+        if constexpr (std::is_same_v<inputT, bool>) {
+            // NumPy treats any non-zero byte as True, while a plain
+            // comparison may fold into a raw byte load, see gh-2121
+            const std::uint8_t &u = sycl::bit_cast<std::uint8_t>(val);
+            return (u == std::uint8_t{0}) ? out_zero : out_one;
+        }
+        else {
+            static constexpr inputT val_zero(0);
+            return (val == val_zero) ? out_zero : out_one;
+        }
     }
 };
 
@@ -1301,8 +1309,10 @@ struct Cumsum1DContigFactory
     {
         if constexpr (std::is_integral_v<T>) {
             using cumsumT = std::int64_t;
-            fnT fn =
-                cumsum_val_contig_impl<T, cumsumT, NoOpTransformer<cumsumT>>;
+            // CastTransformer, not NoOpTransformer: an implicit bool
+            // conversion would read the raw byte, see gh-2121
+            fnT fn = cumsum_val_contig_impl<T, cumsumT,
+                                            CastTransformer<T, cumsumT>>;
             return fn;
         }
         else {
@@ -1419,8 +1429,10 @@ struct Cumsum1DStridedFactory
     {
         if constexpr (std::is_integral_v<T>) {
             using cumsumT = std::int64_t;
-            fnT fn =
-                cumsum_val_strided_impl<T, cumsumT, NoOpTransformer<cumsumT>>;
+            // CastTransformer, not NoOpTransformer: an implicit bool
+            // conversion would read the raw byte, see gh-2121
+            fnT fn = cumsum_val_strided_impl<T, cumsumT,
+                                             CastTransformer<T, cumsumT>>;
             return fn;
         }
         else {
--- a/dpnp/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpnp/tensor/libtensor/include/kernels/accumulators.hpp
@@ -1311,8 +1311,8 @@ struct Cumsum1DContigFactory
             using cumsumT = std::int64_t;
             // CastTransformer, not NoOpTransformer: an implicit bool
             // conversion would read the raw byte, see gh-2121
-            fnT fn = cumsum_val_contig_impl<T, cumsumT,
-                                            CastTransformer<T, cumsumT>>;
+            fnT fn =
+                cumsum_val_contig_impl<T, cumsumT, CastTransformer<T, cumsumT>>;
             return fn;
         }
         else {
--- a/dpnp/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpnp/tensor/libtensor/include/kernels/accumulators.hpp
@@ -76,8 +76,8 @@ struct NonZeroIndicator
         static constexpr outputT out_zero(0);
 
         if constexpr (std::is_same_v<inputT, bool>) {
-            // NumPy treats any non-zero byte as True, while a plain
-            // comparison may fold into a raw byte load, see gh-2121
+            // NumPy treats any non-zero byte as True; read the raw byte
+            // rather than the bool value, see gh-2121
             const std::uint8_t u = sycl::bit_cast<std::uint8_t>(val);
             return (u == std::uint8_t{0}) ? out_zero : out_one;
         }
--- a/dpnp/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpnp/tensor/libtensor/include/kernels/accumulators.hpp
@@ -78,7 +78,7 @@ struct NonZeroIndicator
         if constexpr (std::is_same_v<inputT, bool>) {
             // NumPy treats any non-zero byte as True, while a plain
             // comparison may fold into a raw byte load, see gh-2121
-            const std::uint8_t &u = sycl::bit_cast<std::uint8_t>(val);
+            const std::uint8_t u = sycl::bit_cast<std::uint8_t>(val);
             return (u == std::uint8_t{0}) ? out_zero : out_one;
         }
         else {
--- a/dpnp/tensor/libtensor/include/kernels/accumulators.hpp
+++ b/dpnp/tensor/libtensor/include/kernels/accumulators.hpp
@@ -93,7 +93,12 @@ struct NoOpTransformer
 {
     constexpr NoOpTransformer() {}
 
-    T operator()(const T &val) const { return val; }
+    // bool is normalized: a byte other than 0x00/0x01 would otherwise reach
+    // the scan operation and leak into the result, see gh-2121
+    T operator()(const T &val) const
+    {
+        return dpnp::tensor::type_utils::normalize_bool(val);
+    }
 };
 
 template <typename srcTy, typename dstTy>
--- a/dpnp/tensor/libtensor/include/kernels/elementwise_functions/common.hpp
+++ b/dpnp/tensor/libtensor/include/kernels/elementwise_functions/common.hpp
@@ -46,6 +46,7 @@
 #include "utils/offset_utils.hpp"
 #include "utils/sycl_alloc_utils.hpp"
 #include "utils/sycl_utils.hpp"
+#include "utils/type_utils.hpp"
 
 #include "kernels/alignment.hpp"
 #include "kernels/dpnp_tensor_types.hpp"
@@ -59,6 +60,7 @@ using dpnp::tensor::kernels::alignment_utils::required_alignment;
 
 using dpnp::tensor::sycl_utils::sub_group_load;
 using dpnp::tensor::sycl_utils::sub_group_store;
+using dpnp::tensor::type_utils::normalize_bool;
 
 /*! @brief Functor for unary function evaluation on contiguous array */
 template <typename argT,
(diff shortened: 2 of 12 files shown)


ndgrigorian (approved):
I've looked over the changes, seems like a welcome fix for a previously unnoticed bug, LGTM

ndgrigorian:
@abagusetty seems that the tests fail with the open-source compiler, interestingly enough. Possible bug in the nightly DPC++?

dpnp/tensor/libtensor/include/kernels/elementwise_functions/common.hpp
```diff
@@ -148,7 +152,7 @@ struct UnaryContigFunctor
                 const std::size_t lane_id = sg.get_local_id()[0];
                 for (std::size_t k = base + lane_id; k < nelems_; k += sgSize) {
                     // scalar call
-                    out[k] = op(in[k]);
+                    out[k] = op(normalize_bool(in[k]));
```
antonwolfy: It seems not needed here, since bool type was excluded above

abagusetty:
> @abagusetty seems that the tests fail with the open-source compiler, interestingly enough. Possible bug in the nightly DPC++?

No bug in nightly. Interestingly, it is doing great. The issue was UB in the PR that the two compilers are treating it differently. A bool whose byte isnt `0x00`/`0x01` has an intermediate value that the compiler is free to assume 0 or 1. 2026.1 seems to just hide it. I did apply the same normalization of the bytes that was missed to the site it was failing to comply.

ndgrigorian (approved):

dpnp/tensor/libtensor/include/utils/type_utils.hpp
```diff
@@ -60,11 +60,28 @@ struct is_complex<
...
 
+// NumPy reads any non-zero byte of a bool as True, while a plain comparison
+// may fold into a raw byte load, see gh-2121
+template <typename T>
+T normalize_bool(const T &v)
+{
+    if constexpr (std::is_same_v<T, bool>) {
+        // read the storage as a byte: a bool copy would let the compiler
+        // assume a 0/1 value and fold this away
+        const std::uint8_t u = *reinterpret_cast<const std::uint8_t *>(&v);
```
antonwolfy: Should we use `sycl::bit_cast<std::uint8_t>` as in `convert_impl` below?

dpnp/tensor/libtensor/include/kernels/elementwise_functions/common.hpp
```diff
@@ -456,7 +464,7 @@ struct BinaryContigFunctor
             else {
                 const std::size_t lane_id = sg.get_local_id()[0];
                 for (std::size_t k = base + lane_id; k < nelems_; k += sgSize) {
-                    out[k] = op(in1[k], in2[k]);
+                    out[k] = op(normalize_bool(in1[k]), normalize_bool(in2[k]));
```
antonwolfy: Seems no needed, as input array is never bool here

dpnp/tensor/libtensor/include/kernels/elementwise_functions/common_inplace.hpp
```diff
@@ -123,7 +127,7 @@ struct BinaryInplaceContigFunctor
             else {
                 const std::size_t lane_id = sg.get_local_id()[0];
                 for (std::size_t k = base + lane_id; k < nelems_; k += sgSize) {
-                    op(lhs[k], rhs[k]);
+                    op(lhs[k], normalize_bool(rhs[k]));
```
antonwolfy: The same, not need to call `normalize_bool` here

abagusetty: Perfect, removed it!

dpnp/tensor/libtensor/include/kernels/elementwise_functions/common.hpp
abagusetty: done!

dpnp/tensor/libtensor/include/utils/type_utils.hpp
abagusetty: Thanks! I contemplated on using `sycl::bit_cast<std::uint8_t>` but thought about the support for CPU backend support (via opencl:cpu or native_cpu) and fell back to using `std::uint8_t`. Now that `sycl::bit_cast` support is robust, cleaner to have it

dpnp/tensor/libtensor/include/kernels/elementwise_functions/common.hpp
abagusetty: good catch! thanks

dpnp/tensor/libtensor/include/kernels/elementwise_functions/common_inplace.hpp
```diff
@@ -59,6 +60,7 @@ using dpnp::tensor::kernels::alignment_utils::required_alignment;
 
 using dpnp::tensor::sycl_utils::sub_group_load;
 using dpnp::tensor::sycl_utils::sub_group_store;
+using dpnp::tensor::type_utils::normalize_bool;
```
antonwolfy, line 63: It seems there is one more gap with `BitwiseXorInplaceFunctor` in `dpnp/tensor/libtensor/include/kernels/elementwise_functions/bitwise_xor.hpp:300`:
```cpp
        if constexpr (std::is_same_v<resT, bool>) {
            res = (res != in);
        }
```
The in-place functor normalizes only the RHS (`op(lhs[k], normalize_bool(rhs[k]))`); res/lhs is read raw; `!= is` value-use, so a non-canonical destination byte gives the wrong answer:
```python
import numpy as np, dpnp

def bviews(u8):                      # (numpy bool view, dpnp bool view) of same bytes
    a = np.array(u8, dtype=np.uint8)
    return a.view(bool), dpnp.asarray(a).view(dpnp.bool)

# non-canonical lhs (0x02=True), canonical rhs (0x01=True); True ^ True = False
na, da = bviews([2])
nb, db = bviews([1])

# out-of-place xor: PR normalizes BOTH inputs -> expect PASS on PR branch
print("a ^ b   numpy:", np.bitwise_xor(na, nb),
      " dpnp:", dpnp.asnumpy(dpnp.bitwise_xor(da, db)))
# Out: a ^ b   numpy: [False]  dpnp: [False]

# in-place xor: PR normalizes only rhs, lhs read raw via (res != in) -> expect FAIL
na2, da2 = bviews([2]); _, db2 = bviews([1])
na2 ^= nb
da2 ^= db2
print("a ^= b  numpy:", na2, " dpnp:", dpnp.asnumpy(da2))
# Out: a ^= b  numpy: [False]  dpnp: [ True]
```

So we probably need to update `BitwiseXorInplaceFunctor` with:
```cpp
if constexpr (std::is_same_v<resT, bool>) {
    res = (normalize_bool(res) != in);   // in already normalized by caller; keeps output canonical
}
```
and to cover with a test.

dpnp/tensor/libtensor/include/utils/rich_comparisons.hpp
```diff
@@ -116,6 +118,26 @@ inline constexpr bool is_fp_v =
     (std::is_same_v<T, sycl::half> || std::is_same_v<T, float> ||
      std::is_same_v<T, double>);
 
+// takes by reference: copying a bool first would let the compiler assume a
+// 0/1 byte and fold the normalization away, see gh-2121
```
antonwolfy: The comment does not seem correct for value merge/insertion sort, which copies elements into `auto` locals before the comparator runs (`merge_sort.hpp` lines ~83–84, ~228).

The reference binds to those copies, not memory. It's harmless (the in-comparator `normalize_bool` is what makes sort correct, and CI's sort tests pass), but the rationale is wrong and only true for the argsort path.

We should either correct or drop the comment.

abagusetty: Removed it!

dpnp/tensor/libtensor/include/kernels/elementwise_functions/common_inplace.hpp
abagusetty: Fixed it in the caller (`apply_inplace` in `common_inplace.hpp`) instead, since `+=` and `*=` leaks the sameway. Happy to move it if you would prefer a per-functor solution

dpnp/tensor/libtensor/include/kernels/accumulators.hpp
```diff
@@ -74,9 +74,17 @@ struct NonZeroIndicator
     {
         static constexpr outputT out_one(1);
         static constexpr outputT out_zero(0);
-        static constexpr inputT val_zero(0);
 
-        return (val == val_zero) ? out_zero : out_one;
+        if constexpr (std::is_same_v<inputT, bool>) {
+            // NumPy treats any non-zero byte as True, while a plain
+            // comparison may fold into a raw byte load, see gh-2121
```
antonwolfy:
```suggestion
            // NumPy treats any non-zero byte as True; read the raw byte
            // rather than the bool value, see gh-2121
```

dpnp/tensor/libtensor/include/kernels/accumulators.hpp
```diff
@@ -74,9 +74,17 @@ struct NonZeroIndicator
...
+            const std::uint8_t &u = sycl::bit_cast<std::uint8_t>(val);
```
antonwolfy: no need to bind a reference to a `bit_cast` prvalue:
```suggestion
            const std::uint8_t u = sycl::bit_cast<std::uint8_t>(val);
```

dpnp/tensor/libtensor/include/utils/type_utils.hpp
```diff
@@ -60,11 +60,28 @@ struct is_complex<
 template <typename T>
 inline constexpr bool is_complex_v = is_complex<T>::value;
 
+// NumPy reads any non-zero byte of a bool as True, while a plain comparison
+// may fold into a raw byte load, see gh-2121
```
antonwolfy:
```suggestion
// NumPy reads any non-zero bool byte as True; a non-canonical byte
// (not 0x00/0x01) used as a value would disagree, see gh-2121
```

dpnp/tensor/libtensor/include/utils/type_utils.hpp
```diff
@@ -60,11 +60,28 @@ struct is_complex<
...
+template <typename T>
+T normalize_bool(const T &v)
+{
+    if constexpr (std::is_same_v<T, bool>) {
+        // read the storage as a byte: a bool copy would let the compiler
+        // assume a 0/1 value and fold this away
```
antonwolfy:
```suggestion
        // read the raw storage byte and test non-zero; using a
        // non-canonical bool as a value is unreliable (UB)
```

dpnp/tests/test_logic.py
```diff
@@ -900,3 +902,140 @@ def test_isin_errors(self):
...
+
+    def test_result_has_canonical_bytes(self):
+        # a computed bool result must only ever hold 0x00 or 0x01; `sort` and
+        # `unique` are excluded, they select elements and so carry the bytes
+        _, ia = self._views([0, 1, 2, 255, 3, 0])
+        _, ib = self._views([1, 1, 0, 2, 0, 7])
+
+        for res in (
+            dpnp.equal(ia, ib),
+            dpnp.logical_not(ia),
```
antonwolfy, line 1000: It'd make sense to extend then:
```suggestion
            dpnp.not_equal(ia, ib),
            dpnp.logical_not(ia),
            dpnp.logical_and(ia, ib),
            dpnp.logical_or(ia, ib),
            dpnp.logical_xor(ia, ib),
            dpnp.bitwise_and(ia, ib),
            dpnp.bitwise_or(ia, ib),
            dpnp.bitwise_xor(ia, ib),
```

dpnp/tests/test_logic.py
```diff
@@ -900,3 +902,140 @@ def test_isin_errors(self):
...
+        a, ia = self._views([0, 1, 2, 255, 3, 0])
+
+        assert_array_equal(dpnp.unique(ia), numpy.unique(a))
+
+    def test_isin(self):
+        a, ia = self._views([0, 1, 2, 255, 3, 0])
+        b, ib = self._views([1, 1, 0, 2, 0, 7])
+
+        assert_array_equal(dpnp.isin(ia, ib), numpy.isin(a, b))
+
```
antonwolfy, line 1042: and add `searchsorted`:
```suggestion

    def test_searchsorted(self):
        a, ia = self._views([0, 0, 1, 1])
        v, iv = self._views([2, 255, 0])

        assert_array_equal(dpnp.searchsorted(ia, iv), numpy.searchsorted(a, v))

```

antonwolfy (approved):
Thank you @abagusetty for resolving that, only few more minor nits, but overall LGTM.

abagusetty:
> Thank you @abagusetty for resolving that, only few more minor nits, but overall LGTM.

Thanks @antonwolfy !

PR #3059, opened 2026-09-03
Use `numpydoc` to render the API reference

vlad-perevezentsev:
This PR switches the documentation build from `sphinx.ext.napoleon` to `numpydoc` to render the API reference. 
`numpydoc` renders the `Parameters` and `Returns` sections as definition lists out of the box, which lets us remove:

- the custom `doc/_static/dpnp-custom.js` that reformatted parameter
  descriptions, together with the CSS rules that only supported it;
- the napoleon monkeypatch for namedtuple `Returns` (numpydoc handles it
  natively).

A small hook keeps the dpnp-specific `Limitations` section and a some docstrings are adjusted for numpydoc's stricter parser.


antonwolfy:
Small rendering mismatch vs the legacy build worth a fix: parameter/return **types** now render upright + semibold instead of *italic*.

- New: https://intelpython.github.io/dpnp/pull/3059/reference/generated/dpnp.clip.html
- Legacy: https://intelpython.github.io/dpnp/reference/generated/dpnp.clip.html

Cause is purely CSS. numpydoc emits the type as `<span class="classifier">`, and the italic comes from Sphinxs built-in `basic.css` (`.classifier { font-style: oblique }`). NumPys pydata-sphinx-theme `@import`s `basic.css` so types are italic, and napoleon previously wrapped the type in `<em>`. Furo neither imports `basic.css` nor styles `.classifier`, and it makes the whole `<dt>` term semibold, so the type ends up bold/upright.

Proposed fix, add to `doc/_static/dpnp-custom.css`:

```css
/* numpydoc param/return types: italic like NumPy, not bold */
.classifier {
    font-style: italic;
    font-weight: 400;
}
```

doc/conf.py
```diff
@@ -238,66 +240,64 @@ def _can_document_member(member, *args, **kwargs):
...
-# Napoleon settings
-napoleon_use_ivar = True
-napoleon_include_special_with_doc = True
-napoleon_custom_sections = ["limitations"]
-
-
-# Napoleon extension can't properly render "Returns" section in case of
-# namedtuple as a return type. That patch proposes to extend the parse logic
-# which allows text in a header of "Returns" section.
-def _parse_returns_section_patched(self, section: str) -> list[str]:
```
antonwolfy, line 250: Please edit the docstrings so `Returns` contains only `name : type` items, relocating the intro sentence. Example for `qr`:
Before:
```
    Returns
    -------
    When mode is "reduced" or "complete", the result will be a namedtuple with
    the attributes `Q` and `R`:

    Q : dpnp.ndarray of float or complex, optional
        ...
```
After (move the sentence into a Notes section — or the extended summary):
```
    Returns
    -------
    Q : dpnp.ndarray of float or complex, optional
        ...
    R : dpnp.ndarray of float or complex, optional
        ...
    (h, tau) : tuple of dpnp.ndarray of float or complex, optional
        ...

    Notes
    -----
    When `mode` is ``"reduced"`` or ``"complete"``, the result is a namedtuple
    with the attributes ``Q`` and ``R``. When `mode` is ``"raw"``, it returns
    ``(h, tau)``.
```
Same pattern for the other 8 (for the `unique_*`/`eig`/`eigh`/`slogdet`/`lu` cases, the intro is just "A namedtuple with the following attributes:" — that line can simply be deleted, since listing the named returns already conveys it).

vlad-perevezentsev: Good catch, fixed
Thanks!

doc/conf.py
antonwolfy: `dpnp-custom.js` is removing by that PR

vlad-perevezentsev: Done, thanks

PR #3063, opened 2026-09-08
Fix row corruption in batched multi-level cumulative scan

antonwolfy:
## Summary

`dpnp.cumsum`, `dpnp.cumprod`, and their `nan`/`cumulative_*` variants (including `dpnp.tensor.cumulative_sum`/`cumulative_prod`) could silently return incorrect results when accumulating along an axis of a multi-row array, with no error raised. The corruption is triggered when the accumulation axis is long enough to require three or more reduction levels in the internal block-scan (axis length greater than `chunk_size ** 2`, where `chunk_size = wg_size * n_wi`, i.e. roughly a million elements on GPU and about four million on CPU) and the array has more than one batch row. The elementwise total is conserved but redistributed across rows.

## Root cause

In the batched scan driver `inclusive_scan_iter` (`accumulators.hpp`), the intermediate block-scan fixup `update_local_chunks` passed a `NoOpIndexer` for the per-row (iter) offset into the `src` buffer. The intermediate block-scan buffers, however, are laid out strided per row with a stride of `src_size`. The `local_scans` read already used the correct `iter_gid * local_stride` offset, but the `src` write offset was unstrided, so rows overwrote each other. This cancels out only when `iter_nelems == 1` (a single row) or when fewer than three levels are needed, which is why the bug went unnoticed.

## Fix

Give `update_local_chunks` a `Strided1DIndexer{iter_nelems, src_size}` for the row offset instead of the `NoOpIndexer` (the within-row `out_indexer` stays `NoOp`, since the level buffers are contiguous within a row).

### DIFF
--- a/dpnp/tests/tensor/test_tensor_accumulation.py
+++ b/dpnp/tests/tensor/test_tensor_accumulation.py
@@ -128,8 +128,8 @@ def test_strided_cumsum_axis_sint(dt):
 
 @pytest.mark.parametrize("func", ["cumulative_sum", "cumulative_prod"])
 def test_batched_multilevel_scan(func):
-    # axis > chunk_size**2 (chunk_size <= 2048) needs >=3 scan levels; with >1
-    # batch row the intermediate update mis-strided rows. 5e6 forces that path.
+    # Regression test for gh-3063: multi-row scan over an axis needing >=3
+    # levels (axis > chunk_size**2, chunk_size <= 2048) mis-strided rows.
     get_queue_or_skip()
     n0, n1 = 3, 5_000_000
     x = dpt.ones((n0, n1), dtype="i1")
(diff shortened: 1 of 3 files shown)


dpnp/tests/tensor/test_tensor_accumulation.py
```diff
@@ -126,6 +126,23 @@ def test_strided_cumsum_axis_sint(dt):
     assert dpt.all(res == dpt.expand_dims(expected, axis=1))
 
 
+@pytest.mark.parametrize("func", ["cumulative_sum", "cumulative_prod"])
+def test_batched_multilevel_scan(func):
```
ndgrigorian, line 130: Can we note this to be a regression test?

ndgrigorian (approved):
LGTM other than minor nit

ndgrigorian (approved):

PR #3066, opened 2026-09-10
Sort NaNs to the end for descending order

antonwolfy:
`dpnp.sort`, `dpnp.argsort`, and their `dpnp.ndarray`/`dpnp.tensor` counterparts placed `NaN` values at the *beginning* when sorting in descending order, while NumPy 2.5 keeps `NaN` values at the *end* for both ascending and descending order.

For descending order the sort kernels simply reversed the ascending comparison, which also moved `NaN` (treated as the largest value) to the front.

The PR proposes to change:
- Merge-sort comparators: only the comparison between non-`NaN` values is reversed for descending order, so `NaN`s stay at the end. For complex values the `NaN`-ness groups `(no nan) -> (imag nan) -> (real nan) -> (all nan)` keep the same trailing order as ascending, and only the finite-component lexicographic comparison is reversed.
- Radix-sort float casts (half / float32 / float64): map `NaN` to the maximum key, so it lands in the last bucket regardless of direction. This keeps the descending float radix fast path intact rather than falling back to a slower comparator.
- `dpnp.tensor.top_k` inherits the change through the shared descending comparator: `NaN` values (and complex values with a `NaN` component) are now treated as the smallest, so `mode="largest"` no longer returns them ahead of finite values.

`NaN` ordering under `sort`/`argsort` is left implementation-defined by the Python Array API standard, and `top_k` explicitly permits sorting `NaN` values to either end. The new "`NaN` last" behavior is therefore spec-conforming and is chosen to match NumPy 2.5; it is now consistent across the `dpnp` and `dpnp.tensor` namespaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,7 @@ This release is compatible with NumPy 2.5.
 * Fixed `dpnp.cumsum`, `dpnp.cumprod`, and their `nan`/`cumulative_*` variants (including `dpnp.tensor.cumulative_sum`/`cumulative_prod`) silently returning incorrect results when accumulating along an axis of an array with more than one row [#3063](https://github.com/IntelPython/dpnp/pull/3063)
 * Fixed `dpnp.einsum` returning a result whose memory layout differs from NumPy for the default `order="K"`, and ignoring `out` and `order` for a contraction over a size-0 dimension [#3058](https://github.com/IntelPython/dpnp/pull/3058)
 * Fixed operations on a boolean array whose bytes are not `0x00`/`0x01` [#3055](https://github.com/IntelPython/dpnp/pull/3055)
+* Fixed `dpnp.sort`, `dpnp.argsort`, and their `dpnp.ndarray`/`dpnp.tensor` counterparts placing `NaN` values first instead of last when sorting in descending order
 
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ This release is compatible with NumPy 2.5.
 * `dpnp` uses pybind11 3.1.0 [#3015](https://github.com/IntelPython/dpnp/pull/3015)
 * Reworked the ASV benchmarks and added end-to-end workload benchmarks derived from dpBench [#2996](https://github.com/IntelPython/dpnp/pull/2996)
 * Reduced allocations in `dpnp.linalg.norm` by reusing the reduction result as the `sqrt` output buffer in the 2-norm and Frobenius-norm branches [#3062](https://github.com/IntelPython/dpnp/pull/3062)
-* Changed `dpnp.sort`, `dpnp.argsort`, and their `dpnp.ndarray`/`dpnp.tensor` counterparts placing `NaN` values first instead of last when sorting in descending order [#3066](https://github.com/IntelPython/dpnp/pull/3066)
+* Changed `dpnp.sort`, `dpnp.argsort`, and their `dpnp.ndarray`/`dpnp.tensor` counterparts to place `NaN` values last instead of first when sorting in descending order [#3066](https://github.com/IntelPython/dpnp/pull/3066)
 * `dpnp.tensor.top_k` now treats `NaN` values (and complex values with a `NaN` component) as the smallest, consistent with `dpnp.tensor.sort`, so `mode="largest"` no longer returns them ahead of finite values [#3066](https://github.com/IntelPython/dpnp/pull/3066)
 
 ### Deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ This release is compatible with NumPy 2.5.
 * Reworked the ASV benchmarks and added end-to-end workload benchmarks derived from dpBench [#2996](https://github.com/IntelPython/dpnp/pull/2996)
 * Reduced allocations in `dpnp.linalg.norm` by reusing the reduction result as the `sqrt` output buffer in the 2-norm and Frobenius-norm branches [#3062](https://github.com/IntelPython/dpnp/pull/3062)
 * Changed `dpnp.sort`, `dpnp.argsort`, and their `dpnp.ndarray`/`dpnp.tensor` counterparts to place `NaN` values last instead of first when sorting in descending order [#3066](https://github.com/IntelPython/dpnp/pull/3066)
-* `dpnp.tensor.top_k` now treats `NaN` values (and complex values with a `NaN` component) as the smallest, consistent with `dpnp.tensor.sort`, so `mode="largest"` no longer returns them ahead of finite values [#3066](https://github.com/IntelPython/dpnp/pull/3066)
+* Changed `dpnp.tensor.top_k` with `mode="largest"` to no longer return `NaN` values (or complex values with a `NaN` component) ahead of finite values, matching the `NaN`-last order of `dpnp.tensor.sort` [#3066](https://github.com/IntelPython/dpnp/pull/3066)
 
 ### Deprecated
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ This release is compatible with NumPy 2.5.
 * `dpnp` uses pybind11 3.1.0 [#3015](https://github.com/IntelPython/dpnp/pull/3015)
 * Reworked the ASV benchmarks and added end-to-end workload benchmarks derived from dpBench [#2996](https://github.com/IntelPython/dpnp/pull/2996)
 * Reduced allocations in `dpnp.linalg.norm` by reusing the reduction result as the `sqrt` output buffer in the 2-norm and Frobenius-norm branches [#3062](https://github.com/IntelPython/dpnp/pull/3062)
+* Changed `dpnp.sort`, `dpnp.argsort`, and their `dpnp.ndarray`/`dpnp.tensor` counterparts placing `NaN` values first instead of last when sorting in descending order [#3066](https://github.com/IntelPython/dpnp/pull/3066)
```
vlad-perevezentsev: Reads as if NaNs are now placed first
Should be `placing `NaN` values last instead of first`, right?

dpnp/tensor/_sorting.py
```diff
@@ -315,6 +319,10 @@ def top_k(x, k, /, *, axis=None, mode="largest"):
             - `"largest"`: return the `k` largest elements.
             - `"smallest"`: return the `k` smallest elements.
 
+            NaN values (and complex values with a NaN component) are treated
+            as the smallest, so `"largest"` does not return them ahead of
```
vlad-perevezentsev: This isn't accurate for `mode="smallest"` because `NaNs` are ordered to the end in both modes, so in `"smallest"` mode they behave as the largest, not the smallest.
```
In [7]: x = dpnp.tensor.asarray([3., 1., dpnp.nan, 2., dpnp.nan, 5.])

In [8]: dpnp.tensor.top_k(x, 2, mode="smallest")
Out[8]: TopKResult(values=usm_ndarray([1., 2.]), indices=usm_ndarray([1, 3])) # NaNs not returned
```

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ This release is compatible with NumPy 2.5.
...
+* `dpnp.tensor.top_k` now treats `NaN` values (and complex values with a `NaN` component) as the smallest, consistent with `dpnp.tensor.sort`, so `mode="largest"` no longer returns them ahead of finite values [#3066](https://github.com/IntelPython/dpnp/pull/3066)
```
vlad-perevezentsev: `as the smallest` only holds for `mode="largest"`

dpnp/dpnp_iface_sorting.py
```diff
@@ -119,7 +119,8 @@ def argsort(
     descending : bool, optional
         Sort order. If ``True``, the array must be sorted in descending order
         (by value). If ``False``, the array must be sorted in ascending order
-        (by value).
+        (by value). NaN values (and complex values with a NaN component) are
+        ordered to the end regardless of `descending`.
```
vlad-perevezentsev, line 123: The same note is missing from the `dpnp.ndarray` methods (argsort, sort )

dpnp/tests/test_sort.py
```diff
@@ -545,24 +543,57 @@ def test_kind(self, kind):
...
         ia.sort(descending=descending)
-        a.sort(kind="stable")
-        if descending:
-            a = numpy.flip(a)
+        a.sort(stable=True, descending=descending)
         assert_array_equal(ia, a)
 
+    @testing.with_requires("numpy>=2.5")
+    @pytest.mark.parametrize("descending", [False, True])
+    @pytest.mark.parametrize("dtype", get_float_dtypes(no_float16=False))
```
vlad-perevezentsev, line 564: Could you add please
`@pytest.mark.parametrize("kind", [None, "stable", "mergesort", "radixsort"])`

dpnp/tests/third_party/cupy/sorting_tests/test_sort.py
```diff
@@ -515,9 +644,9 @@ def test_lexsort_workspace_oom(self):
             )
         )
 
-    # thread_unsafe marker requires pytest-run-parallel, not used by dpnp
```
vlad-perevezentsev, line 518: Why was this removed?

dpnp/tests/third_party/cupy/sorting_tests/test_sort.py
```diff
@@ -308,14 +375,23 @@ def test_F_order(self, xp):
...
+        xp = cupy.get_array_module(a)
+        if descending is None:
             # Need to explicitly specify kind="stable"
             # numpy uses "quicksort" as default
-            xp = cupy.get_array_module(a)
-            return xp.argsort(a, axis=axis, kind="stable")
+            kwargs = {"kind": "stable"}
+        else:
+            # numpy rejects `kind` combined with `descending`; `stable=True`
+            # is its replacement for forcing determinism. cupy's argsort is
```
vlad-perevezentsev, line 386: `dpnp.argsort` does accept `stable`
```
In [10]: dpnp.argsort(a, descending=True, stable=True)
Out[10]: array([5, 4, 3, 2, 1, 0])
```

dpnp/tensor/libtensor/include/utils/rich_comparisons.hpp
```diff
@@ -106,10 +107,38 @@ struct ExtendedComplexFPLess
 template <typename cT>
 struct ExtendedComplexFPGreater
```
vlad-perevezentsev, line 108: `ExtendedComplexFPGreater` line by line copy of `ExtendedComplexFPLess ` with operands swapped 
Probably easier to use something like `return ExtendedComplexFPLess<cT>{}(-v1, -v2);` below

dpnp/tests/third_party/cupy/sorting_tests/test_sort.py
antonwolfy: Reducing the number of lines in the diff against the upstream.
Commenting below line is clear message we don't support that, no need for the clarification.

CHANGELOG.md
antonwolfy: Good catch, thanks — you're right. Reworded to "to place `NaN` values last instead of first when sorting in descending order" so the entry states the new behavior.

CHANGELOG.md
antonwolfy: You're right. Since `NaN`s are ordered to the end in both modes, only `mode="largest"` actually changed, so I scoped the entry accordingly and dropped the "as the smallest" wording.

dpnp/tensor/_sorting.py
antonwolfy: Agreed, thanks for the example. Updated the docstring to say `NaN` values are ordered last for both modes, so they are not returned ahead of finite values.

dpnp/dpnp_iface_sorting.py
antonwolfy: Good catch — added the same note to the `dpnp.ndarray.sort` and `dpnp.ndarray.argsort` docstrings.

dpnp/tests/test_sort.py
antonwolfy: Done — parametrized `test_descending_nan` over `kind` ([None, "stable", "mergesort", "radixsort"]), so the descending NaN-at-end behavior is now checked for the radix-sort path too, not just merge sort.

dpnp/tests/third_party/cupy/sorting_tests/test_sort.py
antonwolfy: This comment is copied verbatim from the upstream cupy test (`tests/cupy_tests/sorting_tests/test_sort.py`); it's accurate for cupy, whose `argsort` is always stable and doesn't take a `stable` kwarg. I'd prefer to keep it unchanged to minimize the diff against the upstream test suite — the `if xp is numpy:` guard already ensures `stable=True` is passed only to numpy, and dpnp's default-stable `argsort` matches.

dpnp/tensor/libtensor/include/utils/rich_comparisons.hpp
antonwolfy: Good suggestion, done — `ExtendedComplexFPGreater` now delegates to `ExtendedComplexFPLess<cT>{}(-v1, -v2)`. Negation preserves NaN-ness, so the NaN grouping (and its ordering to the end) is unchanged while the finite comparison is reversed.

antonwolfy: One perf note for the record: this form constructs two `std::complex` temporaries (4 float negations) per comparison, whereas the previous hand-written body compared existing references. It's safe to ignore in practice — complex sort goes through the merge path (complex isn't radix-eligible), the per-compare cost is dominated by the branchy NaN-group logic inside `ExtendedComplexFPLess`, the negations are trivial and typically optimized out, and complex isn't a hot dtype for large sorts. The readability win of dropping the duplicated ~30-line body outweighs the negligible cost; we can revisit if profiling ever flags complex descending sort.

ndgrigorian: While it's more readable and less code duplication, I think we should revert it. Potential performance issues and while negation preserves `NaN`-ness, isn't it possible this produces inconsistent sign bits?

PR #3068, opened 2026-09-10
Fix for dpnp buffer-arg ndarray ignoring offset

### LINKED ISSUE #3067: `dpnp.ndarray` ctor returns a view at wrong addr with `buffer=` has non-zero offset
Reproducer:
```python
import numpy
import dpnp

for xp, name in [(numpy, "numpy"), (dpnp, "dpnp")]:
    pool = xp.arange(8, dtype=xp.complex64)  # itemsize 8
    tail = pool[4:]  # element offset 4 (complex64 units) = byte offset 32

    view = xp.ndarray((4,), dtype=xp.uint16, buffer=tail)  # itemsize 2

    expect_byte_offset = 32
    got_byte_offset = view.__array_interface__["data"][0] - pool.__array_interface__["data"][0] \
        if xp is numpy else view.data.ptr - pool.data.ptr

    print(f"{name}: byte_offset={got_byte_offset} (expected {expect_byte_offset})")
    assert got_byte_offset == expect_byte_offset, f"{name} FAILED"

print("OK: dpnp matches numpy")
```

output:
```bash
numpy: byte_offset=32 (expected 32)
dpnp: byte_offset=8 (expected 32)
  File "/home/abagusetty/gpu4pyscf-testing/dpnp_sparse/repro_offset_bug.py", line 19, in <module>
    assert got_byte_offset == expect_byte_offset, f"{name} FAILED"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: dpnp FAILED
```

abagusetty:
Fixes: #3067 

### DIFF
--- a/dpnp/dpnp_array.py
+++ b/dpnp/dpnp_array.py
@@ -127,9 +127,26 @@ class dpnp_array:
             # or as USM memory allocation
             if isinstance(buffer, dpnp_array):
                 buffer = buffer.get_array()
-                offset += buffer._element_offset
 
-            if dtype is None and hasattr(buffer, "dtype"):
+            if isinstance(buffer, dpt.usm_ndarray):
+                if dtype is None:
+                    dtype = buffer.dtype
+
+                # `buffer._element_offset` is expressed in units of the
+                # buffer's own dtype, while `offset` is interpreted in units
+                # of `dtype`, so the displacement has to be rescaled through
+                # bytes whenever the two itemsizes differ
+                byte_offset = buffer._element_offset * buffer.itemsize
+                new_itemsize = dpnp.dtype(dtype).itemsize
+                add_offset, rem = divmod(byte_offset, new_itemsize)
+                if rem != 0:
+                    raise ValueError(
+                        "The offset of the buffer's data in memory is not "
+                        "a multiple of the requested dtype size and so the "
+                        "requested view is not possible"
+                    )
+                offset += add_offset
+            elif dtype is None and hasattr(buffer, "dtype"):
                 dtype = buffer.dtype
         else:
             buffer = usm_type
--- a/dpnp/dpnp_array.py
+++ b/dpnp/dpnp_array.py
@@ -128,26 +128,14 @@ class dpnp_array:
             if isinstance(buffer, dpnp_array):
                 buffer = buffer.get_array()
 
-            if isinstance(buffer, dpt.usm_ndarray):
-                if dtype is None:
-                    dtype = buffer.dtype
-
-                # `buffer._element_offset` is expressed in units of the
-                # buffer's own dtype, while `offset` is interpreted in units
-                # of `dtype`, so the displacement has to be rescaled through
-                # bytes whenever the two itemsizes differ
-                byte_offset = buffer._element_offset * buffer.itemsize
-                new_itemsize = dpnp.dtype(dtype).itemsize
-                add_offset, rem = divmod(byte_offset, new_itemsize)
-                if rem != 0:
-                    raise ValueError(
-                        "The offset of the buffer's data in memory is not "
-                        "a multiple of the requested dtype size and so the "
-                        "requested view is not possible"
-                    )
-                offset += add_offset
-            elif dtype is None and hasattr(buffer, "dtype"):
+            if dtype is None and hasattr(buffer, "dtype"):
                 dtype = buffer.dtype
+
+            if isinstance(buffer, dpt.usm_ndarray):
+                # `_element_offset` is in buffer-dtype units; the ctor's
+                # `offset` is in `dtype` units, so rescale via bytes when
+                # itemsizes differ
+                offset += dpnp_array._rescaled_element_offset(buffer, dtype)
         else:
             buffer = usm_type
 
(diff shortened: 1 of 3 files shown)


dpnp/dpnp_array.py
```diff
@@ -127,9 +127,26 @@ def __init__(
...
+                # bytes whenever the two itemsizes differ
+                byte_offset = buffer._element_offset * buffer.itemsize
+                new_itemsize = dpnp.dtype(dtype).itemsize
+                add_offset, rem = divmod(byte_offset, new_itemsize)
+                if rem != 0:
+                    raise ValueError(
+                        "The offset of the buffer's data in memory is not "
+                        "a multiple of the requested dtype size and so the "
+                        "requested view is not possible"
+                    )
```
antonwolfy: This offset-rescaling is now duplicated with `_create_view` — but the two raise different `ValueError` messages.
Could we factor the shared computation into a small helper, e.g.:
```python
@staticmethod
def _rescaled_element_offset(usm_ary, new_dtype):
    """Element offset of `usm_ary` re-expressed in units of `new_dtype`."""
    byte_offset = usm_ary._element_offset * usm_ary.itemsize
    new_itemsize = dpnp.dtype(new_dtype).itemsize
    offset, rem = divmod(byte_offset, new_itemsize)
    if rem:
        raise ValueError(
            "The offset of the buffer's data in memory is not a multiple "
            "of the requested dtype size and so the requested view is not "
            "possible"
        )
    return offset
```
and call it from both `__init__` and `_create_view`?

dpnp/dpnp_array.py
```diff
@@ -127,9 +127,26 @@ def __init__(
             # or as USM memory allocation
             if isinstance(buffer, dpnp_array):
                 buffer = buffer.get_array()
-                offset += buffer._element_offset
 
-            if dtype is None and hasattr(buffer, "dtype"):
+            if isinstance(buffer, dpt.usm_ndarray):
```
antonwolfy: The dtype-defaulting could be hoisted out of the branch so it runs once for any buffer with a `.dtype`, instead of being duplicated between the inner `if dtype is None` and the `elif`. That also makes the ordering explicit: the offset math needs `dtype` for `new_itemsize`.
```python
if isinstance(buffer, dpnp_array):
    buffer = buffer.get_array()

if dtype is None and hasattr(buffer, "dtype"):
    dtype = buffer.dtype

if isinstance(buffer, dpt.usm_ndarray):
    byte_offset = buffer._element_offset * buffer.itemsize
    new_itemsize = dpnp.dtype(dtype).itemsize
    add_offset, rem = divmod(byte_offset, new_itemsize)
    if rem != 0:
        raise ValueError(...)
    offset += add_offset
```

dpnp/dpnp_array.py
```diff
@@ -127,9 +127,26 @@ def __init__(
...
+                if dtype is None:
+                    dtype = buffer.dtype
+
+                # `buffer._element_offset` is expressed in units of the
+                # buffer's own dtype, while `offset` is interpreted in units
+                # of `dtype`, so the displacement has to be rescaled through
+                # bytes whenever the two itemsizes differ
```
antonwolfy:
```suggestion
                # `_element_offset` is in buffer-dtype units; the ctor's `offset` is in
                # `dtype` units — rescale via bytes when itemsizes differ
```

dpnp/tests/test_ndarray.py
```diff
@@ -529,6 +529,68 @@ def test_nonzero_offset_buffer_ctor(self):
...
+        [
+            (dpnp.complex64, dpnp.uint16),
+            (dpnp.complex128, dpnp.float64),
+            (dpnp.float64, dpnp.float32),
+            (dpnp.int64, dpnp.int8),
+            (dpnp.int32, dpnp.int16),
+            (dpnp.int16, dpnp.int64),
+        ],
+    )
+    def test_nonzero_offset_buffer_ctor_dtype_mismatch(self, src_dt, new_dt):
```
antonwolfy, line 543: Two small gaps worth filling while we're here (non-blocking):

1. No test exercises a non-zero incoming `offset=` arg *together* with a dtype-mismatched buffer — the `offset=1` case uses a same-itemsize dtype, so the `offset += add_offset` addition is never checked with a non-trivial `add_offset`. A case like `dpnp.ndarray((size,), dtype=new_dt, buffer=sl, offset=1)` would cover it.

2. The misalignment `ValueError` is only asserted for a `dpnp_array` buffer (`test_misaligned_offset_buffer_ctor_error`). Since the fix routes bare `usm_ndarray` buffers through the same branch, a matching case with `buffer=base[3:].get_array()` would guard that path too.

antonwolfy, line 543: This test hard-fails on devices without native fp64/complex128 support.

dpnp/dpnp_array.py
abagusetty: good idea. fixed it!

dpnp/dpnp_array.py
abagusetty: Done!

dpnp/dpnp_array.py
abagusetty: addressed!

dpnp/tests/test_ndarray.py
abagusetty: guarded it

abagusetty: Addressed both aspects

dpnp/dpnp_array.py
vlad-perevezentsev: @antonwolfy `hasattr(buffer, "dtype")` is true exactly when `isinstance(buffer, dpt.usm_ndarray)` is. The ctor accepts only `_Memory`, `str`/`bytes` and `usm_ndarray` and only the last carries a dtype. So the wider condition adds no case, and it leaves the offset math relying on a different predicate to have defaulted `dtype` which the call site no longer shows
I would suggest something like this:
```python
if isinstance(buffer, dpnp_array):
    buffer = buffer.get_array()

if isinstance(buffer, dpt.usm_ndarray):
    # a `usm_ndarray` is the only supported buffer carrying a dtype,
    # USM memory allocations and `usm_type` strings do not
    if dtype is None:
        dtype = buffer.dtype

    # `_element_offset` is in buffer-dtype units, while the ctor's
    # `offset` is in `dtype` units, so it has to be rescaled through bytes
    offset += dpnp_array._rescaled_element_offset(buffer, dtype)

```

vlad-perevezentsev (approved):
LGTM
Thank you @abagusetty

PR #3058, opened 2026-09-02
Fix `dpnp.einsum` memory-layout

abagusetty:
One more minor layout issue detected during an app-testing in comparison to numpy behavior:

Fixes: #3056 

- [x] Have you provided a meaningful PR description?
- [x] Have you added a test, reproducer or referred to an issue with a reproducer?
- [x] Have you tested your changes locally for CPU and GPU devices?
- [x] Have you made sure that new changes do not introduce compiler warnings?
- [x] Have you checked performance impact of proposed changes?
- [ ] Have you added documentation for your changes, if necessary?
- [x] Have you added your changes to the changelog?

PR #3069, opened 2026-09-14
Avoid a C-order copy in `dpnp.einsum`

vlad-perevezentsev:
This PR improves `dpnp.einsum` performance by avoiding a copy into C-order.

Commit 039b7f14 (#3058) remaps `order="K"` to `"C"` which turned the final `dpnp.asarray` from a no-op into a full device copy because dpnp contracts operands in reverse order so `"ij,jk"` computes `(a @ b).T` and the result is naturally `f-contiguous`

This PR builds the result directly in the requested layout instead of fixing it afterwards. The last contraction determines the output axis order, so it can choose which operand is on the left side of the matmul and avoid an extra copy.
This is only applied for `"C"` order, since for other layouts the default operand order already matches the expected `order="K" + optimize=True` behavior. The swap does not move any data. It only changes the operand order in the matmul

The performance result `bench_linalg.MatMul.time_einsum_ij_jk` on PVC:

| dtype | size | main | fix|
|---|---|---|---|
| float64 | 1024 | 0.953 ms | **0.612 ms (x0.64)** |
| float64 | 4096 | 8.590 ms | **7.490 ms (x0.87)** |
| float32 | 1024 | 0.574 ms | **0.439 ms (x0.76)** |
| float32 | 4096 | 7.577 ms | **7.005 ms (x0.92)** |
| int32 | 1024 | 1.891 ms | **1.301 ms (x0.69)** |
| int32 | 4096 | 54.324 ms | **35.045 ms (x0.65)** |
| int64 | 1024 | 4.637 ms | **3.735 ms (x0.81)** |
| int64 | 4096 | 200.454 ms | **162.920 ms (x0.81)** |


ndgrigorian (approved):
LGTM

PR #3072, opened 2026-09-19
Fix event dependency bugs and initial displacement initialization

ndgrigorian:
This PR fixes bugs in libtensor:
* `copy_for_reshape` default-constructed events unnecessarily
* `fill` dropped its `depends` when constructing strides on device
* Stride simplification did not consistently set initial displacements to 0


dpnp/tensor/libtensor/source/copy_for_reshape.cpp
```diff
@@ -152,9 +152,10 @@ std::pair<sycl::event, sycl::event>
     const char *src_data = src.get_data();
     char *dst_data = dst.get_data();
 
-    std::vector<sycl::event> all_deps(depends.size() + 1);
-    all_deps.push_back(copy_shape_ev);
+    std::vector<sycl::event> all_deps;
+    all_deps.reserve(depends.size() + 1);
     all_deps.insert(std::end(all_deps), std::begin(depends), std::end(depends));
```
vlad-perevezentsev: This pattern also repeats at lines 236 and 360 below.
Does it make sense to apply the fix there as well?

